Issue #31, opened 2022-02-01
Add a re-reading of the configuration when an instance is restarted


LeonidVas:
Now when restarting the instance, the old configuration will be used, but we can change some settings (like restarting) and we don't want to force a restart of the instance to apply them. The suggestion is to add a re-read of the configuration when the instance is restarted.

LeonidVas:
Some clarifications:
* This implies that the configure has to be re-read at the watchdog level. For example, the parameter `restart` was set to false at startup and decided to change it to true, then the changes are made in tarantool.yaml and `tt restart` is not called. In case of a crash, the config must be reread before trying to restart the instance
* The configuration parsing must be moved from `pakage modules` to `pakage configure`. Or explain why this should not be done

LeonidVas:
After discussion, it was decided that using `tt run` to restart an instance in watchdog was not convenient. It was decided to implement the re-reading of the config by refactoring watchdog.

Issue #29, opened 2022-01-31
Implement `tt run`


LeonidVas:
`tt run` should to have the same functionality as `tt start`, except that it does not use `watchdog` and output should not be redirected, just write to `stdout`, `stderr`. The main purpose of this command is to check during development how the application will work under `tt`.
Also, in this mode. no one command of `tt` which work through `watchdog` can be used to interact with an instance.

PR #49, opened 2022-04-14
Add a re-reading of the configuration when an instance is restarted

better0fdead:
Moved configure to package config.
Added re-reading of the configuration at watchdog level. Currently it changes only watchdog.restartable if u have any other suggestions I am willing to hear them.


cli/configure/config.go
```diff
@@ -1,60 +1,20 @@
-package modules
+package configure
```
0x501D, line 1: Rename config.go -> configure.go

cli/config/cfg.go
```diff
@@ -0,0 +1,42 @@
+package config
```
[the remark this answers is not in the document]
0x501D: Why config structure is located in separate directiry('config') instead of ('configure') ?

cli/running/lua/launcher.lua
```diff
@@ -61,4 +61,4 @@ local function start_instance()
     return 0
 end
 
-start_instance()
+start_instance()
```
0x501D, line 64: Unnecessary changes

cli/running/watchdog.go
```diff
@@ -27,14 +28,15 @@ type Watchdog struct {
 	restartTimeout time.Duration
 	// done channel used to inform the signal handle goroutine
 	// about termination of the Instance.
```
0x501D, line 30: Comment does not match next variable

cli/running/watchdog.go
```diff
@@ -62,7 +64,11 @@ func (wd *Watchdog) Start() {
 		wd.done <- true
 		// Wait for the signal processing goroutine to complete.
 		wd.doneBarrier.Wait()
-
+		// Parse cfg again to check if values changed.
```
0x501D, line 67: to check that the values have changed

0x501D (changes requested):
Please resolve conflicts.

cli/config/cfg.go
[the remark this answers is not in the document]
better0fdead: It is located in separate directory because modules.go needs it, and configure needs modules, so if config will be part of configure, configure and modules will be dependencies of each other.

cli/running/watchdog.go
better0fdead: Fixed

cli/running/watchdog.go
better0fdead: Fixed

cli/running/lua/launcher.lua
better0fdead: Removed

cli/configure/config.go
better0fdead: Fixed

cli/config/cfg.go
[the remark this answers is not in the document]
0x501D: accepted

go.mod
```diff
@@ -39,4 +39,5 @@ require (
 	golang.org/x/sys v0.0.0-20211205182925-97ca703d548d // indirect
 	golang.org/x/term v0.0.0-20201126162022-7de9c90e9dd1 // indirect
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
+ 	gopkg.in/natefinch/lumberjack.v2 v2.0.0 // indirect
```
0x501D, line 42: Is this addition belongs to this feature?

better0fdead: No, fixed

0x501D (approved):
LGTM

LeonidVas (changes requested):
How do you think, it would be better to move `ModuleInfo` to the `configure` package? Or that won't solve the problem.
Mark `tt: restruct project so parsing of cfg is now in configure package` as `Part of #31`

LeonidVas (commented):
`watchdog: add re-reading of cfg after instance is killed`
`After instance shutdown added update of wachtdog settings` - Set a dot at the end of the sentence.
Perhaps `When an instance restarts after shutting down` would be better.

LeonidVas:
In fact, we now have options that will be re-read from the config in case the instance is restarted by watchdog. I think we should describe this in the README.

cli/running/watchdog.go
```diff
@@ -62,7 +66,11 @@ func (wd *Watchdog) Start() {
 		wd.done <- true
 		// Wait for the signal processing goroutine to complete.
 		wd.doneBarrier.Wait()
-
+		// Parse config again to check if the values have changed.
+		if wd.configPath != "" {
+			newCliopts, _ := configure.GetCliOpts(wd.configPath)
```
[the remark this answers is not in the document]
better0fdead: Fixed

cli/config/config.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// CliOpts stores information about Tarantool CLI configuration.
+// Filled in when parsing the tarantool.yaml configuration file.
+//
+// tarantool.yaml file format:
+// tt:
+//   modules:
+//     directory: path/to
+//...app:
```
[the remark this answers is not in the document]
better0fdead: Fixed

better0fdead:
> Hi! Thank you for the patchset. How do you think, it would be better to move `ModuleInfo` to the `configure` package? Or that won't solve the problem. Mark `tt: restruct project so parsing of cfg is now in configure package` as `Part of #31`

No, that wont solve the problem since we need to call Runcmd from package modules. As for me that structure of project is the best we could do.

LeonidVas (changes requested):
I had a discussion with my colleagues about this task and it looks like rereading the configuration would be logical to implement by starting to use `tt run` instead of reusing `Instance`. I propose to close this ticket by taking the refactoring commits and implement the rereading within the `tt run` [task](https://github.com/tarantool/tt/issues/29).

PR #50, opened 2022-04-14
Add processing of error when no application name is specified

better0fdead:
Previously if no application name was specified output was:
"Currently, you can specify only one instance at a time."
which is incorrect. So I processed it.

### DIFF
--- a/cli/running/running.go
+++ b/cli/running/running.go
@@ -196,7 +196,11 @@ func createPIDFile(pidFileName string) error {
 // FillCtx fills the RunningCtx context.
 func FillCtx(cliOpts *modules.CliOpts, cmdCtx *cmdcontext.CmdCtx, args []string) error {
 	if len(args) != 1 {
-		return fmt.Errorf("Currently, you can specify only one instance at a time.")
+		if len(args) > 1 {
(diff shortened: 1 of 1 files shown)


better0fdead:
I am not sure why 1 check failed, I think it is on gh side bcs other checks passed and fix has nothing to do with log rotation.

LeonidVas:
> I am not sure why 1 check failed, I think it is on gh side bcs other checks passed and fix has nothing to do with log rotation.

It's like a race in the test.

LeonidVas (changes requested):
Hi! Thank you for the patch.
I don't think that it's useful, but I don't mind.
See https://github.com/tarantool/tarantool/wiki/Code-review-procedure#commit-message 
> Commit message should fit 72 symbols line width.

Add <submodule> (`running: ....`).
init file -> instance file
Proccesed -> Processed
Processed -> `add processing` (It's convenient when the title sounds like a command and in some clever book was it written to do so)
output was:Currently, you can specify only one instance at a time which is incorrect ->  output was: "Currently, you can specify only one instance at a time which is incorrect"
So i -> So I

LeonidVas (approved):

PR #51, opened 2022-04-19
Implementation of `tt connect'

LeonidVas:
Implementation of the `tt connect` module.


cli/util/util.go
```diff
@@ -1,20 +1,33 @@
...
+	"os/user"
 	"path/filepath"
 	"runtime/debug"
 
+	"github.com/apex/log"
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v2"
 )
 
+var bufSize int64 = 10000
```
0x501D, line 17: const bufSize int64 = 10000 ?

cli/util/util.go
```diff
@@ -91,3 +140,137 @@ func GetHelpCommand(cmd *cobra.Command) *cobra.Command {
...
+	return lastNewLinePos, nil
+}
+
+func GetLastNLines(filepath string, linesN int) ([]string, error) {
+	lastNLinesBeginPos, err := GetLastNLinesBegin(filepath, linesN)
+	if err != nil {
+		return nil, err
+	}
+
+	file, err := os.Open(filepath)
```
0x501D, line 225: Why not using mmap technique for working with files. IMHO it will be simpler.

cli/connector/connector.go
```diff
@@ -0,0 +1,125 @@
...
+)
+
+func Connect(connString string, opts Opts) (*Conn, error) {
+	var err error
+
+	conn := &Conn{}
+
+	connOpts := getConnOpts(connString, opts.Username, opts.Password)
+
+	// connect to specified address
```
0x501D: Comment style.

0x501D: It already fixed in next commit, my fault.

cli/connect/console.go
```diff
@@ -128,6 +129,11 @@ func (console *Console) Run() error {
 
 	console.prompt.Run()
 
+	// Sets the terminal modes to “sane” values to workaround
+	// bug https://github.com/c-bata/go-prompt/issues/228
+	reinitialization := exec.Command("reset")
```
0x501D, line 134: Nit: reset is not good way to fix this. It will clear previous stdin|stdout. 
The user may not expect this behavior.
It's good practice to restore the terminal without clearing the screen

0x501D (approved):

cli/util/util.go
LeonidVas: Fixed.

cli/util/util.go
LeonidVas: This is a good suggestion, I will implement it as part of the "folow up".

PR #57, opened 2022-04-24
launcher: fix stdout buffering

### LINKED ISSUE #36: Fix the problem with the log write
When a new instance of tarantool is started in `instance.go`, by default the log will be written with block buffering. The `stdbuf` utility was used to change this behavior, but it is not installed on macOS by default. Needs to investigate the problem and implement a universal solution.
```
// Start starts the Instance with the specified parameters.
func (inst *Instance) Start() error {
  // By default (when using "glibc") "stdout" is line buffered when connected
  // to a TTY and block buffered (one page 4KB) when connected to a pipe / file.
  // This is not how we want to log work, so set "stdout" to line buffered mode
  // by using "stdbuf" utility. "strderr" is set to no-buffering by default.
  //
  // Several useful links:
  // https://www.pixelbeat.org/programming/stdio_buffering/
  // https://man7.org/linux/man-pages/man3/setbuf.3.html
  // https://github.com/coreutils/coreutils/blob/master/src/stdbuf.c
  inst.Cmd = exec.Command("stdbuf", "-o", "L",
    inst.tarantoolPath, "-e", instanceLauncher)
  inst.Cmd.Stdout = inst.logger.Writer()
  inst.Cmd.Stderr = inst.logger.Writer()
  inst.Cmd.Env = append(os.Environ(), "TT_CLI_INSTANCE="+inst.appPath)
  inst.Cmd.Env = append(inst.Cmd.Env,
    "TT_CLI_CONSOLE_SOCKET="+inst.consoleSocket)

  // Imitate the "tarantoolctl".
  inst.Cmd.Env = append(inst.Cmd.Env, "TARANTOOLCTL=true")
  // Set the sign that the program is running under "tt".
  inst.Cmd.Env = append(inst.Cmd.Env, "TT_CLI=true")

  // Start an Instance.
  if err := inst.Cmd.Start(); err != nil {
    return err
  }
  inst.done = false

  return nil
}
```

Note that the original problem may have been misunderstood by the author.

0x501D:
This patch enables linebuffering for stdout fstream.

Closes #36

### DIFF
--- a/cli/running/lua/launcher.lua
+++ b/cli/running/lua/launcher.lua
@@ -5,6 +5,7 @@ local os = require('os')
 local console = require('console')
 local log = require('log')
 local title = require('title')
+local ffi = require('ffi')
 
 
 --- Start an Instance. The "init" file of the Instance passes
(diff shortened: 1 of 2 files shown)


0x501D:
This patch was tested with:
* gentoo (glibc)
* alpine (musl)
* macosx

cli/running/lua/launcher.lua
```diff
@@ -20,6 +21,32 @@ local function start_instance()
         __defer_update = true
     }
 
+    -- Make stdout linebuffered. gh-36.
+    -- By default (when using "glibc") "stdout" is line buffered when connected
+    -- to a TTY and block buffered (one page 4KB) when connected to a pipe / file.
+    -- This is not how we want to log work, so set "stdout" to line buffered mode
```
Totktonada, line 27: It is not about log, but about `print()` in Lua, see https://github.com/tarantool/tt/issues/36#issuecomment-1106915036.

cli/running/lua/launcher.lua
```diff
@@ -20,6 +21,32 @@ local function start_instance()
...
+    -- by calling "setlinebuf(3)". "strderr" is set to no-buffering by default.
```
Totktonada, line 28: Typo: strderr -> stderr.

cli/running/lua/launcher.lua
0x501D: Comment rewritten.

cli/running/lua/launcher.lua
0x501D: Fixed

LeonidVas (approved):

PR #59, opened 2022-04-29
readme: minimum golang version requirements

0x501D:
tt rocks uses go:embed for embeding luarocks sourses.
This feature is supported in Go since version 1.16.


LeonidVas:
The version is specified in go.mod, so:
* Why do we need to specify it in more than one place?
* I think, this version must be the same in all places.

0x501D:
go.mod contains go version at the moment of creation go.mod file. 
But this version is not the real minimal requirement.

LeonidVas:
I read the https://go.dev/ref/mod#go-mod-file-go .
And see the following repositories:
https://github.com/etcd-io/etcd
https://github.com/kubernetes/kubernetes
https://github.com/gohugoio/hugo
https://github.com/coredns/coredns
So, I agree to specify the go version in the README, but it must be the same version as in the `go.mod` file, because the module was written using this version and this version must be used in CI tests.(1.17+)

0x501D:
Version was updated to 1.17+

PR #68, opened 2022-05-16
Fix logrotate test

LeonidVas:
Before the patch, a race may occur in the "logrotate" test, when the log files are checked on existen in the directory but they have not yet been created.


0x501D (approved):
LGTM

Issue #58, opened 2022-04-26
Update/extend the functionality of the `tt connect` module


LeonidVas:
* The eval (`console.eval`) should be used only in cases:

1. `tt connect -f FILE_NAME`
2. `COMMAND | tt connect -f -`

* Add `-x` option enables a mode where all executed commands are printed to the terminal (trace mode like in [bash](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html))
* Add `--language` option (Lua(`\set language lua`), SQL(`\set language sql`))

LeonidVas:
Add possibility to pass credentials through URI.

LeonidVas:
A few more wishes:
- pass credentials via env. vars, as in the AWS CLI
- `COMMAND | tt connect ` should execute commands line by line and leave the connection console to continue working after

unera:
> Add -x option enables a mode where all executed commands are printed to the terminal (trace mode like in [bash](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html))
>    Add --language option (Lua(\set language lua), SQL(\set language SQL))

stop

to be like as bash - wrong idea, it would be nice be similar to psql

so

- `-x sql`
- `-x lua`
- (future): `-x table,yaml,table,sql` 
- etc

oleg-jukovec:
The problem with traces:

If we start executing valid commands one by one via connector eval then we are facing a problem:
```
console.eval("local i = 10")
console.eval("return i")
```

is not equal to:

```
console.eval("local i = 10\nreturn i")
```

So it will be different behavior in case `file.lua`:
```Lua
local i = 10
return i
```
and tt connect -f file.lua with/without -trace option.


A patch file to play:
[trace.patch.txt](https://github.com/tarantool/tt/files/9582766/trace.patch.txt)

oleg-jukovec:
Tarantool does not have SQL autocompletion: https://github.com/tarantool/tarantool/issues/2304

oleg-jukovec:
Let's summarize:

There is only one wish left here `Add -x option enables a mode where all executed commands are printed to the terminal (trace mode like in [bash](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html))`. See also: https://github.com/tarantool/tt/issues/58#issuecomment-1249139450

LeonidVas:
For the input language selection option we left `-l` . `-x` will probably be used to select the output format (`table,yaml`).

PR #77, opened 2022-06-17
Add `-f` flag for evaluation

LeonidVas:
This patch changes the behavior of `connect` so that the source for the code evaluation is now passed by using the `-f` flag.

Part of #58

### DIFF
--- a/cli/cmd/connect.go
+++ b/cli/cmd/connect.go
@@ -20,6 +20,7 @@ import (
 var (
 	connectUser     string
 	connectPassword string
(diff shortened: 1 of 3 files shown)


LeonidVas:
This is only part of the functionality to be added as the result of the task, but since it changes the behavior of the team, I decided to submit it for review early.

LeonidVas:
Tests will be added along with other tests for the `connect` module.

0x501D (approved):
LGTM

PR #75, opened 2022-06-14
Add tt crud-import module

GRISHNOV:
Add a module responsible for importing the contents from file to Tarantool via crud.

Closes #35

PR #88, opened 2022-06-27
Update logic of crud batching api

GRISHNOV:
A change in the crud api has been taken into account, now in the error variable from batching operation the key with tuple value has actual name (`tuple` -> `operation_data`).

Related with [this](https://github.com/tarantool/tt/pull/75#discussion_r906307118).

Part of #87


LeonidVas (changes requested):
Hi! Thank you for the patch.
- `fix` not is some subsystem. See https://github.com/tarantool/tarantool/wiki/Code-review-procedure#commit-message In your case you can use the `crud`.
- if the logic has been `updated`, it is not a fix (choose one of the two).
- specify the release of the `crud` from which the change was applied,

GRISHNOV:
Thank you for your feedback! I tried to answer all the comments on the code review

PR #107, opened 2022-07-26
Add search module

### LINKED ISSUE #95: Implement `tt search` module
Implement a module that will be used to search for utilities to install using `tt install`.
The `tt install` command should return a list of utilities.
The command `tt install UTILITY` should return a list of available versions of UTILITY.

better0fdead:
Added search module to to show available tt/tarantool versions.
Closes #95

### DIFF
--- a/cli/cmd/search.go
+++ b/cli/cmd/search.go
@@ -1,0 +1,41 @@
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/apex/log"
+	"github.com/spf13/cobra"
+	"github.com/tarantool/tt/cli/cmdcontext"
+	"github.com/tarantool/tt/cli/modules"
+	"github.com/tarantool/tt/cli/search"
+)
+
+// NewSearchCmd creates search command.
+func NewSearchCmd() *cobra.Command {
+	var searchCmd = &cobra.Command{
+		Use:   "search <PROGRAM>",
+		Short: "Search available versions of tarantool/tt",
+		Run: func(cmd *cobra.Command, args []string) {
+			err := modules.RunCmd(&cmdCtx, cmd.Name(), &modulesInfo, internalSearchModule, args)
+			if err != nil {
+				log.Fatalf(err.Error())
+			}
+		},
+	}
+	return searchCmd
+}
+
+// internalSearchModule is a default search module.
+func internalSearchModule(cmdCtx *cmdcontext.CmdCtx, args []string) error {
+	if len(args) == 0 {
+		log.Warnf("Avaliable programms: \n" +
+			"tarantool - OpenSource tarantool\n" +
+			"tt - OpenSource tarantool CLI")
+		return nil
+	}
+	if len(args) != 1 {
+		return fmt.Errorf("Incorrect arguments")
+	}
+	err := search.SearchVersions(args[0])
+	return err
+}
--- a/cli/search/search.go
+++ b/cli/search/search.go
@@ -1,0 +1,72 @@
+package search
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/apex/log"
+)
+
+// isDeprecated checks if the program version is lower than 1.10.0.
+func isDeprecated(version string) bool {
+	splitedVersion := strings.Split(version, ".")
+	if len(splitedVersion) < 2 {
+		return false
+	}
+	if splitedVersion[0] == "1" && len(splitedVersion[1]) < 2 {
+		return true
+	}
+	return false
+}
+
+// SearchVersions outputs available versions of program.
+func SearchVersions(program string) error {
+	var cmd *exec.Cmd
+
+	if _, err := exec.LookPath("git"); err != nil {
+		return fmt.Errorf("'git' is needed for 'tt search' to work")
+	}
+
+	if program == "tarantool" {
+		cmd = exec.Command("git", "-c", "versionsort.suffix=-", "ls-remote", "--tags", "--sort="+
+			"v:refname", "https://github.com/tarantool/tarantool.git")
+	} else if program == "tt" {
+		cmd = exec.Command("git", "-c", "versionsort.suffix=-", "ls-remote", "--tags", "--sort="+
+			"v:refname", "https://github.com/tarantool/tt.git")
+	} else {
+		return fmt.Errorf("Search supports only tarantool/tt")
+	}
+
+	log.Warn("Available versions of " + program + ":")
+
+	readPipe, writePipe, _ := os.Pipe()
+	cmd.Stdout = writePipe
+	cmd.Stderr = os.Stderr
+	cmd.Start()
+	err := cmd.Wait()
+	writePipe.Close()
+	if err != nil {
+		return err
+	}
+
+	var buf bytes.Buffer
+	io.Copy(&buf, readPipe)
+	versions := buf.String()
+	versionsArray := strings.Split(versions, "\n")
+	for i := 0; i < len(versionsArray); i++ {
+		trimPos := strings.LastIndex(versionsArray[i], "/") + 1
+		versionsArray[i] = versionsArray[i][trimPos:]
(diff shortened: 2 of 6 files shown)


cli/search/search.go
```diff
@@ -0,0 +1,62 @@
...
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/apex/log"
+)
+
+// isDeprecated returns if version is greater than 1.9.0.
```
0x501D, line 14: returns what?

cli/search/search.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	if splitedVersion[0] == "1" && len(splitedVersion[1]) < 2 {
+		return true
+	}
+	return false
+}
+
+// SearchVersions outputs available versions of program.
+func SearchVersions(programm string) error {
+	cmd := exec.Command("foobar")
```
0x501D, line 28: forgotten debug message? (foobar)

cli/search/search.go
```diff
@@ -0,0 +1,62 @@
...
+	if programm == "tarantool" {
+		cmd = exec.Command("bash", "-c", "git ls-remote --tags --sort="+
+			"v:refname https://github.com/tarantool/tarantool.git | cut --delimiter='/' --fields=3")
+	} else if programm == "tt" {
+		cmd = exec.Command("bash", "-c", "git ls-remote --tags --sort="+
+			"v:refname https://github.com/tarantool/tt.git | cut --delimiter='/' --fields=3")
+	} else {
+		return fmt.Errorf("Search supports only tarantool/tt")
+	}
+	log.Warn("Available versions of " + programm + ":")
```
0x501D, line 38: There are no line breaks between logically different actions, at least separate the data collection and their processing with a new line.

cli/search/search.go
better0fdead: Changed message

cli/search/search.go
better0fdead: Changed

cli/search/search.go
```diff
@@ -0,0 +1,67 @@
...
+		return true
+	}
+	return false
+}
+
+// SearchVersions outputs available versions of program.
+func SearchVersions(program string) error {
+	var cmd *exec.Cmd
+	if program == "tarantool" {
+		cmd = exec.Command("bash", "-c", "git ls-remote --tags --sort="+
```
0x501D, line 30: Why not `sh -c`. We do not use bash specific syntax here.
Also we add 'sed' dependencies here. We can parse output by golang instead of using sed.

better0fdead: Changed to sh, removed dependence

better0fdead: But I am little confused, because in conversation with @LeonidVas about coredump module we made a decision to use bash -c against sh, I understand that there is no real difference (as much as I know), but a little difference in style.

0x501D: After removing pipe we can get rid of using sh -c at all

better0fdead: Changed

cli/search/search.go
```diff
@@ -0,0 +1,67 @@
...
+		return true
+	}
+	return false
+}
+
+// SearchVersions outputs available versions of program.
+func SearchVersions(program string) error {
+	var cmd *exec.Cmd
+	if program == "tarantool" {
+		cmd = exec.Command("git", "ls-remote", "--tags", "--sort="+
```
0x501D, line 30: suffix=- will prevent 2.10.0-rc1 coming "after" 2.10.0

proposed patch:
```
--- a/cli/search/search.go
+++ b/cli/search/search.go
@@ -27,10 +27,10 @@ func isDeprecated(version string) bool {
 func SearchVersions(program string) error {
        var cmd *exec.Cmd
        if program == "tarantool" {
-               cmd = exec.Command("git", "ls-remote", "--tags", "--sort="+
+               cmd = exec.Command("git", "-c", "versionsort.suffix=-", "ls-remote", "--tags", "--sort="+
                        "v:refname", "https://github.com/tarantool/tarantool.git")
        } else if program == "tt" {
-               cmd = exec.Command("git", "ls-remote", "--tags", "--sort="+
+               cmd = exec.Command("git", "-c", "versionsort.suffix=-", "ls-remote", "--tags", "--sort="+
                        "v:refname", "https://github.com/tarantool/tt.git")
        } else {
                return fmt.Errorf("Search supports only tarantool/tt")
```
result:
```
--- /tmp/sort.old	2022-08-18 17:39:27.582526803 +0300
+++ /tmp/sort.new	2022-08-18 17:41:13.532532219 +0300
@@ -49,10 +49,10 @@
 2.8.3
 2.8.4
 2.9.0
-2.10.0
 2.10.0-beta1
 2.10.0-beta2
 2.10.0-rc1
+2.10.0
 2.10.1
 2.11.0-entrypoint
 ```

File: cli/cmd/search.go [hunk 6]
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/tarantool/tt/cli/cmdcontext"
+	"github.com/tarantool/tt/cli/modules"
+	"github.com/tarantool/tt/cli/search"
+)
+
+// NewSearchCmd creates search command.
+func NewSearchCmd() *cobra.Command {
+	var searchCmd = &cobra.Command{
+		Use:   "search [PROGRAM]",
+		Short: "search available versions of tarantool/tt",
```
0x501D, line 17: `search` should be capitalized for the same style of help output,

cli/search/search.go
better0fdead: Changed

cli/cmd/search.go
better0fdead: Fixed

0x501D (approved):
LGTM

cli/cmd/search.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/spf13/cobra"
+	"github.com/tarantool/tt/cli/cmdcontext"
+	"github.com/tarantool/tt/cli/modules"
+	"github.com/tarantool/tt/cli/search"
+)
+
+// NewSearchCmd creates search command.
+func NewSearchCmd() *cobra.Command {
+	var searchCmd = &cobra.Command{
+		Use:   "search [PROGRAM]",
```
LeonidVas, line 16: `<PROGRAM>` if I understand correctly (http://docopt.org/).

cli/search/search.go
```diff
@@ -0,0 +1,67 @@
...
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/apex/log"
+)
+
+// isDeprecated checks if the program version is lower than 1.9.0.
```
LeonidVas, line 14: Looks like it checks if the program version is lower than 1.10.0

cli/search/search.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	if splitedVersion[0] == "1" && len(splitedVersion[1]) < 2 {
+		return true
+	}
+	return false
+}
+
+// SearchVersions outputs available versions of program.
+func SearchVersions(program string) error {
+	var cmd *exec.Cmd
```
LeonidVas, line 28: I think we should check if `git' is installed on the host.
In the case of production `git' may be absent.

LeonidVas (changes requested):
HI! Thank you for the patchset.
See several comments bellow.

cli/search/search.go
0x501D: Example from current PR
https://github.com/tarantool/tt/pull/119/files#diff-0d5489a394b59a95c5e89f46fbf9afb6715c790a6ddb80626e99a917c508dedcR43

cli/cmd/search.go
better0fdead: changed

cli/search/search.go
better0fdead: Changed

cli/search/search.go
better0fdead: Added

PR #112, opened 2022-08-09
test: add free port auto resolve mechanism

### LINKED ISSUE #71: Create a helper to find a free TCP port
In integration tests we have hardcoded ports:
[Example](https://github.com/tarantool/tt/blob/master/test/integration/play/test_play.py#L45)

This can cause tests to fail:
```
    def test_play_test_remote_instance(tt_cmd, tmpdir):
        # Copy the .xlog and instance config files to the "run" directory.
        test_app_path = os.path.join(os.path.dirname(__file__), "test_file")
        shutil.copy(test_app_path + "/test.xlog", tmpdir)
        shutil.copy(test_app_path + "/remote_instance_cfg.lua", tmpdir)
    
        # Play .xlog file to the remote instance.
        cmd = [tt_cmd, "play", "localhost:3301", "test.xlog", "--space=999"]
        instance = subprocess.Popen(["tarantool", "remote_instance_cfg.lua"], cwd=tmpdir)
        # The delay is needed so that the instance has time to start and configure itself
        sleep(1)
        rc, output = run_command_and_get_output(cmd, cwd=tmpdir)
        instance.kill()
>       assert rc == 0
E       assert 1 == 0

test/integration/play/test_play.py:51: AssertionError
```
It is proposed to create a helper that looks for a free TCP port. This will greatly reduce the chance of falsely failed tests.

GRISHNOV:
Added free port auto resolve mechanism for integration tests, which require the running of a tarantool instance:

The idea is using a `box.cfg({listen = '127.0.0.1:0'})` call in the test instance configuration file and get an automatically assigned port with [box.info.listen](https://github.com/tarantool/tarantool/commit/1f4b9afb9db740aeeba71d7667ce13905597430d). For tarantool v1 we use [snippet](https://github.com/tarantool/tt/pull/112#issuecomment-1209661575) below for getting assigned port with `SO_REUSEADDR` socket filter by [this](https://github.com/tarantool/tt/pull/112#issuecomment-1232949223) approach. 

Now the solution looks like this:

If the user needs to raise a test instance for integration tests, then the user writes the configuration file of the test instance (for example, `test/integration/play/test_file/remote_instance_cfg.lua`). Within this config, the user needs to call `require('testutils').bind_free_port(arg[0])`. After that, the free port will be assigned and saved to disk for pytest. If you need to get the port value in the lua test instance cfg file, you can use `require('utils').get_bound_port()` after `require('utils').bind_free_port(arg[0])`.

Next, in pytest the user call `TarantoolTestInstance()` from utils.py. This call returns the test instance object, then use method `start()`, that creates a subprocess. This class implementation also copies the file utils.lua with `bind_free_port` implementation to pytest temp dir with instance config file.

Also added mechanism for guaranteed instances termination after integration testing:

There was a problem in which when the test failed, the created instance may not stop, but remained in the system. Now a wrapper `@pytest.fixture(autouse=True) kill_remain_instance_wrapper()` in `test_play.py` was implemented, which is guaranteed to stop the instance for each test. This approach can be used in further tests that will require the creation of their own test instances.

Closes #71

### DIFF
--- a/test/integration/play/test_play.py
+++ b/test/integration/play/test_play.py
@@ -2,11 +2,23 @@ import os
 import re
 import shutil
 
-from utils import TarantoolTestInstance, run_command_and_get_output
+import pytest
+from utils import (TarantoolTestInstance, kill_child_process,
+                   run_command_and_get_output)
 
 # The name of instance config file within this integration tests.
 # This file should be in /test/integration/play/test_file/.
 INSTANCE_NAME = "remote_instance_cfg.lua"
+
+
+# In case of unsuccessful completion of tests, tarantool test instances may remain running.
+# This is autorun wrapper for each test case in this module.
+@pytest.fixture(autouse=True)
+def kill_remain_instance_wrapper():
+    # Run test.
+    yield
+    # Kill a test instance if it was not stopped due to a failed test.
+    kill_child_process()
 
 
 def test_play_unset_arg(tt_cmd, tmpdir):
--- a/test/integration/play/test_play.py
+++ b/test/integration/play/test_play.py
@@ -1,10 +1,12 @@
 import os
 import re
 import shutil
-import subprocess
-from time import sleep
 
-from utils import run_command_and_get_output
+from utils import TarantoolTestInstance, run_command_and_get_output
+
+# The name of instance config file within this integration tests.
+# This file should be in /test/integration/play/test_file/.
+INSTANCE_NAME = "remote_instance_cfg.lua"
 
 
 def test_play_unset_arg(tt_cmd, tmpdir):
@@ -17,7 +19,7 @@ def test_play_unset_arg(tt_cmd, tmpdir):
 
 def test_play_non_existent_uri(tt_cmd, tmpdir):
     # Testing with non-existent uri.
-    cmd = [tt_cmd, "play", "localhost:0", "_"]
+    cmd = [tt_cmd, "play", "127.0.0.1:0", "_"]
     rc, output = run_command_and_get_output(cmd, cwd=tmpdir)
     assert rc == 1
     assert re.search(r"no connection to the host", output)
@@ -25,29 +27,36 @@ def test_play_non_existent_uri(tt_cmd, tmpdir):
 
 def test_play_non_existent_file(tt_cmd, tmpdir):
     # Testing with non-existent .xlog or .snap file.
-    cmd = [tt_cmd, "play", "localhost:49001", "path-to-non-existent-file"]
-    instance = subprocess.Popen(["tarantool", "-e", "box.cfg{listen = 'localhost:49001';}"])
-    # The delay is needed so that the instance has time to start and configure itself
-    sleep(1)
+    test_app_path = os.path.join(os.path.dirname(__file__), "test_file")
+
+    # Create tarantool instance for testing and start it.
+    path_to_lua_utils = os.path.join(os.path.dirname(__file__), "test_file/../../../")
+    test_instance = TarantoolTestInstance(INSTANCE_NAME, test_app_path, path_to_lua_utils, tmpdir)
+    test_instance.start()
+
+    # Run play with non-existent file.
+    cmd = [tt_cmd, "play", "127.0.0.1:" + test_instance.port, "path-to-non-existent-file"]
     rc, output = run_command_and_get_output(cmd, cwd=tmpdir)
-    instance.kill()
+    test_instance.stop()
     assert rc == 1
     assert re.search(r"No such file or directory", output)
 
 
 def test_play_test_remote_instance(tt_cmd, tmpdir):
-    # Copy the .xlog and instance config files to the "run" directory.
+    # Testing play using remote instance.
     test_app_path = os.path.join(os.path.dirname(__file__), "test_file")
+    # Copy the .xlog file to the "run" directory.
     shutil.copy(test_app_path + "/test.xlog", tmpdir)
-    shutil.copy(test_app_path + "/remote_instance_cfg.lua", tmpdir)
+
+    # Create tarantool instance for testing and start it.
+    path_to_lua_utils = os.path.join(os.path.dirname(__file__), "test_file/../../../")
+    test_instance = TarantoolTestInstance(INSTANCE_NAME, test_app_path, path_to_lua_utils, tmpdir)
+    test_instance.start()
 
     # Play .xlog file to the remote instance.
-    cmd = [tt_cmd, "play", "localhost:3301", "test.xlog", "--space=999"]
-    instance = subprocess.Popen(["tarantool", "remote_instance_cfg.lua"], cwd=tmpdir)
-    # The delay is needed so that the instance has time to start and configure itself
-    sleep(1)
+    cmd = [tt_cmd, "play", "127.0.0.1:" + test_instance.port, "test.xlog", "--space=999"]
     rc, output = run_command_and_get_output(cmd, cwd=tmpdir)
-    instance.kill()
+    test_instance.stop()
     assert rc == 0
     assert re.search(r"Play result: completed successfully", output)
 
--- a/test/utils.lua
+++ b/test/utils.lua
@@ -1,0 +1,101 @@
+-- This module contains auxiliary functions for integration tests that use lua code.
+-- To use this module, you need to copy this file to the temp pytest directory for the duration
+-- of the integration tests, then use the lua require function in your instance cfg file.
+-- The copying of this file to pytest temdir is already implemented in class
+-- TarantoolTestInstance of test/utils.py module.
+-- Just use require('utils').bind_free_port(arg[0]) inside your cfg instance file.
+-- If you need to get the port value in the lua test instance cfg file,
+-- use require('utils').get_bound_port() after require('utils').bind_free_port(arg[0]).
+
+local box    = require('box')
+local ffi    = require('ffi')
+local socket = require('socket')
+
+local testutils = {}
+
+function testutils.get_bound_address_old_tarantool()
+    -- Get bound via box.cfg({listen = '127.0.0.1:0'}) socket addr for tarantool major version 1.
+    -- For tarantool major version higher then 1, this function can be replaced by box.info.listen.
+    -- Table res contains listening and client sockets.
+    local res = {}
+    for fd = 0, 65535 do
+        local addrinfo = socket.internal.name(fd)
+        -- Assume that the socket listens on 127.0.0.1.
+        local is_matched = addrinfo ~= nil and addrinfo.host == '127.0.0.1' and
+            addrinfo.family == 'AF_INET' and addrinfo.type == 'SOCK_STREAM' and
+            addrinfo.protocol == 'tcp' and type(addrinfo.port) == 'number'
+        if is_matched then
+            addrinfo.fd = fd
+            table.insert(res, addrinfo)
+        end
+    end
+
+    -- Table l_res contains listening sockets.
+    local l_res = {}
+    -- We need only listening, not client sockets.
+    -- Filters sockets with SO_REUSEADDR.
+    for _, sock in pairs(res) do
+        local value  = ffi.new('int[1]')
+        local len    = ffi.new('size_t[1]', ffi.sizeof('int'))
+        local level  = socket.internal.SOL_SOCKET
+        local status = ffi.C.getsockopt(
+            sock.fd,
+            level,
+            socket.internal.SO_OPT[level].SO_REUSEADDR.iname,
+            value, len
+        )
+        if status ~= 0 then
+            error('problem with calling getsockopt() function')
+        end
+        if value[0] > 0 then
+            table.insert(l_res, sock)
+        end
+    end
+
+    -- If there are several listening sockets, we don't know which one is iproto's one.
+    if #l_res ~= 1 then
+        error(('zero or more than one listening TCP sockets: %d'):format(#l_res))
+    end
+
+    return ('%s:%s'):format(l_res[1].host, l_res[1].port)
+end
+
+function testutils.get_bound_port()
+    -- Returns bound port.
+    -- Can be used after call testutils.bind_free_port() if you need to get the port value
+    -- in the lua test instance cfg file.
+    local address = box.info.listen
+    if address == nil then
+        -- In case of tarantool major version 1.
+        address = testutils.get_bound_address_old_tarantool()
+    end
+    -- Get port from address string '127.0.0.1:*'.
+    local port = address:match(':(.*)')
+
+    if port == nil then
+        error('unable to get bound port, perhaps forgot to call testutils.bind_free_port()')
+    end
+
+    return port
+end
+
+function testutils.dump_bound_port(instance_file_name)
+    -- Dump bound port to file for pytest.
+    -- It will be regular text file which name is 'instance_file_name.port'.
+    local file, err = io.open(instance_file_name .. '.port', 'w')
+    if file == nil then
+        error(err)
+    end
+
+    local port = testutils.get_bound_port()
+    file:write(port)
+    file:close()
+end
+
+function testutils.bind_free_port(instance_file_name)
+    -- Bind free port and save it to file for pytest.
+    box.cfg({listen = '127.0.0.1:0'})
+    testutils.dump_bound_port(instance_file_name)
+end
+
+return testutils
--- a/test/utils.py
+++ b/test/utils.py
@@ -88,7 +88,7 @@ def wait_file(dir_name, file_pattern, exclude_list, timeout_sec=1):
     return ""
 
 
-def kill_child_process(pid):
+def kill_child_process(pid=psutil.Process().pid):
     parent = psutil.Process(int(pid))
     procs = parent.children()
 
--- a/test/utils.py
+++ b/test/utils.py
@@ -1,9 +1,11 @@
 import os
 import re
+import shutil
 import subprocess
 import time
 
 import psutil
+import tarantool
 import yaml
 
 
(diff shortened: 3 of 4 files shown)


test/utils.py
```diff
@@ -79,3 +81,24 @@ def wait_file(dir_name, file_pattern, exclude_list, timeout_sec=1):
...
+def kill_instance(instance_file_name):
+    # Kill tarantool instance by file config name.
+    pid_instance = subprocess.run(["pgrep", instance_file_name], capture_output=True)
+    if len(pid_instance.stdout) != 0:
+        os.kill(int(pid_instance.stdout), signal.SIGKILL)
+
+
+def find_free_port():
+    # Return free port in range [49152, 65535], else return None.
+    for port in range(49152, 65536):
```
Totktonada, line 95: It is from the local port range:

```sh
$ cat /proc/sys/net/ipv4/ip_local_port_range
32768	60999
```

From the kernel [documentation](https://www.kernel.org/doc/Documentation/networking/ip-sysctl.rst):

```
ip_local_port_range - 2 INTEGERS
	Defines the local port range that is used by TCP and UDP to
	choose the local port. The first number is the first, the
	second the last local port number.
	If possible, it is better these numbers have different parity
	(one even and one odd value).
	Must be greater than or equal to ip_unprivileged_port_start.
	The default values are 32768 and 60999 respectively.
```

In other words, when a client calls `connect()` (without prior explicit `bind()`, it is usual for a client to do just `connect()`), it'll choose a port from this range. The following sequence of events is possible so:

* You checked a port, it is free. You write it to a file.
* A client connects using TCP to some server.
* You attempt to listen on this port, but it is already stolen.

Unlikely you'll see this situation in a single thread testing, but if we'll use `xdist` in a future, it is possible. We observed the similar problem in test-run (https://github.com/tarantool/test-run/issues/141), where the port assignment mechanism is quite similar (except that we pass the address to tarantool using an environment variable).

I have a proposal how to stick with bind-to-zero-and-listen approach even on 1.10, will write below.

test/utils.py
```diff
@@ -79,3 +81,24 @@ def wait_file(dir_name, file_pattern, exclude_list, timeout_sec=1):
...
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        try:
+            sock.bind(("0.0.0.0", port))
+            sock.close()
+            return port
+        except Exception:
+            pass
```
Totktonada, line 102: BTW, I would highlight a problem I meet a long time ago on Travis CI: https://github.com/tarantool/test-run/commit/be4ca7dad54a3028d36e87cd1fca9bd0d0da4d4a. I can't precisely describe this old situation, but what I remember:

* IPv6 stack was enabled, but was not workable.
* I decided to consider a port as in use if it is in use either in IPv4 or in IPv6 as extra protection from port clashing errors. I meet the similar problem later in https://github.com/tarantool/test-run/pull/197.

As a general recommendation: always keep in the mind, whether you're working with IPv4, IPv6 or both in a particular place of the code. Prefer to pass addresses, not just ports to avoid ambiguity. Aware that `getaddrinfo()` may return several addresses and if both IPv4 and IPv6 ports have a listening server, the choice is arbitrary (but usually IPv6). So set host/ip/protocol explicitly, where applicable.

Totktonada:
We can find the listening ip:port on tarantool 1.10 using the following snippet:

```lua
tarantool> box.cfg({listen = '127.0.0.1:0'})
tarantool> do local res = {} for fd = 0, 65535 do local addrinfo = require('socket').internal.name(fd) if addrinfo ~= nil and addrinfo.host == '127.0.0.1' and addrinfo.family == 'AF_INET' and addrinfo.type == 'SOCK_STREAM' and addrinfo.protocol == 'tcp' and type(addrinfo.port) == 'number' then table.insert(res, addrinfo) end end if #res ~= 1 then error(('zero or more than one listening TCP sockets: %d'):format(#res)) end return ('%s:%s'):format(res[1].host, res[1].port) end
---
- 127.0.0.1:44865
...
```

Formatted for readability:

```lua
local socket = require('socket')

local res = {}
for fd = 0, 65535 do
    local addrinfo = socket.internal.name(fd)
    -- Assume that the socket listens on 127.0.0.1.
    local is_matched = addrinfo ~= nil and addrinfo.host == '127.0.0.1' and
        addrinfo.family == 'AF_INET' and addrinfo.type == 'SOCK_STREAM' and 
        addrinfo.protocol == 'tcp' and type(addrinfo.port) == 'number'
    if is_matched then
        table.insert(res, addrinfo)
    end
end
if #res ~= 1 then
    error(('zero or more than one listening TCP sockets: %d'):format(#res))
end
return ('%s:%s'):format(res[1].host, res[1].port) 
```

(It works, when there are no other listening TCP socket on 127.0.0.1. I would use it right after `box.cfg({listen = '127.0.0.1:0'})`. I would also highlight that it is preferable to bind on the local interface rather than on `0.0.0.0` or `::`. I would also stick with `127.0.0.1` rather than `localhost`, because the later is `127.0.0.1` or `::1`, which may be a source of problems. But if you want to support IPv6, take care to the address formatting -- the host should be in square brackets for IPv6.)

Please, use it only if `box.info.listen` gave `nil`. This way the snippet will run only on the old tarantool versions and unlikely will lead to problems due to changed internal API.

We can implement the same using `ffi` + `getsockname()` + `getnameinfo()`, but it'll require more code and will require to bother about closs platform compatibility.

test/utils.py
GRISHNOV: Thanks for the feedback! Now I have made an implementation using `box.cfg({listen = '127.0.0.1:0'})`. The port is obtained by `box.info.listen` or (if `box.info.listen = nil`) with using snippet below. Pytest gets binded port from file inside temp dir during testing.

test/utils.py
Totktonada: It seems, it is irrelevant now.

Totktonada:
I see that you implemented the idea with binding on zero port, thanks!

I would prefer to don't review the rest of the pull request, considering that there are two reviewers assigned. Feel free engage me into a discussion if there is a certain topic, where you want my participation.

0x501D (approved):
After a detailed analysis of the PR, it turned out that improvements were needed.

Please split in two commits:
* get free port number
*  cleanup instances

test/utils.py
```diff
@@ -79,3 +80,10 @@ def wait_file(dir_name, file_pattern, exclude_list, timeout_sec=1):
         iter_count = iter_count + 1
 
     return ""
+
+
+def kill_instance(instance_file_name):
```
0x501D, line 85: Why not pid number?
We may want to run tests simultaneously in two different branches, for example

0x501D (changes requested):
The result should be a function that returns a free port, here you will have to copy the code from test to test. It is necessary to make a general function without subsequent duplication of code.

Totktonada:
I was a bit wrong in my snippet regarding sockets filtering: it catches both listening and client sockets. So it'll not work as expected if, say, we start an instance as a replica.

I guess there are different ways to check whether a socket is a listening socket. Say, we can execute `ioctl()` with `SIOCINQ` or `SIOCOUTQ` (see `man 7 tcp`), which gives an error for a listening socket, using FFI. I don't know anything about portability of this solution.

I looked around `src/lua/socket.{c,lua}` and found `socket_getsockopt()`, which allows to get tcp socket options (`man 7 tcp`). I think that we can just filter socket with `SO_REUSEADDR`, because AFAIU it only hsa meaning for listening sockets and because iproto sets it on the binary protocol socket.

```lua
tarantool> ffi = require('ffi')
tarantool> socket = require('socket')
tarantool> fd = <...>
tarantool> value = ffi.new('int[1]')
tarantool> len = ffi.new('size_t[1]', ffi.sizeof('int'))
tarantool> res = ffi.C.getsockopt(fd, socket.internal.SOL_SOCKET, socket.internal.SO_OPT[socket.internal.SOL_SOCKET].SO_REUSEADDR.iname, value, len)
tarantool> value[0]
```

See `socket_getsockopt()` from `src/lua/socket.lua` for proper error checking.

(We can reuse `<socket object>:getsockopt(<...>)` without repeating its code, but it'll require wrapping of a file descriptior into a Lua socket object, which, I guess, would look even more complicated than direct FFI call above.)

test/utils.py
GRISHNOV: Done. Now it uses kill by PID

0x501D (approved):
LGTM

test/utils.lua
```diff
@@ -0,0 +1,75 @@
...
+        -- Filter socket with SO_REUSEADDR.
+        local reuseaddr = ffi.C.getsockopt(
+            address.fd,
+            socket.internal.SOL_SOCKET,
+            socket.internal.SO_OPT[socket.internal.SOL_SOCKET].SO_REUSEADDR.iname,
+            value, len
+        )
+        if reuseaddr == 0 then
+            return ('%s:%s'):format(res[key].host, res[key].port)
+        end
```
Totktonada, line 46: Zero return value of a `getsockopt()` call means successful acquiring of the option value. It doesn't state anything about the option value itself: it is stored in `value[0]`.

Totktonada, line 46: It is possible that the process has several SO_REUSEADDR sockets. I think it is good to detect this situation and raise an error expicitly rather than choose an arbirtary address.

test/utils.lua
```diff
@@ -0,0 +1,75 @@
...
+    end
+
+
+    error('all detected TCP sockets is not SO_REUSEADDR')
+end
+
+function testutils.dump_bound_socket_addr(instance_file_name)
+    -- Dump bound socket address to file for pytest.
+    -- It will be regular text file which name is 'instance_file_name.port'.
+    local file, err = io.open(instance_file_name .. '.port', 'w')
```
Totktonada, line 56: Nit: If it is an address, why the file is named `<...>.port`?

test/utils.py
```diff
@@ -79,3 +80,47 @@ def wait_file(dir_name, file_pattern, exclude_list, timeout_sec=1):
...
+    """
+
+    # Copy the instance config files to the run pytest tmpdir directory.
+    shutil.copy(instance_cfg_file_dir + "/" + instance_cfg_file_name, tmpdir)
+    # Copy the lua module with the auxiliary functions required by the instance config file.
+    shutil.copy(instance_cfg_file_dir + "/../../../" + "utils.lua", tmpdir)
+    # Start test instance for testing.
+    instance_obj = subprocess.Popen(["tarantool", instance_cfg_file_name], cwd=tmpdir)
+    # The delay is needed so that the instance has time to start and configure itself.
+    time.sleep(1)
```
Totktonada, line 128: Yeah, not a nice thing.

It is better to spin in a loop around some condition: say, presence of the file with the address to connect. Or in a loop with attempt to read the file and connect to the address. I also suggest to spin here until a timeout: otherwise hard to debug 'hangs' may appear.

I can guide you around the similar code in test-run, which run, I guess, around 100k tests per day (so any race in it is pretty visible). You can glance over the code around https://github.com/tarantool/test-run/pull/348 and https://github.com/tarantool/test-run/pull/349 or ask me to present it.

test/utils.py
```diff
@@ -79,3 +80,47 @@ def wait_file(dir_name, file_pattern, exclude_list, timeout_sec=1):
...
+    # Read bound port of test instance from file in temp pytest directory.
+    with open(str(tmpdir) + '/' + instance_cfg_file_name + '.port', 'r') as file_with_address:
+        instance_address = file_with_address.read()
+    instance_port = instance_address.split(":")[-1]
```
Totktonada, line 132: Nit: Why should we strip the host to concatenate it back in the caller code? I would just pass an address.

test/utils.py
```diff
@@ -79,3 +80,47 @@ def wait_file(dir_name, file_pattern, exclude_list, timeout_sec=1):
...
+
+    return instance_obj, instance_port
```
Totktonada, line 134: Nit: I guess a class that incapsulates all this instance related data would suit your needs better than a tuple of seems-to-be-unrelated values.

You'll able to define convenient methods to make the caller code abstracted out from just `Popen`: say, `start()`, `stop()` and so on. You likely will want to wait starting of the instance in `start()` -- it is some code around `Popen`, and likely will want to wait until an instance is stopeed in `stop()` -- again, some code around sending a signal to the process.

test/integration/play/test_play.py
```diff
@@ -2,13 +2,27 @@
...
+# In case of unsuccessful completion of tests, tarantool test instances may remain running.
+# This is autorun wrapper for each test case in this module.
+@pytest.fixture(autouse=True)
+def kill_remain_instance_wrapper():
+    # Kill remain test instance if it is still running from past integration test.
+    kill_instances()
+    # Run test.
+    yield
+    # Kill a test instance if it was not stopped due to a failed test.
+    kill_instances()
```
Totktonada, line 23: AFAIU, the second `kill_instances()` makes the first one no-op. So, maybe, it would be better to just assert that there are no 'orphan' instances?

Otherwise it looks like:

```python
if x:
    if x:  # For extra sure.
        foo()
```

test/utils.lua
GRISHNOV: Done, now checks `value[0]`

GRISHNOV: Done

test/utils.lua
GRISHNOV: Done, now only the port is written to the file, without `127.0.0.1:*`

test/utils.py
GRISHNOV: Done, now only the port is written to the file, without 127.0.0.1:*. 
Related to [this](https://github.com/tarantool/tt/pull/112#discussion_r975696355)

test/integration/play/test_play.py
GRISHNOV: Done, now it works only before the `yield `

test/utils.py
GRISHNOV: Done, implemented class `TarantoolTestInstance`

test/utils.py
```diff
@@ -82,6 +84,12 @@ def wait_file(dir_name, file_pattern, exclude_list, timeout_sec=1):
     return ""
 
 
+def kill_instances():
+    # Kill tarantool instances that were started as children from current process.
+    for child in psutil.Process().children(recursive=True):
+        os.kill(int(child.pid), signal.SIGKILL)
```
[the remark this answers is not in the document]
GRISHNOV: Done, implemented 3 sec timeout with `psutil.pid_exists(child.pid) or child.status() == 'zombie'` check

test/utils.py
GRISHNOV: Done, implemented 3 sec timeout with existence of port file check

test/utils.lua
```diff
@@ -0,0 +1,79 @@
...
+            addrinfo.family == 'AF_INET' and addrinfo.type == 'SOCK_STREAM' and
+            addrinfo.protocol == 'tcp' and type(addrinfo.port) == 'number'
+        if is_matched then
+            addrinfo.fd = fd
+            table.insert(res, addrinfo)
+        end
+    end
+    if #res ~= 1 then
+        error(('zero or more than one listening TCP sockets: %d'):format(#res))
+    end
```
Totktonada, line 31: Here we have zero or more than one listening **or client** sockets. It is okay to have many client sockets, for example if you are bootstrapping an instance from a remote master. The code below should filter out client sockets and leave only listening ones. If there are 0 or several **listening** sockets, we don't know which one is iproto's one. That's why the error should be raised.

test/utils.lua
Totktonada: Not precisely. I followed this discussion [here](https://github.com/tarantool/tt/pull/112#discussion_r989257844).

test/utils.lua
```diff
@@ -0,0 +1,79 @@
...
+
+    -- We need only listening, not client sockets.
+    local value = ffi.new('int[1]')
+    local len   = ffi.new('size_t[1]', ffi.sizeof('int'))
+    local level = socket.internal.SOL_SOCKET
+    -- Filter socket with SO_REUSEADDR.
+    local reuseaddr = ffi.C.getsockopt(
```
Totktonada, line 38: Why the `getsockopt()` success/error value is named `reuseaddr`? It is not a `reuseaddr` value and we already discussed it [here](https://github.com/tarantool/tt/pull/112#discussion_r975694086).

test/utils.py
```diff
@@ -79,3 +80,83 @@ def wait_file(dir_name, file_pattern, exclude_list, timeout_sec=1):
...
+        popen_obj = subprocess.Popen(["tarantool", self._instance_cfg_file_name], cwd=self._tmpdir)
+
+        # Waiting 3 seconds for instance configure itself and dump bound port to temp file.
+        deadline = time.time() + 3
+        while time.time() < deadline:
+            if os.path.exists(str(self._tmpdir) + '/' + self._instance_cfg_file_name + '.port'):
+                break
+            time.sleep(0.1)
+        else:
+            raise ValueError('Could not find a file with an instance bound port')
```
Totktonada, line 164: AFAIU, `RuntimeError()` is more appropriate for things out of a programmer control (when something external fails).

test/utils.py
```diff
@@ -79,3 +80,83 @@ def wait_file(dir_name, file_pattern, exclude_list, timeout_sec=1):
...
+
+        # Read bound port of test instance from file in temp pytest directory.
+        file_with_port_path = str(self._tmpdir) + '/' + self._instance_cfg_file_name + '.port'
```
Totktonada, line 167: Nit: `str(self._tmpdir) + '/' + self._instance_cfg_file_name + '.port'` is written twice. I would store it in a variable.

test/utils.py
Totktonada: It seems, you prefer to operate everywhere by ports, not addresses. It is okay too.

test/utils.py
```diff
@@ -4,6 +4,7 @@
 import subprocess
 import time
 
+import psutil
```
Totktonada, line 7: It is not shipped together with Python, it is a separate module. So it should be added to `test/requirements.txt`.

test/utils.py
```diff
@@ -82,6 +83,22 @@ def wait_file(dir_name, file_pattern, exclude_list, timeout_sec=1):
     return ""
 
 
+def kill_instances():
+    # Kill tarantool instances that were started as children from current process.
+    children = psutil.Process().children(recursive=True)
+    for child in children:
+        os.kill(int(child.pid), 0)
```
Totktonada, line 90: Do you have it added to raise `ProcessLookupError` if the process does not exists? I don't understand this logic. You already perform the `pid_exists()` check below (it is just wrapper around sending a zero signal, at least on Unix).

test/utils.py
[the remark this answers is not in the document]
Totktonada: This check looks okay on a brief glance, but I don't understand why `os.kill(pid, 0)` is also added, see [here](https://github.com/tarantool/tt/pull/112#discussion_r989359121).

test/utils.py
```diff
@@ -79,3 +80,83 @@ def wait_file(dir_name, file_pattern, exclude_list, timeout_sec=1):
...
+
+        self.popen_obj = popen_obj
+        self.port = instance_port
+
+    def stop(self):
+        """Stop tarantool test instance by popen.kill call.
+
+        Popen.kill uses SIGKILL for stop instance.
+        """
+        self.popen_obj.kill()
```
Totktonada, line 179: In test-run we had races of the following kind:

* Run server (in directory X).
* Perform a test.
* Stop server (send a signal).
* Run server (in directory X): error. The previous server has not been stopped yet.

So I think that a stop function should wait for stopping (till a timeout).

GRISHNOV: Done

test/utils.py
GRISHNOV: Done

test/utils.py
GRISHNOV: Done

test/utils.py
GRISHNOV: [Done](https://github.com/tarantool/tt/blob/igrishnov/gh-71-free-port-autoresolve-for-tests/test/requirements.txt#L7)

test/utils.py
GRISHNOV: Ouch, here I wanted to use not a zero signal, but `signal.SIGTERM`

test/utils.py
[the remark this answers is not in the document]
GRISHNOV: Fixed with `os.kill(int(child.pid), signal.SIGTERM)`

test/utils.py
Totktonada: I overlooked it somehow, sorry.

test/utils.py
```diff
@@ -83,6 +84,23 @@ def wait_file(dir_name, file_pattern, exclude_list, timeout_sec=1):
...
+        # Waiting for the completion of the process with 3 second timeout.
+        deadline = time.time() + 3
+        while time.time() < deadline:
+            if not psutil.pid_exists(child.pid) or child.status() == 'zombie':
+                # There is no more process with child.pid or it is zombie.
+                break
+            else:
+                time.sleep(0.1)
+        else:
+            raise RuntimeError("PID {} couldn't stop after receiving SIGKILL".format(child.pid))
```
Totktonada, line 101: > after receiving SIGKILL

But the function sends SIGTERM above.

test/utils.py
Totktonada: Now I see the logic, thanks.

test/utils.py
GRISHNOV: sorry, typo, done with `SIGKILL`

test/utils.lua
GRISHNOV: Done, now value is named `status`

test/utils.lua
GRISHNOV: Done, changed error msg `zero or more than one listening TCP sockets:` -> `zero or more than one listening or client TCP sockets:`

Totktonada: But why? Why not to collect sockets, shrink them to listening ones (strictly speaking, to ones with SO_REUSEADDR option set) and only then check that there is only one such socket?

GRISHNOV: Done, now code corresponds to the described sequence

Totktonada: Nice, thanks!

Totktonada (approved):
I glanced over the code once again (not really deeply) and it looks okay.

PR #116, opened 2022-08-10
ci: fix no line numbers shown for golangci

0x501D:
Solution is taken from:
golangci/golangci-lint-action#119 (comment)


0x501D:
Test commit was shown with line numbers and file name:
https://github.com/tarantool/tt/runs/7765606513?check_suite_focus=true

LeonidVas (approved):
Hi!
If it works fine. I'm ok.

PR #118, opened 2022-08-11
tt: cli language refactoring

### LINKED ISSUE #63: Format the command descriptions in `tt` according to the command line interface language
I'm not so good in command-line interface description language, but seems that the generally accepted standard is not always followed in `tt help'.

Need checks this guess and fix the typos.

GRISHNOV:
Refactoring of cli language that describe usage of tt commands.

Now a single notation is used to specify positional arguments through brackets `<>`, and 3 points `...` is also used to indicate the possibility of transmitting multiple parameters.

Fixed inaccuracies in the description of some parameters that must be not optional, but required.

Close #63


0x501D (approved):
LGTM

PR #140, opened 2022-09-09
create: fix template file name rendering

psergee:
If a file name is in <name>.tt.template format, it is renamed to the <name>. The issue is: if <name> also contains template variables like {{.var}} the instantiation engine tries to instantiate old file name (<name>.tt.template) and rename it. But this file does not exists after first renaming. So there is "no such file or directory error". The fix is to update file name after the first renaming.
Covered by unit test.


LeonidVas (approved):
Hi! Thank you for the patch. Please insert the PR message in the commit message.

Issue #125, opened 2022-08-30
Use `bin_dir` from `tarantool.yaml` for search the binaries 


LeonidVas:
Now the "root" directory (the directory with `tarantool.yaml`) is used to search for "local" binary files `tt` and `tarantool`. Instead, binary files should be searched in the `bin` directory.

0x501D:
Fixed in a9f774d

PR #141, opened 2022-09-12
tt: fix failure on non-executable files/dirs

psergee:
tt searches for tt & tarantool binaries in an environment root dir. If it finds tt or tarantool files there and they are not executables, tt launch fails with the "permission denied" error. The solution is to warn about non-executables, but do not fail.

### DIFF
--- a/cli/configure/configure.go
+++ b/cli/configure/configure.go
@@ -2,13 +2,13 @@ package configure
 
 import (
 	"fmt"
-	"log"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
 	"syscall"
 
+	"github.com/apex/log"
 	"github.com/mitchellh/mapstructure"
 	"github.com/spf13/cobra"
 	"github.com/tarantool/tt/cli/cmdcontext"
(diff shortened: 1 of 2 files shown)


cli/configure/configure.go
```diff
@@ -2,13 +2,13 @@ package configure
...
 import (
 	"fmt"
-	"log"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
 	"syscall"
 
+	"github.com/apex/log"
```
LeonidVas, line 11: Change used `log` library by separate commit.

cli/configure/configure.go
```diff
@@ -252,16 +250,16 @@ func configureLocalCli(cmdCtx *cmdcontext.CmdCtx, launchDir string) error {
...
-					`Found tt binary in local directory "%s" isn't executable: %s`, launchDir, err)
-			}
-
-			// We are not using the "RunExec" function because we have no reason to have several
-			// "tt" processes. Moreover, it looks strange when we start "tt", which starts "tt",
-			// which starts tarantool or some external module.
-			err = syscall.Exec(localCli, append([]string{localCli}, os.Args[1:]...), os.Environ())
-			if err != nil {
-				return err
+				log.Warnf(`"%s" isn't executable: %s`, localCli, err)
```
LeonidVas, line 253: I don't understand 2 points:
- why is it bad to fail?
- If I understand correctly, before the changes, it should print an error that this is not an executable file, go fix it. Or am I missing something?

LeonidVas (changes requested):
Hi! Thank you for the patch.
See several comments bellow.

LeonidVas:
After discussion, it was decided that whatever we wanted would be fine when the https://github.com/tarantool/tt/issues/125 will done.

PR #143, opened 2022-09-12
cartridge cli: improved patching system

0x501D:
The patching system can now distinguish between an error and the fact that a patch has already been applied.

Also, the output is now printed only after the patch has been applied.


LeonidVas (approved):
LGTM.

PR #144, opened 2022-09-13
build: added build module

### LINKED ISSUE #40: Implement `tt build` module
The module will be used to build an application for local development.

psergee:
Added build module with local build implementation. --spec option is supported to specify rockspec file to use.

Closes #40

### DIFF
--- a/cli/util/execute.go
+++ b/cli/util/execute.go
@@ -1,0 +1,164 @@
+package util
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/mattn/go-isatty"
+
+	"github.com/apex/log"
+	"github.com/briandowns/spinner"
+)
+
+type emptyStruct struct{}
+
+// readyChan is a channel used to signal completion of command execution.
+type readyChan chan emptyStruct
+
+// execOwnerPerm is a bitmask to check owner exec bit.
+const execOwnerPerm uint32 = 0100
+
+var (
+	spinnerPicture    = spinner.CharSets[9]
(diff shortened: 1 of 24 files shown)


cli/util/execute.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/mattn/go-isatty"
+
+	"github.com/apex/log"
+	"github.com/briandowns/spinner"
+)
+
+type emptyStruct struct{}
+
+// readyChannel is a channel used to signal completion of command execution.
+type readyChannel chan emptyStruct
```
vr009, line 23: I would rename it to readyChan, sounds clear and more compactly and similar with chan type. But I don't insist, just idea. Feel free to ignore.

vr009 (approved):
LGTM. Some moments were discussed by voice.

cli/util/execute.go
```diff
@@ -0,0 +1,164 @@
...
+
+// IsExecOwner checks if specified file has owner execute permissions.
+func IsExecOwner(path string) (bool, error) {
+	fileInfo, err := os.Stat(path)
+	if err != nil {
+		return false, err
+	}
+
+	perm := fileInfo.Mode().Perm()
+	return perm&execOwnerPerm != 0, nil
```
0x501D, line 152: Nit: Maybe we shoult add functions for bit manipulation to utils: like this:
```go
func BitSet(b, flag Bits) Bits    { return b | flag }
func BitClear(b, flag Bits) Bits  { return b &^ flag }
func BitToggle(b, flag Bits) Bits { return b ^ flag }
func BitHas(b, flag Bits) bool    { return b&flag != 0 }
```

psergee: I have added only bit set check function for 32-bit uints. Other functions are not required yet.

PR #147, opened 2022-09-15
connect: fix connect via URI

### LINKED ISSUE #138: Unable to connect via `tt connect URI`
```bash
$ ./tt connect --help
Connect to the tarantool instance

USAGE
  tt connect (<INSTANCE_NAME> | <URI>) [<FILE> | <COMMAND>] [flags]
  COMMAND | tt connect (<INSTANCE_NAME> | <URI>) [flags]
...
$ ./tt connect localhost:3013
   ⨯ Can't find an application init file: stat /home/oleg/src/git/tt/localhost.lua: no such file or directory
```

After the commit:

016b827fcf4ccaf0c97abc68c570ad568c677aef

oleg-jukovec:
The patch fixes connection to a Tarantool instance via URI. The issue was introduced with a pull request to work with a set of intances [1]. The rules for determining the URI are similar to used by a connector [2].

1. https://github.com/tarantool/tt/pull/113
2. https://github.com/tarantool/tt/blob/da319ff9a1d0d7bd12e6f682d6ef906c965dbfa5/cli/connector/conn_opts.go#L47-L69

Closes #138

I was looking at the connector code to determine is a string a URI. But I want to suggest making rules to detect valid URI for `tt connect` more strict:

1. Left only `tcp://host:numeric_port` or `host:numeric_port` for tcp-socket connections.
2. Left only `unix://path/to/file` for unix-socket connections.

From my point of view it will simplify both our and users lives.

### DIFF
--- a/cli/cmd/connect.go
+++ b/cli/cmd/connect.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"regexp"
 	"syscall"
 
 	"github.com/apex/log"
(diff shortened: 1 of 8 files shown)


cli/cmd/connect.go
```diff
@@ -44,25 +46,41 @@ func NewConnectCmd() *cobra.Command {
 	return connectCmd
 }
 
+// isURI returns true if a string is a valid URI.
+func isURI(str string) bool {
+	uriRe := regexp.MustCompile("^((\\w+):(\\/+)?)?([\\w.-]+):(\\d+)$")
```
0x501D, line 51: Why it was impossible to use only one regexp?

oleg-jukovec: I thought it would be too complicated.

But it doesn't look too scary. Thank you, updated.

0x501D (approved):
LGTM

GRISHNOV (approved):
Hi! LGTM

PR #163, opened 2022-09-29
create: instantiate app in temporary directory

psergee:
tt create will overwrite target application directory if
--force flag is set and the target directory already exists.
This is incorrect behavior if app creation is interrupted
by user or error. The solution is to use temporary directory
for app instantiation and move in the last step.


vr009 (approved):
Everything looks clear to me. LGTM

0x501D (approved):
LGTM

PR #168, opened 2022-10-03
rocks: add added path to header files

0x501D:
Added environment variable `TT_CLI_TARANTOOL_INCLUDE`. It stores the path to the header files supplied with the tarantool. This path is needed to build modules from rocks that require tarantool header files.

tt will automatically determine the path to the header files at runtime and will set the environment variable as a variable.


cli/util/util.go
```diff
@@ -284,6 +285,44 @@ func GetTarantoolVersion(cli *cmdcontext.CliCtx) (string, error) {
 	return cli.TarantoolVersion, nil
 }
 
+// GetTarantoolIncludePath returns and caches a directory with
+// header files supplied with tarantool.
+func GetTarantoolIncludePath(cli *cmdcontext.CliCtx, cliOpts *config.CliOpts) (string, error) {
```
psergee, line 290: Unit test for this functions is missing.

0x501D: This function This function depends on runtime and I don't know how to correctly write unit tests for it.

cli/util/util.go
```diff
@@ -284,6 +285,51 @@ func GetTarantoolVersion(cli *cmdcontext.CliCtx) (string, error) {
 	return cli.TarantoolVersion, nil
 }
 
+// SetupTarantoolPrefix defines the installation prefix and the path to the tarantool header files.
+func SetupTarantoolPrefix(cli *cmdcontext.CliCtx, cliOpts *config.CliOpts) error {
```
psergee, line 289: Unit-test is missing.

cli/util/util.go
psergee: Just fill CliCtx and CliOpts with the data for test and call the function. If there is executable to be run, generate a shell script with the specified output, set TarantoolExecutable to this script and call the function.

cli/util/util.go
0x501D: Added

psergee (approved):

PR #171, opened 2022-10-05
install: file name format changed

0x501D:
Old format `program`=`version` resulted in escaping with quotes when working with a file from a shell.

New format is `program`_`version`.

### DIFF
--- a/cli/search/search.go
+++ b/cli/search/search.go
@@ -20,6 +20,13 @@ const (
 	GitRepoTT        = "https://github.com/tarantool/tt.git"
 )
 
+const (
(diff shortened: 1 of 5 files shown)


cli/search/search.go
```diff
@@ -20,6 +20,11 @@ const (
 	GitRepoTT        = "https://github.com/tarantool/tt.git"
 )
 
+const (
+	VersionCliSeparator = "="
```
psergee, line 24: Please add comments for these constants.

psergee (approved):

cli/search/search.go
0x501D: Added

PR #179, opened 2022-10-10
tt: support application in current dir

psergee:
Added support of start/stop application in current directory. If application name passed to start/stop command is equal to current directory name, current directory is considered as application.


0x501D (approved):
LGTM

PR #185, opened 2022-10-17
pack: fix empty app pack crash

psergee:
Here is the result of empty app integration test:
```
  • Apps to pack: empty_app  
2022/10/17 12:23:12 Whoops! It looks like something is wrong with this version of Tarantool CLI.
Error: Unhandled internal error: runtime error: invalid memory address or nil pointer dereference
Version: Tarantool CLI version 0.1.0, linux/amd64. commit: 2520a4f
```

If empty directory name is specified in app-list option, this app list is not checked if the specified app names are real applications. The solutions is to add these check and skip directories that are not apps.

### DIFF
--- a/cli/pack/common.go
+++ b/cli/pack/common.go
@@ -2,6 +2,11 @@ package pack
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+
 	"github.com/apex/log"
 	"github.com/otiai10/copy"
 	"github.com/tarantool/tt/cli/build"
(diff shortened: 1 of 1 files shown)


cli/pack/common.go
```diff
@@ -117,35 +117,47 @@ func prepareBundle(cmdCtx *cmdcontext.CmdCtx) (string, error) {
...
+	if packCtx.AppList == nil {
+		if packCtx.App.InstancesEnabled != "." {
 
-		excludeList, err := prepareExcludeList(ExcludeListExpressions)
-		if err != nil {
-			return "", fmt.Errorf("failed to compile regular expressions: %s", err)
-		}
+			if err != nil {
+				return "", fmt.Errorf("failed to compile regular expressions: %s", err)
+			}
```
vr009, line 131: That check must follow prepareExcludeList call.

PR #187, opened 2022-10-19
tt: run unit test in tmp dir

psergee:
Some unit tests rely on missing tt configs in parent directories.
To make sure developer config does not affect these unit tests, make
chdir to temporary directory for these tests.


cli/configure/configure_test.go
```diff
@@ -156,6 +156,13 @@ func TestDetectLocalTarantool(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, expected, cmdCtx.Cli.TarantoolExecutable)
 
+	// Chdir from tmp directory to avoid loading tarantool.yaml from parent directories.
```
0x501D, line 159: from?

LeonidVas (approved):

PR #192, opened 2022-10-24
install: fix tt install

0x501D:
Fix for ignoring tt installation version.

### DIFF
--- a/cli/install/install.go
+++ b/cli/install/install.go
@@ -448,13 +448,13 @@ func installTt(version string, binDir string, flags InstallationFlag, distfiles 
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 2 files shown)


cli/install/install.go
```diff
@@ -448,13 +448,13 @@ func installTt(version string, binDir string, flags InstallationFlag, distfiles
 			if err != nil {
 				return err
 			}
-			util.ExecuteCommand("git", flags.Verbose, logFile, path, "checkout", "master")
+			util.ExecuteCommand("git", flags.Verbose, logFile, path, "checkout", ttVersion)
```
psergee, line 451: Need to cover this with test to avoid such errors in future.

0x501D: Done

psergee (approved):

PR #200, opened 2022-10-25
ci: run flake8 linter separately

0x501D:
Flake8 tests were forgotten when implementing golangci-lint.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -84,11 +84,14 @@ jobs:
       - name: Codespell checks
         run: mage codespell
 
-      - name: Linter
+      - name: Go Linter
         uses: golangci/golangci-lint-action@v3
         with:
           args: --config=golangci-lint.yml --out-${NO_FUTURE}format colored-line-number
           skip-cache: true
+
+      - name: Python Linter
+        run: python3 -m flake8 test
 
       - name: Unit tests
         run: mage unitfull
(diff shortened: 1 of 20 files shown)


test/conftest.py
```diff
@@ -5,7 +5,6 @@
 
 import py
 import pytest
-
```
LeonidVas, line 8: It's not an error.
If flake8 reacts to this, it must be configured.

https://peps.python.org/pep-0008/#imports
```
Imports should be grouped in the following order:

1. Standard library imports.
2. Related third party imports.
3. Local application/library specific imports.
```

LeonidVas (changes requested):
Hi!
It's great that you're making our code base cleaner, thank you.
Several questions:

1. Will it use config https://github.com/tarantool/tt/blob/master/.flake8 ?
2. Can you also add  `go fmt` and `goimports` checks?
3. Do we now do a check for imports (`isort`) and how (https://github.com/tarantool/cartridge-cli/commit/a34758c4f964c2baa1c936c226dc6a062b215bb8#diff-9c592cb90c321bcda4069ebf13a7b7f6c666f931a61fd01b98d00631dd113faaR8) ?

test/conftest.py
0x501D: Fixed with .isort.conf:
`[settings]
known_local_folder=utils`

0x501D: 1. Will it use config master/.flake8 ? - Yes
2. gofmt added (goimports is buggy)
3. isort is checked on CI.

LeonidVas (approved):
LGTM

PR #204, opened 2022-10-27
init: add check for tt env config existence

psergee:
If tt config already exists, init command will ask for confirmation to overwrite it.

### DIFF
--- a/cli/cmd/init.go
+++ b/cli/cmd/init.go
@@ -39,5 +39,7 @@ func internalInitModule(cmdCtx *cmdcontext.CmdCtx, args []string) error {
 		SkipConfig: skipConfig,
 	}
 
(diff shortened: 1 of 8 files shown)


LeonidVas (changes requested):
"init: added check for tt env config existence" -> "init: add check for tt env config existence"
"init: added force flag to re-write config" -> "init: add force flag to re-write config"
> Description is what the patch does, started from lowercase letter, without a dot in the end, in the imperative mood.

https://github.com/tarantool/tarantool/wiki/Code-review-procedure#commit-message

If tarantool.yml is present, it should also count as tarantool.yaml (Warn about exists and rewrites if necessary).

0x501D (approved):
LGTM

LeonidVas (approved):
LGTM

PR #206, opened 2022-10-28
tt: remove extract module and its deps

psergee:
Some fixes are made in existing extract tgz API:

- Added symlinks support
- Added file permissions support.


LeonidVas (approved):
LGTM

PR #208, opened 2022-10-31
tt: support yaml/yml config file extensions

psergee:


LeonidVas (changes requested):
Hi! Thank you for the patch.
AFAIU if we find both (`tarantool.yaml` and `tarantool.yml`), we go further searching (in the parent directory). I think it is better to return an error (with description) in this case.

psergee:
> Hi! Thank you for the patch. AFAIU if we find both (`tarantool.yaml` and `tarantool.yml`), we go further searching (in the parent directory). I think it is better to return an error (with description) in this case.

Good catch. Fixed, and unit-test is added for getConfigPath function.

PR #220, opened 2022-11-10
test: fix cartridge test

### LINKED ISSUE #128: Tests: cleanup running tarantool instances if test failed.
At the moment, if the cartridge tests fail (test/integration/cartridge/test_cartridge.py), then running instances of tarantool remain.
Before restarting the tests, you must manually shut down these services. It is necessary to automatically shut down tarantool instances in case of test failures. It is also necessary to check the rest of the integration tests for the presence of this problem.

better0fdead:
Added stop of instance after finish of the test, if instance is still running.
Also looked through other test to check for the same behavior.

Closes #128

### DIFF
--- a/test/integration/cartridge/test_cartridge.py
+++ b/test/integration/cartridge/test_cartridge.py
@@ -2,12 +2,29 @@ import re
 import subprocess
 import time
 
+import pytest
+
 from utils import run_command_and_get_output, wait_file
+
+cartridge_name = "test_app"
+
+
+@pytest.fixture(autouse=True)
+def stop_cartridge_app(tt_cmd, tmpdir):
+    # Run test.
+    yield
+    # Stop instance if it is still running.
+    status_cmd = [tt_cmd, "status", cartridge_name]
+    status_rc, status_out = run_command_and_get_output(status_cmd, cwd=tmpdir)
+    assert status_rc == 0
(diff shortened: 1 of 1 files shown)


test/integration/cartridge/test_cartridge.py
```diff
@@ -2,12 +2,29 @@
...
+import pytest
+
 from utils import run_command_and_get_output, wait_file
 
+cartridge_name = "test_app"
+
+
+@pytest.fixture(autouse=True)
+def stop_cartridge_app(tt_cmd, tmpdir):
+    # Run test
```
0x501D, line 14: Missing dot at the end of a sentence.

test/integration/cartridge/test_cartridge.py
```diff
@@ -2,12 +2,29 @@
...
+    yield
+    # Stop instance if it is still running
```
0x501D, line 16: Missing dot at the end of a sentence.

test/integration/cartridge/test_cartridge.py
better0fdead: Added

test/integration/cartridge/test_cartridge.py
better0fdead: Added

PR #224, opened 2022-11-11
test: move long running tests to separate workflow

### LINKED ISSUE #154: Move long-running tests to a separate workflow
The average test execution time is about seven minutes, then it will become longer.
 We need to move move long-running tests to a separate workflow (full-ci label).

psergee:
Setup-python and cache actions versions were updated to fix warnings about using deprecated functionality.

Closes #154


0x501D (approved):
LGTM

PR #226, opened 2022-11-15
doc: update cartridge app creation example

psergee:
Add vshard bootstrap command as a final step for cartridge app creation example.


0x501D (approved):
LGTM

LeonidVas (approved):

PR #228, opened 2022-11-16
ci: fix EE tests for outside pull requests

oleg-jukovec:
Such pull requests may be labeled with `ee-ci` or `full-ci` to run tests with Tarantool EE. To avoid security problems, the label must be reset manually for every run.


oleg-jukovec:
See #229

psergee (approved):

PR #235, opened 2022-11-18
ci: fix checkout with pull_request_target

oleg-jukovec:
It is required to set `ref` and `repository` param for checkout in case of `pull_request_target` because the target runs in the context of the base commit.


oleg-jukovec:
It now checkouts at the right commit:

Untitled

oleg-jukovec:
See #236

LeonidVas (approved):

PR #236, opened 2022-11-18
playground for #235

oleg-jukovec:
The pull_request_target works only on base branch GitHub Actions code. So we need an outside branch based on the https://github.com/tarantool/tt/pull/235 for tests .


oleg-jukovec:
It looks fine after the fix:

Untitled

oleg-jukovec:
It looks fine at now:

Untitled

PR #240, opened 2022-11-21
install: support new tag format

0x501D:
The tag format in tt will be changed from X.Y.Z to vX.Y.Z. The old format also needs to be supported. For example:

`tt install tt=0.2.0` should install v0.2.0 if such a version exists.

### DIFF
--- a/cli/install/install.go
+++ b/cli/install/install.go
@@ -391,11 +391,14 @@ func installTt(version string, binDir string, flags InstallationFlag, distfiles 
 	}
 
 	// Check that the version exists.
+	// The tag format in tt is vX.Y.Z, but the user can use the X.Y.Z format
+	// and this option needs to be supported.
 	if ttVersion != "master" {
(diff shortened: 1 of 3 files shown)


cli/install/install.go
```diff
@@ -391,11 +391,14 @@ func installTt(version string, binDir string, flags InstallationFlag, distfiles
 	}
 
 	// Check that the version exists.
+	// The tag format in tt is vX.Y.Z, but the user can use the X.Y.Z format
```
psergee, line 394: Should we add a note to the change log about this?

psergee (approved):

cli/install/install.go
0x501D: Good question. @LeonidVas what do you think?

LeonidVas: I don't think this is necessary because the user will not see these changes.

Issue #246, opened 2022-11-22
Minor changes after feedback on release 0.2.0


LeonidVas:
* remove `create` and `build` from `tt cartridge`
* update the error when executing `tt search tarantool-ee` without credentials
* update help of `tt run`
* add a confirmation request when executing `tt restart ...`
* rename `tt remove` to `tt uninstall`
* `tt connect` does not print the help when it fails. This situation must be fixed. Needs also check that it works for all other modules

PR #255, opened 2022-11-23
Minor changes for cartridge and restart

psergee:
- Added `tt restart` confirmation prompt. `-y` option is added to accept restart without prompting.
- `tt cartridge` sub-commands `create`, `build`, `pack` are removed. Some message from cartridge were added to `tt build` and `tt create`

Part of #246

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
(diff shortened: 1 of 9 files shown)


0x501D (approved):

Totktonada:
A general suggestion: introduce a variable similar to `DEBIAN_FRONTEND=noninteractive`, which lets the tool assume a default answer. It is highly desirable for scripting.

I would even say more. I saw so many problems with stuck `apt-get`, because `DEBIAN_FRONTEND=noninteractive` was forgotten, that something is definitely wrong with apt-get defaults. Maybe it doesn't lean on `isatty()`? Don't remember. I would only say that years of hitting by this problem made me sensitive to it.

psergee:
> 

Thanks for the suggestion! Created a task for it: #257

PR #266, opened 2022-11-29
status: make instance status colorful

psergee:
Screenshot from 2022-12-06 12-24-44

### DIFF
--- a/cli/cmd/status.go
+++ b/cli/cmd/status.go
@@ -41,7 +41,8 @@ func internalStatusModule(cmdCtx *cmdcontext.CmdCtx, args []string) error {
 
 	for _, run := range runningCtx.Instances {
 		fullInstanceName := running.GetAppInstanceName(run)
-		log.Infof("%s: %s", fullInstanceName, running.Status(&run))
+		procStatus := running.Status(&run)
+		log.Infof("%s: %s", procStatus.ColorSprint(fullInstanceName), procStatus.Text)
 	}
 
 	return nil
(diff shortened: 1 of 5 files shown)


0x501D:
non-standard unicode character. even pregmatapro doesn't draw it

0x501D:
NO_COLOR has no effect. We shouls support it.
https://no-color.org/

0x501D (changes requested):
I think it would be better to manipulate the color of the instance name than to use a unicode character, which is not supported by all fonts. And of course handle the NO_COLOR variable

psergee:
> 
> 
> non-standard unicode character. even pragmatapro doesn't draw it

Sure, I know :)

cli/cmd/status.go
```diff
@@ -39,7 +39,8 @@ func internalStatusModule(cmdCtx *cmdcontext.CmdCtx, args []string) error {
 	}
 
 	for _, run := range runningCtx.Instances {
-		log.Infof("%s: %s", run.InstName, running.Status(&run))
+		procStatus := running.Status(&run)
+		log.Infof("%c %s: %s", procStatus.Color, run.InstName, procStatus.Text)
```
AnaNek, line 43: Maybe we should just use the colorful output using a standard [package color](https://pkg.go.dev/github.com/fatih/color)?

psergee: This package is not standard.

0x501D (approved):

AnaNek (approved):

PR #294, opened 2022-12-16
doc: fix packing environments section

AnaNek:
Before this PR directory tree was displayed
incorrect in `Packing environments` section:

image


vr009 (approved):
Thank you!

PR #297, opened 2022-12-26
Add app template follow up message

psergee:
Added support of follow-up messages to make it simpler to start to work with recently created app.
```
$ ../tt create cartridge --name app1 -f
   • Creating application in /tmp/test/app1
   • Using built-in 'cartridge' template.
   • Application 'app1' created successfully
What's next?
Build and start 'app1' application:
    $ tt build app1
    $ tt start app1

To bootstrap vshard run the following command:
    $ tt cartridge replicasets setup --bootstrap-vshard --name app1 --run-dir var/run/app1
```

### DIFF
--- a/cli/templates/internal/engines/go_text_engine.go
+++ b/cli/templates/internal/engines/go_text_engine.go
@@ -4,10 +4,32 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"path/filepath"
 	"text/template"
 )
 
 type GoTextEngine struct {
+}
+
+var commonTemplateFuncs map[string]interface{}
+
+// relativeToCurrentWorkingDir returns a path relative to current working dir.
+// In case of error, fullpath is returned.
+func relativeToCurrentWorkingDir(fullpath string) string {
+	cwd, err := os.Getwd()
+	if err != nil {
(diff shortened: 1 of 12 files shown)


cli/templates/internal/engines/go_text_engine.go
```diff
@@ -4,12 +4,34 @@ import (
...
 )
 
 type GoTextEngine struct {
 }
 
+var commonTemplateFuncs map[string]interface{}
+
+// relativeToCurrentWorkingDir returns a path relative to current working dir.
+// In case of error, fullpath is returned.
+func relativeToCurrentWorkingDir(fullpath string) string {
```
LeonidVas, line 18: Hmm, I don't understand, "How does this change relate to the current commit?"

LeonidVas (commented):
Generally OK, see one question bellow.
Rebase the patchset onto master.

LeonidVas (approved):

cli/templates/internal/engines/go_text_engine.go
psergee: This is required to print more compact path in vshard bootstrapping command.

PR #301, opened 2022-12-28
Minor code refactor

psergee:
- use more restrictive interfaces in some functions in util package
- rename "engines" package to "templates"
- simplify reader code in create
- fix some staticcheck complaints in create module.


LeonidVas (approved):
LGTM.

Issue #264, opened 2022-11-29
Race condition between tt start and tt stop/restart


psergee:
There is a race condition between `tt start` & `tt stop/restart` commands.

`tt start --watchdog`, which is invoked by `tt start` contains the following main steps:
1. Create PID-file.
2. Start tarantool instance
3. Start signals handling loop

Signals handling loop is required to process interrupt signals and send them to a child Tarantool instance process. But there is a race condition here. If the system is under high load, a user can run `tt stop` between:
- 1 and 2 steps. This causes tt watchdog to be killed, but PID file is not removed and contains PID of the dead tt process.
- 2 and 3 steps. This causes Tarantool instance hanging, because interrupt signal is not passed to it by watchdog tt:
```
$ tt start app1:router && tt stop app1:router
   • Starting an instance [router]...
2022/11/29 18:08:16 The Instance (PID = 141846) has been terminated.
$ pidof tarantool
141856
```

To reproduce this issue manually, put `time.Sleep` right after PID file creation or after Tarantool instance start and invoke the command from the example above.

PR #304, opened 2023-01-09
test: fix flaky test

better0fdead:
Increased timeout for several tests.
Added retry for setup docker in macOs full-ci workflow.

### DIFF
--- a/.github/workflows/full-ci.yml
+++ b/.github/workflows/full-ci.yml
@@ -124,10 +124,14 @@ jobs:
           fetch-depth: 0
           submodules: recursive
 
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+
       - name: Install dependencies
         run: |
           brew install --overwrite go mage
-          brew install python3
           pip3 install pytest tarantool requests psutil pyyaml netifaces
 
       - name: Setup Docker
@@ -144,3 +148,9 @@ jobs:
 
       - name: Run integration tests
         run: mage integrationfull
+
+      - name: Uninstall tarantool
+        if: always()
+        run:  |
+          brew uninstall --force tarantool
+          rm -rf /usr/local/etc/tarantool
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -117,10 +117,14 @@ jobs:
           fetch-depth: 0
           submodules: recursive
 
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+
       - name: Install dependencies
         run: |
           brew install --overwrite go mage
-          brew install python3
           pip3 install pytest tarantool requests psutil pyyaml netifaces
 
       - name: Build tt
@@ -134,3 +138,9 @@ jobs:
 
       - name: Run integration tests
         run: mage integration
+
+      - name: Uninstall tarantool
+        if: always()
+        run: |
+          brew uninstall --force tarantool
+          rm -rf /usr/local/etc/tarantool
--- a/test/integration/restart/test_restart.py
+++ b/test/integration/restart/test_restart.py
@@ -60,7 +60,7 @@ def test_restart_with_auto_yes(tt_cmd, tmpdir_with_cfg):
         assert "Confirm restart of 'test_app' [y/n]" not in restart_output[0]
         assert "has been terminated" in restart_output[0]
         assert "Starting an instance" in restart_output[1]
-        wait_file(os.path.join(tmpdir_with_cfg, run_path, app_name), 'test_app.pid', [], 5)
+        wait_file(os.path.join(tmpdir_with_cfg, run_path, app_name), 'test_app.pid', [], 25)
 
         restart_output = app_cmd(tt_cmd, tmpdir_with_cfg, ["restart", "--yes", app_name], [])
         assert "Confirm restart of 'test_app' [y/n]" not in restart_output[0]
--- a/test/integration/running/test_running.py
+++ b/test/integration/running/test_running.py
@@ -3,6 +3,7 @@ import re
 import shutil
 import subprocess
 import tempfile
+import time
 
 import yaml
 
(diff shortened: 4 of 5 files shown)


better0fdead:
Not so sure about last commit, but it seems like best solution, I couldn't reproduce some flaky tests even under 95% of load on my personal mac. It seems like github runners are VERY slow for macOs.

0x501D (approved):

test/integration/daemon/test_daemon.py
```diff
@@ -269,6 +270,7 @@ def test_daemon_http_requests_with_cfg(tt_cmd, tmpdir_with_cfg):
     status_rc, status_out = utils.run_command_and_get_output(status_cmd, cwd=tmpdir)
     assert status_rc == 0
     assert re.search(r"RUNNING. PID: \d+.", status_out)
+    time.sleep(5)
```
LeonidVas, line 273: Why do we need to add a timeout here and why is it so long (5 seconds)?

test/integration/restart/test_restart.py
```diff
@@ -60,7 +60,7 @@ def test_restart_with_auto_yes(tt_cmd, tmpdir_with_cfg):
         assert "Confirm restart of 'test_app' [y/n]" not in restart_output[0]
         assert "has been terminated" in restart_output[0]
         assert "Starting an instance" in restart_output[1]
-        wait_file(os.path.join(tmpdir_with_cfg, run_path, app_name), 'test_app.pid', [], 5)
+        wait_file(os.path.join(tmpdir_with_cfg, run_path, app_name), 'test_app.pid', [], 25)
```
LeonidVas, line 63: The problem can't be related with https://github.com/tarantool/tt/issues/264 ?

test/integration/running/test_running.py
```diff
@@ -190,6 +191,7 @@ def test_clean(tt_cmd, tmpdir_with_cfg):
     stop_rc, stop_out = run_command_and_get_output(stop_cmd, cwd=tmpdir)
     assert stop_rc == 0
     assert re.search(r"The Instance test_app \(PID = \d+\) has been terminated.", stop_out)
+    time.sleep(5)
```
LeonidVas, line 194: Why do we need to add a timeout here?

test/integration/running/test_running.py
```diff
@@ -512,6 +514,12 @@ def test_no_args_usage(tt_cmd):
                              status_out)
             assert re.search(r"app2: logs has been rotated. PID: \d+.", status_out)
 
+            status_cmd = [tt_cmd, "status"]
```
LeonidVas, line 517: Not related with commit message?

test/integration/running/test_running.py
```diff
@@ -568,6 +576,9 @@ def test_running_env_variables(tt_cmd, tmpdir_with_cfg):
     # Check that log format is in json.
     isJson = False
     logPath = os.path.join(tmpdir, "var", "log", "test_env_app", "test_env_app.log")
+    logDir = os.path.join(tmpdir, "var", "log", "test_env_app")
```
LeonidVas, line 579: The process has been correctly terminated, but the log file was not created?

.github/workflows/full-ci.yml
```diff
@@ -131,6 +131,12 @@ jobs:
           pip3 install pytest tarantool requests psutil pyyaml netifaces
 
       - name: Setup Docker
+        id: setupDocker
+        continue-on-error: true
+        uses: docker-practice/actions-setup-docker@master
+
+      - name: Retry setup Docker
```
LeonidVas, line 138: It looks ugly, and there's no comment on why it's needed.
Why only one extra try?

.github/workflows/tests.yml
```diff
@@ -133,4 +134,4 @@ jobs:
         run: mage unit
 
       - name: Run integration tests
-        run: mage integration
+        run: python3 -m pytest --reruns 5 -m "not slow" test/integration
```
LeonidVas, line 137: I'm willing to believe that the MacOS ranners on the githab are very slow and I have 2 questions:
- how about testing on self-hosted runners
- at what points do the tests crash (I assume that at some timeouts)

LeonidVas (changes requested):
Hi. Thank you for the patchset.
See several comments bellow.

.github/workflows/full-ci.yml
better0fdead: Setup docker action fails often because it has 180 timeout for docker to bootstrap but it is sometimes not enough and it fails in the middle of an action. Second action skips previous steps and waits for another 180 seconds and continues setup. In the repo of an action there are issues about that problem but dev. doesn't seem to resolve them in near future.

test/integration/running/test_running.py
better0fdead: Yes :), I have no clue how it is even possible and couldn't reproduce it even under heavy stress test on my working machine.

test/integration/restart/test_restart.py
better0fdead: Yes, it can actually be related.

test/integration/running/test_running.py
better0fdead: Sometimes app on ci doesn't stop in time after log message. Probably somehow related with #264. Will wait for fix and rebase.

.github/workflows/tests.yml
better0fdead: Tests crash after 5 reruns of failed tests.

better0fdead: Deleted

.github/workflows/full-ci.yml
better0fdead: Will be fixed after switching to self-hosted runners

test/integration/running/test_running.py
better0fdead: Deleted

test/integration/daemon/test_daemon.py
better0fdead: Deleted

PR #307, opened 2023-01-18
tt: disable ssl for lint and unit builds

psergee:
mage lint and mage unit fails without pkg-config tool. This tool is required to build go-openssl. tt is built with disabled ssl, so it must be disabled for lint and unit tests also.

### DIFF
--- a/golangci-lint.yml
+++ b/golangci-lint.yml
@@ -1,5 +1,7 @@
 run:
   timeout: 3m
+  build-tags:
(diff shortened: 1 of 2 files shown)


LeonidVas:
In the future we should solve the ssl problem and add it (https://github.com/tarantool/tt/issues/308).

LeonidVas (approved):

0x501D (approved):

Issue #286, opened 2022-12-09
Replace the `tarantoolctl` with tt


LeonidVas:
Now in packages we deploy tarantoolctl with tarantool as the cli. In this task it is necessary to:
- check that tt can replace tarantoolctl
- add tarantool's dependency on tt

PR #322, opened 2023-01-26
doc: add tarantoolctl/cartridge to tt transition notes

psergee:
Part of #286

### DIFF
--- a/doc/examples.rst
+++ b/doc/examples.rst
@@ -389,3 +389,244 @@ Flag without argument:
    http://127.0.0.1:1024/tarantool
    {"res":"Tarantool CLI version 0.1.0, darwin/amd64. commit: bf83f33\n
     • Tarantool executable found: '/usr/local/bin/tarantool'\n"}
+
+Transition from tarantoolctl to tt
+----------------------------------
+
+System-wide configuration
+~~~~~~~~~~~~~~~~~~~~~~~~~
+``tt`` packages come with a system-wide environment configuration which supports ``tarantoolctl``
+configuration defaults. So, after installation from repository ``tt`` can be used along with
+``tarantoolctl`` for managing applications instances. Here is an example:
+
+.. code-block::
+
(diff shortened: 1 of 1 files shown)


doc/examples.rst
```diff
@@ -387,3 +387,126 @@ Flag without argument:
...
+    modules:
+        directory: modules
+    app:
+        run_dir: ./run/tarantool
+        log_dir: ./log/tarantool
+        log_maxsize: 100
+        log_maxage: 8
+        log_maxbackups: 10
+        restart_on_failure: false
+        data_dir: ./lib/tarantool
```
LeonidVas, line 429: This parameter has been split.

LeonidVas: The `tarantoolctl_layout` parameter is now also added.

doc/examples.rst
```diff
@@ -387,3 +387,126 @@ Flag without argument:
    http://127.0.0.1:1024/tarantool
    {"res":"Tarantool CLI version 0.1.0, darwin/amd64. commit: bf83f33\n
     • Tarantool executable found: '/usr/local/bin/tarantool'\n"}
+
+Transition from tarantoolctl to tt
+----------------------------------
+Suppose we have a set of Tarantool instances managed by ``tarantoolctl`` utility. In order for the tt
```
LeonidVas, line 393: I think we should also describe the case where the user does not have a `.tarantoolctl' file but uses `tarantoolctl' to run tarantool instances on the system.

LeonidVas (changes requested):
Hi! Thank you for the patch.
See several comments bellow.

doc/examples.rst
psergee: Added a section for system-wide config.

doc/examples.rst
psergee: Updated.

LeonidVas (approved):

PR #327, opened 2023-02-02
Ignore SIGURG at Start command

### LINKED ISSUE #325: tt start/stop commands does not always work
Looks like this is some kind of race in start command. It does not start all instances under high load:
```
$ ../../tt/tt start myapp
   • Starting an instance [myapp:s1-master]...
   • Starting an instance [myapp:s1-replica]...
   • Starting an instance [myapp:s2-master]...
   • Starting an instance [myapp:s2-replica]...
   • Starting an instance [myapp:stateboard]...
   • Starting an instance [myapp:router]...
$ ../../tt/tt status myapp
   • myapp:s1-replica: ERROR. The process is dead
   • myapp:s2-master: RUNNING. PID: 48573.
   • myapp:s2-replica: RUNNING. PID: 48581.
   • myapp:stateboard: RUNNING. PID: 48585.
   • myapp:router: RUNNING. PID: 48595.
   • myapp:s1-master: ERROR. The process is dead

$ ../../tt/tt start myapp
   • Starting an instance [myapp:router]...
   • Starting an instance [myapp:s1-master]...
   • Starting an instance [myapp:s1-replica]...
   • Starting an instance [myapp:s2-master]...
   • Starting an instance [myapp:s2-replica]...
   • Starting an instance [myapp:stateboard]...
$ ../../tt/tt status myapp
   • myapp:s1-replica: RUNNING. PID: 48696.
   • myapp:s2-master: RUNNING. PID: 48700.
   • myapp:s2-replica: RUNNING. PID: 48701.
   • myapp:stateboard: ERROR. The process is dead
   • myapp:router: ERROR. The process is dead
   • myapp:s1-master: RUNNING. PID: 48691.
```

`tt stop` does not stop all instances:
```
$ ../../tt/tt stop myapp
   • the process is already dead. Error: "no such process"
   • The Instance myapp:s1-master (PID = 48691) has been terminated.
   • The Instance myapp:s1-replica (PID = 48696) has been terminated.
   • The Instance myapp:s2-master (PID = 48700) has been terminated.
   • The Instance myapp:s2-replica (PID = 48701) has been terminated.
   • the process is already dead. Error: "no such process"
$ ps awux | grep taran
user    48737  0.2  0.1 775644 25524 pts/4    Sl   16:02   0:00 tarantool stateboard.init.lua <running>
```

This issue reproduces on my system after the following commit:
`4b0d12d  tt: fix a race condition between tt start and tt stop`

vr009:
This patch fixes the problem, which was occurring when the watchdog
process received a signal SIGURG from go runtime[1][2] and passed it
to the forked process before the exec call. Fixed by adding a call of
the Ignore function. Receiving this signal is unexpected, cause tt
doesn't work with sockets at all.

Also this patch fixes the signals handling strategy. Now there is only
one handling loop for all signals. Added two fields for controlling
a state of the instance being watched and a mutex for synchronizing
the goroutines changing them.

[1] - https://go.googlesource.com/proposal/+/master/design/24543-non-cooperative-preemption.md
[2] - https://github.com/golang/go/issues/37942

Closes https://github.com/tarantool/tt/issues/325

### DIFF
--- a/cli/running/running.go
+++ b/cli/running/running.go
@@ -6,7 +6,6 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
-	"os/signal"
 	"path"
 	"path/filepath"
 	"strings"
@@ -20,7 +19,6 @@ import (
 	"github.com/tarantool/tt/cli/process_utils"
 	"github.com/tarantool/tt/cli/ttlog"
 	"github.com/tarantool/tt/cli/util"
-	"golang.org/x/net/context"
 	"gopkg.in/yaml.v2"
 )
 
@@ -516,32 +514,21 @@ func FillCtx(cliOpts *config.CliOpts, cmdCtx *cmdcontext.CmdCtx,
 
 // Start an Instance.
 func Start(cmdCtx *cmdcontext.CmdCtx, run *InstanceCtx) error {
-	// Register a context, related with a signal handler that
-	// removes a pid file if interrupt signals were sent before starting
-	// an instance and watchdog signal handlers.
-	sigCtx, stop := signal.NotifyContext(context.Background(),
-		syscall.SIGINT, syscall.SIGTERM)
-	if err := process_utils.CreatePIDFile(run.PIDFile); err != nil {
-		return err
+	logger := createLogger(run)
+	provider := providerImpl{cmdCtx: cmdCtx, instanceCtx: run}
+	preStartAction := func() error {
+		if err := process_utils.CreatePIDFile(run.PIDFile); err != nil {
+			return err
+		}
+		return nil
 	}
+	wd := NewWatchdog(run.Restartable, 5*time.Second, logger, &provider, preStartAction)
 
 	defer func() {
 		cleanup(run)
-		stop()
 	}()
 
-	logger := createLogger(run)
-	provider := providerImpl{cmdCtx: cmdCtx, instanceCtx: run}
-	wd := NewWatchdog(run.Restartable, 5*time.Second, logger, &provider)
-
-	// The Done() state of the context means that the process received
-	// an interrupt signal and there is a need to clean up the resources.
-	select {
-	case <-sigCtx.Done():
-		return nil
-	default:
-		wd.Start()
-	}
+	wd.Start()
 	return nil
 }
 
--- a/cli/running/watchdog.go
+++ b/cli/running/watchdog.go
@@ -35,10 +35,6 @@ type Watchdog struct {
 	// restartTimeout describes the timeout between
 	// restarting the Instance.
 	restartTimeout time.Duration
-	// stopped indicates whether Watchdog was stopped.
-	stopped bool
-	// stopMutex used to avoid a race condition under stopped field.
-	stopMutex sync.Mutex
 	// done channel used to inform the signal handle goroutine
 	// about termination of the Instance.
 	done chan bool
@@ -46,41 +42,63 @@ type Watchdog struct {
 	// and updating may depend on changing external parameters
 	// (such as configuration file).
 	provider Provider
+	// stopMutex used to avoid a race condition under shouldStop field.
+	stopMutex sync.Mutex
+	// shouldStop indicates whether the Watchdog should be stopped.
+	shouldStop bool
+	// preStartAction is a hook that is to be run before the start of a new Instance.
+	preStartAction func() error
 }
 
 // NewWatchdog creates a new instance of Watchdog.
 func NewWatchdog(restartable bool, restartTimeout time.Duration, logger *ttlog.Logger,
-	provider Provider) *Watchdog {
+	provider Provider, preStartAction func() error) *Watchdog {
 	wd := Watchdog{Instance: nil, logger: logger, restartTimeout: restartTimeout,
-		provider: provider}
+		provider: provider, preStartAction: preStartAction}
 
 	wd.done = make(chan bool, 1)
-	wd.stopped = false
 
 	return &wd
 }
 
 // Start starts the Instance and signal handling.
-func (wd *Watchdog) Start() {
+func (wd *Watchdog) Start() error {
+	var err error
+	// Create Instance.
+	if wd.Instance, err = wd.provider.CreateInstance(wd.logger); err != nil {
+		wd.logger.Printf(`Watchdog(ERROR): "%v".`, err)
+		return err
+	}
+	wd.logger = wd.Instance.logger
+	// The signal handling loop must be started before the instance
+	// get started for avoiding a race condition between tt start
+	// and tt stop. This way we avoid a situation when we receive
+	// a signal before starting a handler for it.
+	wd.startSignalHandling()
+
+	if err = wd.preStartAction(); err != nil {
+		wd.logger.Printf(`Pre-start action error: %v`, err)
+		// Finish the signal handling goroutine.
+		wd.done <- true
+		return err
+	}
+
 	// The Instance must be restarted on completion if the "restartable"
 	// parameter is set to "true".
 	for {
 		var err error
-		// Create Instance.
-		if wd.Instance, err = wd.provider.CreateInstance(wd.logger); err != nil {
-			wd.logger.Printf(`Watchdog(ERROR): "%v".`, err)
-			break
-		}
-		wd.logger = wd.Instance.logger
-		// Start the Instance and forwarding signals (except  SIGINT and SIGTERM)
-		wd.startSignalHandling()
+
 		wd.stopMutex.Lock()
-		if !wd.stopped {
-			if err := wd.Instance.Start(); err != nil {
-				wd.logger.Printf(`Watchdog(ERROR): "%v".`, err)
-				wd.stopMutex.Unlock()
-				break
-			}
+		if wd.shouldStop {
+			wd.logger.Printf(`Watchdog(ERROR): terminated before instance start.`)
+			wd.stopMutex.Unlock()
+			return nil
+		}
+		// Start the Instance.
+		if err := wd.Instance.Start(); err != nil {
+			wd.logger.Printf(`Watchdog(ERROR): "%v".`, err)
+			wd.stopMutex.Unlock()
+			break
 		}
 		wd.stopMutex.Unlock()
 
@@ -100,7 +118,7 @@ func (wd *Watchdog) Start() {
 			wd.logger.Println("Watchdog(ERROR): can't check if the instance is restartable.")
 			break
 		}
-		if wd.stopped || !restartable {
+		if wd.shouldStop || !restartable {
 			wd.logger.Println("Watchdog(INFO): the Instance has shutdown.")
 			break
 		}
@@ -112,15 +130,26 @@ func (wd *Watchdog) Start() {
 			wd.logger = logger
 		}
 		time.Sleep(wd.restartTimeout)
+
+		wd.shouldStop = false
+
+		// Recreate Instance.
+		if wd.Instance, err = wd.provider.CreateInstance(wd.logger); err != nil {
+			wd.logger.Printf(`Watchdog(ERROR): "%v".`, err)
+			return err
+		}
+		wd.logger = wd.Instance.logger
+		// Before the restart of an instance start a new signal handling loop.
+		wd.startSignalHandling()
 	}
+	return nil
 }
 
 // startSignalHandling starts signal handling in a separate goroutine.
 func (wd *Watchdog) startSignalHandling() {
 	sigChan := make(chan os.Signal, 1)
-	// Reset unregisters all previous handlers for interrupt signals.
-	signal.Reset(syscall.SIGINT,
-		syscall.SIGTERM, syscall.SIGHUP)
+	// Reset the signal mask before starting of the new loop.
+	signal.Reset()
 	signal.Notify(sigChan)
 
 	// Set barrier to synchronize with the main loop when the Instance stops.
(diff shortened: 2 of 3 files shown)


cli/running/watchdog.go
```diff
@@ -123,6 +116,13 @@ func (wd *Watchdog) startSignalHandling() {
 		syscall.SIGTERM, syscall.SIGHUP)
 	signal.Notify(sigChan)
 
+	// This call is needed to ignore SIGURG signals which are part of
+	// preemptive multitasking implementation in go. See:
+	// https://go.googlesource.com/proposal/+/master/design/24543-non-cooperative-preemption.md.
+	// Also, there is no way to detect if a signal is related to the runtime or not:
+	// https://github.com/golang/go/issues/37942.
+	signal.Ignore(syscall.SIGURG)
```
psergee, line 124: Did you check this does not affect a sig mask of the child Tarantool process?

vr009: The child process is already started by this moment. It will not affect it.

psergee (approved):

cli/running/watchdog.go
psergee: Oh, yes... you've moved start handling again :)

cli/running/watchdog.go
```diff
@@ -73,16 +71,11 @@ func (wd *Watchdog) Start() {
...
-			if err := wd.Instance.Start(); err != nil {
-				wd.logger.Printf(`Watchdog(ERROR): "%v".`, err)
-				wd.stopMutex.Unlock()
-				break
-			}
+		if err := wd.Instance.Start(); err != nil {
+			wd.logger.Printf(`Watchdog(ERROR): "%v".`, err)
+			break
 		}
-		wd.stopMutex.Unlock()
```
psergee, line 85: I think, we have a leak here of the context created in caller of this function. Stop is not called until tt process end. But the context is not used.

cli/running/watchdog.go
```diff
@@ -73,16 +71,11 @@ func (wd *Watchdog) Start() {
...
 		// Start the Instance and forwarding signals (except  SIGINT and SIGTERM)
-		wd.startSignalHandling()
-		wd.stopMutex.Lock()
-		if !wd.stopped {
-			if err := wd.Instance.Start(); err != nil {
-				wd.logger.Printf(`Watchdog(ERROR): "%v".`, err)
-				wd.stopMutex.Unlock()
-				break
-			}
+		if err := wd.Instance.Start(); err != nil {
```
psergee, line 74: If SIGINT is received here (before `Start()`), how will it be handled?

cli/running/watchdog.go
```diff
@@ -73,16 +71,11 @@ func (wd *Watchdog) Start() {
...
+			wd.logger.Printf(`Watchdog(ERROR): "%v".`, err)
+			break
 		}
-		wd.stopMutex.Unlock()
+		wd.startSignalHandling()
```
psergee, line 78: Looks like, we have a gap between Tarantool start and signal handling start. If I'm not mistaken, SIGINT or SIGTERM received here will not be handled correctly and tt process will be terminated. But Tarantool will continue execution. Please, check.

vr009: No, we have registered a handler for those signals before it and have not unregistered. Yes, the handler will not work in this function, but the process will not be terminated. The effect will be the same as by ignoring the signal.

cli/running/watchdog.go
vr009: But we call the stop function in any case, why is it a leak?

cli/running/watchdog.go
vr009: I have replaced using contexts with channels and added an ignoring signals before the start of watchdog, please check it.

cli/running/watchdog.go
vr009: Replaced using contexts with channels.

psergee: > But we call the stop function in any case, why is it a leak?

Looks like, we call stop only after Tarantool instance is stopped, after wd.Start exit. Why do we need the context alive while Tarantool is running?

cli/running/watchdog.go
psergee: Will be SIGTERM & SIGINT silently ignored in this case?

vr009: Yes, until we start a watchdog's handling loop it will be ignored.

psergee: > Yes, until we start a watchdog's handling loop it will be ignored.

So from user's point of view it looks like non-working `tt stop` :) This is not so critical as the start-stop race, but it's still an issue.

cli/running/watchdog.go
```diff
@@ -72,23 +74,32 @@ func (wd *Watchdog) Start() {
 			break
 		}
 		wd.logger = wd.Instance.logger
-		// Start the Instance and forwarding signals (except  SIGINT and SIGTERM)
-		wd.startSignalHandling()
-		wd.stopMutex.Lock()
-		if !wd.stopped {
+
+		wd.watchdogStateMutex.Lock()
+		if !wd.shouldStop {
```
psergee, line 79: Check for `shouldStop` and return if true. Else is not required in this case.
There is no need to have mutex locked during starting an instance. Mutex lock is required only for setting/getting the flag.

cli/running/watchdog.go
```diff
@@ -72,23 +74,32 @@ func (wd *Watchdog) Start() {
...
+			// Start the Instance and forwarding signals (except  SIGINT and SIGTERM)
 			if err := wd.Instance.Start(); err != nil {
 				wd.logger.Printf(`Watchdog(ERROR): "%v".`, err)
-				wd.stopMutex.Unlock()
+				wd.watchdogStateMutex.Unlock()
 				break
 			}
+			wd.started = true
+		} else {
+			wd.logger.Printf(`Watchdog(ERROR): terminated before start.`)
```
psergee, line 88:
```suggestion
			wd.logger.Printf(`Watchdog(ERROR): terminated before instance start.`)
```

cli/running/watchdog.go
vr009: > There is no need to have mutex locked during starting an instance. Mutex lock is required only for setting/getting the flag.

But what if I receive a signal between starting the instance and setting the flag `started` in true? I mean this should happen atomic.

cli/running/watchdog.go
vr009: Fixed,

cli/running/watchdog.go
psergee: Starting application does not have to be atomic. We will set shouldStop to true in this case. And we will check this flag after application is already started.
Ok, I think current variant of locking will not have a big impact on anything. Closing this comment.

cli/running/watchdog.go
```diff
@@ -35,10 +35,12 @@ type Watchdog struct {
 	// restartTimeout describes the timeout between
 	// restarting the Instance.
 	restartTimeout time.Duration
-	// stopped indicates whether Watchdog was stopped.
-	stopped bool
-	// stopMutex used to avoid a race condition under stopped field.
-	stopMutex sync.Mutex
+	// watchdogStateMutex used to avoid a race condition under started and shouldStop fields.
```
LeonidVas, line 38: In fact, mutex and statuses are related to an instance, not to a watchdog. It would be nice to reflect this in the attribute names.

cli/running/watchdog.go
```diff
@@ -35,10 +35,12 @@ type Watchdog struct {
...
+	watchdogStateMutex sync.Mutex
+	// started indicates whether Watchdog has started an Instance.
+	started bool
```
LeonidVas, line 41: This information can be get from the instance object.

LeonidVas (changes requested):
It's the review of the first commit.
* Add `Part of #...` to the commit message.
* Strange that it was not necessary to make changes to the Watchdog tests (Are you sure you don't need them?)
* I don't like the fact that you have to start the signal processing from the outside. Would it make sense to add a callback that would be set externally and called by watchdog after the start of the application or something like that?

cli/running/watchdog.go
```diff
@@ -137,6 +137,13 @@ func (wd *Watchdog) StartSignalHandling() {
 	signal.Reset()
 	signal.Notify(sigChan)
 
+	// This call is needed to ignore SIGURG signals which are part of
```
LeonidVas, line 140: Good catch!!!

cli/running/watchdog.go
vr009: Totally agree, fixed.

cli/running/watchdog.go
vr009: Fixed, thank you.

cli/running/watchdog.go
```diff
@@ -72,23 +74,32 @@ func (wd *Watchdog) Start() {
 			break
```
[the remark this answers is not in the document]
vr009: Fixed.

cli/running/running.go
```diff
@@ -516,32 +514,20 @@ func FillCtx(cliOpts *config.CliOpts, cmdCtx *cmdcontext.CmdCtx,
...
 // Start an Instance.
 func Start(cmdCtx *cmdcontext.CmdCtx, run *InstanceCtx) error {
-	// Register a context, related with a signal handler that
-	// removes a pid file if interrupt signals were sent before starting
-	// an instance and watchdog signal handlers.
-	sigCtx, stop := signal.NotifyContext(context.Background(),
-		syscall.SIGINT, syscall.SIGTERM)
+	logger := createLogger(run)
+	provider := providerImpl{cmdCtx: cmdCtx, instanceCtx: run}
+	wd := NewWatchdog(run.Restartable, 5*time.Second, logger, &provider)
```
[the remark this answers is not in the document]
vr009: Added.

vr009:
> Strange that it was not necessary to make changes to the Watchdog tests (Are you sure you don't need them?).

> I don't like the fact that you have to start the signal processing from the outside. Would it make sense to add a callback that would be set externally and called by watchdog after the start of the application or something like that?

Check the [idea](75ca3d3cfee142c894452fb7d97dba69a92ae85e), please.

cli/running/instance.go
```diff
@@ -50,6 +50,12 @@ type Instance struct {
 	waitMutex sync.Mutex
 	// done represent whether the instance was stopped.
 	done bool
+	// stateMutex used to avoid a race condition under started and shouldStop fields.
+	stateMutex sync.Mutex
+	// started indicates whether the Watchdog has started an Instance.
+	started bool
+	// shouldStop indicates whether the Watchdog should stop the Instance.
```
psergee, line 57: nit: according to this comment, there is no need to call instance.Stop to stop the instance? Is setting this flag really enough to stop the instance?
Instance is a separate struct. Why do we mention watchdog here?

cli/running/instance.go
```diff
@@ -50,6 +50,12 @@ type Instance struct {
 	waitMutex sync.Mutex
 	// done represent whether the instance was stopped.
 	done bool
+	// stateMutex used to avoid a race condition under started and shouldStop fields.
+	stateMutex sync.Mutex
+	// started indicates whether the Watchdog has started an Instance.
+	started bool
```
psergee, line 56: If this is a watchdog flag, it must part of watchdog struct. If not, this state flag must be controlled by instance functions.
Can I use Instance struct without watchdog? If yes, there is no point to mention it in the comment of these flags: Instance is not dependent on watchdog.

cli/running/watchdog.go
psergee: > In fact, mutex and statuses are related to an instance, not to a watchdog. It would be nice to reflect this in the attribute names.

Cannot agree. `shouldStop` flag is watchdog responsibility, not an instance.
Instance is independent entity and shouldStop flag is required by watchdog to make some cleanup, which is not required by Instance.
`started` - it depends...

psergee: I think, mutex should not be a part of Instance struct. Instance does not know anything about signal loop and multithreading. This protection is required only by watchdog, not an instance. So, this is watchdog responsibility to perform safe accesses to an instance.

psergee: And Instance cannot know anything about what fields are accessed simultaneously: states only, or anything else. But this mutex makes unsubstantiated assumptions, which fields will be updated simultaneously somehow, somewhere.

cli/running/watchdog.go
```diff
@@ -100,7 +119,7 @@ func (wd *Watchdog) Start() {
 			wd.logger.Println("Watchdog(ERROR): can't check if the instance is restartable.")
 			break
 		}
-		if wd.stopped || !restartable {
+		if wd.Instance.shouldStop || !restartable {
```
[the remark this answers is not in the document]
vr009: Fixed.

psergee (approved):

cli/running/watchdog.go
vr009: Fixed.

cli/running/watchdog.go
vr009: Discussed by voice - left in the watchdog struct.

cli/running/watchdog.go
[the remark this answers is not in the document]
vr009: Fixed.

cli/running/running.go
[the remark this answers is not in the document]
vr009: Added.

LeonidVas (approved):

PR #344, opened 2023-02-21
instances: add path to 'tt instances' output

better0fdead:
Added path to instanes in 'tt instances' output.

Old output: 

```
f.terekhin@f-terekhin tt221 % ./tt instances
List of enabled applications:
   • 123             
   • anime           
   • app1            
	router
	master
	replica
   • test123        
```

New output:
```
f.terekhin@f-terekhin tt221 % ./tt instances                                     
List of enabled applications:
   • 123 (/Users/f.terekhin/tt221/enables/123)
   • anime (/Users/f.terekhin/tt221/enables/anime)
   • app1 (/Users/f.terekhin/tt221/enables/app1)
	router (router.init.lua)
	master (init.lua)
	replica (init.lua)
   • test123 (/Users/f.terekhin/tt221/enables/test123.lua)
```

### DIFF
--- a/cli/list/list.go
+++ b/cli/list/list.go
@@ -15,13 +15,14 @@ import (
 
 // ListInstances shows enabled applications.
 func ListInstances(cmdCtx *cmdcontext.CmdCtx, cliOpts *config.CliOpts) error {
-	if _, err := os.Stat(cliOpts.App.InstancesEnabled); os.IsNotExist(err) {
+	instanceDir := cliOpts.App.InstancesEnabled
+	if _, err := os.Stat(instanceDir); os.IsNotExist(err) {
 		return fmt.Errorf("instances enabled directory doesn't exist: %s",
-			cliOpts.App.InstancesEnabled)
+			instanceDir)
 	}
 
 	appList, err := util.CollectAppList(cmdCtx.Cli.ConfigDir,
-		cliOpts.App.InstancesEnabled, false)
+		instanceDir, false)
 	if err != nil {
 		return fmt.Errorf("can't collect an application list: %s", err)
 	}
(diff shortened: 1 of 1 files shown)


LeonidVas:
@psergee I think it is better to get a relative path (path inside the env)? It will be shorter and won't give information about the remote host's filesystem when working through the daemon.
What do you think?

psergee:
> @psergee I think it is better to get a relative path (path inside the env)? It will be shorter and won't give information about the remote host's filesystem when working through the daemon. What do you think?

I think relative path is ok for instances. But I'd like to be able to form a whole path without searching for config dir through `tt cfg dump` and appending instance path from `tt instances` to it.
Regarding FS info using the daemon... Is it more dangerous than
```
curl --header "Content-Type: application/json" --request POST --data '{"command_name":"run", "params":["-e", "print(require(\"fio\").open(\"/etc/issue\", {\"O_RDONLY\"}):read())"]}' http://host:1024/tarantool
{"res":"Linux Mint 21.1 Vera \\n \\l\n\n\n"}
```
Shouldn't we restrict/process daemon requests separately, instead of truncating output of the commands for protection? For example, `tt` can check if it is invoked by a daemon, disable verbosity flag to do not print sensitive information, and so on.
If I run the command locally, I'd like to be able to form a full path to the things I'd like to work with. Suppose, I want to make sure that `tt` found my tarantoolctl apps in system directories. I invoke `tt instances` and see `../tarantool/instances.enabled/app1.lua`. From my point of view this is useless a little.

LeonidVas:
After discussion with @psergee, we decided to output the common path to the instance directory and the relative path in that directory on the line with each instance.

better0fdead:
@LeonidVas @psergee 
New output:
```
List of enabled applications:
instances enabled directory: /Users/f.terekhin/tt221/enables
   • 123 (123/)      
   • anime (anime/)  
   • app1 (app1/)    
	router (router.init.lua)
	master (init.lua)
	replica (init.lua)
   • test123 (test123.lua)

```

cli/list/list.go
```diff
@@ -31,15 +32,22 @@ func ListInstances(cmdCtx *cmdcontext.CmdCtx, cliOpts *config.CliOpts) error {
 	}
 
 	fmt.Println("List of enabled applications:")
+	fmt.Printf("instances enabled directory: %s\n", instanceDir)
 
 	for _, app := range appList {
-		log.Infof("%s", color.GreenString(strings.TrimSuffix(app.Name, ".lua")))
-		instances, _ := running.CollectInstances(app.Name, cliOpts.App.InstancesEnabled)
+		appLocation := strings.TrimPrefix(app.Location, instanceDir+"/")
```
psergee, line 38: Please, don't hard-code path separators. Use `os.PathSeparator` or `filepath.Separator`.

better0fdead: Changed

LeonidVas (approved):

PR #348, opened 2023-03-01
connect: allow more characters for URI credential

oleg-jukovec:
Only `:@/` are forbidden to avoid confusion when parsing a URI.

### DIFF
--- a/cli/cmd/connect.go
+++ b/cli/cmd/connect.go
@@ -21,6 +21,7 @@ import (
 const (
 	usernameEnv = "TT_CLI_USERNAME"
 	passwordEnv = "TT_CLI_PASSWORD"
+	userpassRe  = `[^@:/]+:[^@:/]+`
 )
 
 var (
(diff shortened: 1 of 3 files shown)


cli/cmd/connect.go
```diff
@@ -21,6 +21,7 @@ import (
 const (
 	usernameEnv = "TT_CLI_USERNAME"
 	passwordEnv = "TT_CLI_PASSWORD"
+	userpassRe  = `[\w-$!%*#?&]+:[\w-$!%*#?&]+`
```
DifferentialOrange, line 24: Does this charset based on some core tarantool regexps?

DifferentialOrange (approved):

cli/cmd/connect.go
DifferentialOrange: So what was the inspiration for this particular charset?

oleg-jukovec: As an alternative way we could use:

`[^@:/]+:[^@:/]+`

(allow all characters, exclude a special one that configure a path)

DifferentialOrange: As far as I understand, the only restriction in `net.box` is separation symbols, and I don't remember any restrictions in user name and password since they are stored as msgpack strings in tuples. So maybe this approach is the best one.

oleg-jukovec: Yes, it is look like it:
```
> box.schema.user.create('фывфыв1231*??:(?(*?(')
---
...
```

oleg-jukovec: Updated!

DifferentialOrange (approved):

LeonidVas (approved):
Hi! Thank you for the patch.
LGTM.

PR #352, opened 2023-03-03
Implicit platform for docker image

ochaton:
This patch is partial, everywhere we call docker we must pass default platform (linux/amd64). Without this patch tt pack fail to build artefact on M1


psergee (approved):
Thanks for the patch.

LeonidVas (approved):
Hi! Thank you for the patch.
I updated the commit description and take it.

PR #361, opened 2023-03-17
Fix install test hang

psergee:
- Use devnull instead of pipe for stdout.
- Update Ubuntu docker image version to avoid build errors related for no-pic libgomp.a.
- Update cmake and install additional dependencies for building the latest tarantool version.


LeonidVas (approved):
Hi! Thank you for the patch.
LGTM.

Issue #313, opened 2023-01-24
Race condition in tt connect integration tests


psergee:
Connect integration tests workflow:
- start tarantool with app script
- connect to started instance with -l <language> option

There is a race between console socket creation and box.cfg call. The second step may fail with the following error:
```
['/tmp/tmpvxxuh4lf/tt', 'connect', 'test_app', '-f', 'hello.lua', '-l', 'sql']
   ⨯ unable to change a language: unexpected response: ---
- error: Unable to set language to 'sql' with unconfigured box
```
because the socket is already created, but box.cfg from application script is not called yet.

PR #363, opened 2023-03-20
connect: fix change language race in tests

oleg-jukovec:
Connect integration tests with change languages workflow:

1. Start tarantool with app script.
2. Connect to started instance with -l option.

There is a race between console socket creation and box.cfg call. The
second step may fail with the following error:

```
    error: Unable to set language to 'sql' with unconfigured box
```

because the socket is already created, but box.cfg from application
script is not called yet.

The patch creates an additional file `configured` after `box.cfg({})`
to ensure that is is already complited.

Closes https://github.com/tarantool/tt/issues/313

The patchset also fixes non-critical typos.


test/integration/connect/test_connect.py
```diff
@@ -66,7 +66,7 @@ def prepare_test_app_languages(tt_cmd, tmpdir):
     start_app(tt_cmd, tmpdir, "test_app")
 
     # Check for start.
-    file = wait_file(os.path.join(tmpdir, run_path, "test_app"), 'test_app.control', [])
```
psergee, line 69: Can we wait for some sane duration?

psergee: Yes. But I'm worried about very slow Macs. Can we increase the timeout a bit? 10s maybe.

psergee (approved):
lgtm

PR #365, opened 2023-03-20
mage: fix path

better0fdead:
Fixed generateModePath. Previously after filepath.Join() it was cli/codegen/generate_code.go instead of ./cli/codegen/generate_code.go, so build failed in some cases.
previously:
```
go run cli/codegen/generate_code.go
```

now:
```
go run ./cli/codegen/generate_code.go
```

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -17,7 +17,7 @@ import (
 )
 
 const (
-	buildTypeEnv = "TT_CLI_BUILD_SSL"
+	buildTypeEnv  = "TT_CLI_BUILD_SSL"
 	goPackageName = "github.com/tarantool/tt/cli"
 
 	asmflags = "all=-trimpath=${PWD}"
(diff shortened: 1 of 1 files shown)


magefile.go
```diff
@@ -46,7 +46,7 @@ var (
 	pythonExecutableName = "python3"
 	ttExecutableName     = "tt"
 
-	generateModePath = filepath.Join(packagePath, "codegen", "generate_code.go")
+	generateModePath = "./" + filepath.Join(packagePath, "codegen", "generate_code.go")
```
LeonidVas, line 49: Add a description that `filepath.Join()` calls `filepath.Clean()` and removes `./` and the reason why `./cli/some_path` does not equal `cli/some_path`.

LeonidVas (changes requested):
Hi! Thank you for the patch.
`mage:fix path` -> `mage: fix generateModePath`
"some cases" -> describe which cases you mean.

magefile.go
better0fdead: added

PR #375, opened 2023-03-23
pack: fix conflicting dirs for RPM

psergee:
There is an issue with installing packed RPM on CentOS7: conflicting systemd directoriies with systemd RPM. Skip adding these dirs to an RPM package.


LeonidVas (approved):

vr009 (approved):
Thank you for the patch!

PR #379, opened 2023-03-25
Run tests on up-to-date Python dependencies

cclauss:


LeonidVas:
Hi! Thank you for the patch. The current version configuration is fine with us. We don't like to update dependencies just to do an update.

PR #390, opened 2023-03-28
Fix app creation and build issues

psergee:
- Fix typo in `create` help.
- Support building using application name.
- Fix symlink creation error when using custom application destination dir.


LeonidVas (commented):
Hi! Thank you for the patch.
Generally OK.
Please add commit messages to commits:
- create: remove functions from template
- create: fix symlink creation with custom app dir

LeonidVas (approved):

PR #394, opened 2023-03-29
cmd: add config check for almost all commands

### LINKED ISSUE #371: Check all commands for an ability to work without a config
Most of `tt` command shouldn't work without environment configuration. Need to check for config file existence. For example `tt status` does not work without tt.yaml:
```
$ tt  status
   ⨯ tt.yaml not found
```
But `tt install` does:
```
$ tt install tarantool
   • Getting latest tarantool version..
   • Checking existing...     
```

oleg-jukovec:
The check has been added for:

- binaries
- build
- check
- clean
- create
- install
- instances
- logrotate
- pack
- restart
- run
- start
- status
- stop
- uninstall

Closes #371

The patch does not affect:

-  cartridge
-  cat
-  cfg
-  completion
-  connect
-  coredump
-  daemon
-  help
-  init
-  play
-  rocks 
-  search
-  version

I'm not sure about the command list. I suggest to add commands in a review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - tt config is renamed to tt.yaml.
 - Do not use `make` command options for `tarantool` build if `MAKEFLAGS` environment variable
 is set.
+- `binaries`, `build`, `check`, `clean`, `create`, `install`, `instances`, `logrotate`, `pack`,
+  `restart`, `run`, `start`, `status`, `stop`, `uninstall` require environment configuration file.
 
 ### Fixed
 
--- a/cli/cmd/common.go
+++ b/cli/cmd/common.go
@@ -6,6 +6,8 @@ import (
 
 	"github.com/apex/log"
 	"github.com/spf13/cobra"
+	"github.com/tarantool/tt/cli/cmdcontext"
+	"github.com/tarantool/tt/cli/configure"
 	"github.com/tarantool/tt/cli/util"
 )
 
@@ -21,4 +23,14 @@ func handleCmdErr(cmd *cobra.Command, err error) {
 		}
 		log.Fatalf(err.Error())
 	}
+}
+
+// errNoConfig is returned if environment config file tt.yaml not found.
+var errNoConfig = errors.New(configure.ConfigName +
+	" not found, you need to create tt environment config with 'tt init'" +
+	" or provide exact config location with --cfg option")
+
+// isConfigExist returns `true` if environment config file tt.yaml exist.
+func isConfigExist(cmdCtx *cmdcontext.CmdCtx) bool {
+	return cmdCtx.Cli.ConfigPath != ""
 }
--- a/cli/cmd/binaries.go
+++ b/cli/cmd/binaries.go
@@ -24,5 +24,9 @@ func NewBinariesCmd() *cobra.Command {
 
 // internalBinariesModule is a default binaries module.
 func internalBinariesModule(cmdCtx *cmdcontext.CmdCtx, args []string) error {
+	if !isConfigExist(cmdCtx) {
+		return errNoConfig
+	}
+
 	return list.ListBinaries(cmdCtx, cliOpts)
 }
--- a/cli/cmd/build.go
+++ b/cli/cmd/build.go
@@ -32,6 +32,10 @@ func NewBuildCmd() *cobra.Command {
 
 // internalBuildModule is a default build module.
 func internalBuildModule(cmdCtx *cmdcontext.CmdCtx, args []string) error {
+	if !isConfigExist(cmdCtx) {
+		return errNoConfig
+	}
+
 	var buildCtx build.BuildCtx
 	if err := build.FillCtx(&buildCtx, cliOpts, args); err != nil {
 		return err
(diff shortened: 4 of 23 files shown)


cli/cmd/common.go
```diff
@@ -22,3 +23,13 @@ func handleCmdErr(cmd *cobra.Command, err error) {
 		log.Fatalf(err.Error())
 	}
 }
+
+var errNoConfig = errors.New("tt.yaml not found, you need to create tt" +
+	" environment config with tt init")
+
+func checkConfig(cmdCtx *cmdcontext.CmdCtx) error {
```
better0fdead, line 30: i would like to see small description to function

cli/cmd/common.go
```diff
@@ -22,3 +23,13 @@ func handleCmdErr(cmd *cobra.Command, err error) {
 		log.Fatalf(err.Error())
 	}
 }
+
+var errNoConfig = errors.New("tt.yaml not found, you need to create tt" +
+	" environment config with tt init")
```
better0fdead, line 28:
```suggestion
	" environment config with 'tt init'")
```
up to you, i think it looks better to highlight commands in (').

better0fdead:
Thx for patch, just small comments. I think commit title should be:
```cmd: add config check for almost commands```=>```cmd: add config check for almost all commands```

File: cli/cmd/common.go (hunk 1 above)
Comment (oleg-jukovec) [reply cli/cmd/common.go:32]: Added.

File: cli/cmd/common.go (hunk 2 above)
Comment (oleg-jukovec) [reply cli/cmd/common.go]: Thank you, updated.

**Comment by oleg-jukovec** [comment]
> `cmd: add config check for almost commands`=>`cmd: add config check for almost all commands`

Thank you, updated.

**Review by better0fdead — APPROVED** [review]

File: cli/cmd/common.go [hunk 3]
```diff
@@ -22,3 +23,15 @@ func handleCmdErr(cmd *cobra.Command, err error) {
 		log.Fatalf(err.Error())
 	}
 }
+
+// errNoConfig is returned if environment config file tt.yaml not found.
+var errNoConfig = errors.New("tt.yaml not found, you need to create tt" +
```
psergee, line 28: Let's use ConfigName from the `configure` package instead of `tt.yaml`.

cli/cmd/common.go
```diff
@@ -22,3 +23,15 @@ func handleCmdErr(cmd *cobra.Command, err error) {
...
+	" environment config with 'tt init'")
```
psergee, line 29: " or provide exact config location with `--cfg` option" :)

cli/cmd/common.go
```diff
@@ -22,3 +23,15 @@ func handleCmdErr(cmd *cobra.Command, err error) {
...
+
+// checkConfig returns an error if environment config file tt.yaml not found.
+func checkConfig(cmdCtx *cmdcontext.CmdCtx) error {
```
psergee, line 32: It is not clear from the function name what it does. Can we have something like "[is]ConfigFound" or "[is]ConfigFileLoaded" returning bool instead of error?

oleg-jukovec: I've renamed `checkConfig` -> `checkConfigExist`.

It looks easier:

```Go
if err != checkConfigExist(ctx); err != nil {
    return err
}
```

Than:

```Go
if ok != isConfigExist(ctx); !ok {
    return someErr
}
```

for me. But I can redo it if you insist.

cli/cmd/common.go
oleg-jukovec: Thank you, updated.

cli/cmd/common.go
oleg-jukovec: Added.

cli/cmd/common.go
psergee: What do you think about 
```go
if !configExists(ctx) {
    . . . 
}
```
?
Errors are mostly used for exceptional cases. Missing config does not look exceptional for `exists` kind of functions.
But this is for sure exceptional for start command implementation, for example.

psergee (approved):

psergee:
Suppose, we need to check for config in 
connect. to be able to connect using instances names
logrotate. command accepts app name. Where will it search for app log without config?

psergee:
- Update changelog.
- Add a comment with list of the commands which work without a config file.

oleg-jukovec:
> Suppose, we need to check for config in logrotate. command accepts app name. Where will it search for app log without config?

I've added the check for logrotate.

> Update changelog.

Done.

> Add a comment with list of the commands which work without a config file.

I've updated the pull request message. Did you mean to add a comment somewhere else?

CHANGELOG.md
```diff
@@ -40,6 +40,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - tt config is renamed to tt.yaml.
 - Do not use `make` command options for `tarantool` build if `MAKEFLAGS` environment variable
 is set.
+- `binaries`, `build`, `check`, `clean`, `create`, `install`, `instances`, `logrotate`, `pack`,
+  `restart`, `run`, `start`, `stop`, `uninstall` require environment configuration file.
```
psergee, line 44:
```suggestion
  `restart`, `run`, `start`, `status`, `stop`, `uninstall` require environment configuration file.
```

oleg-jukovec: Fixed.

psergee (approved):

PR #422, opened 2023-04-10
Fix tt build failure without system tarantool

psergee:


LeonidVas (approved):
LGTM.

PR #424, opened 2023-04-11
install: make sub-commands from program names

psergee:
tt install uses first cli arguments as a program name to install. This leads to difficulties with cli options separation by program name. Transform program names to sub-commands of tt install to make CLI more convenient.

### DIFF
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -7,32 +7,69 @@ import (
 	"github.com/tarantool/tt/cli/cmdcontext"
 	"github.com/tarantool/tt/cli/install"
 	"github.com/tarantool/tt/cli/modules"
-	"github.com/tarantool/tt/cli/version"
 )
 
-var (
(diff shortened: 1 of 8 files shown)


cli/cmd/install.go
```diff
@@ -7,51 +7,93 @@ import (
...
+	}
+
+	return tntCmd
+}
+
+// newTarantoolCmd creates a command to install tarantool.
+func newTarantoolCmd() *cobra.Command {
+	var tntCmd = &cobra.Command{
+		Use:   "tarantool",
+		Short: "Install tarantool",
```
better0fdead, line 35:
```suggestion
		Short: "Install tarantool community edition",
```

better0fdead (commented):
Looks good to me, just small comment

cli/cmd/install.go
psergee: Added.

PR #427, opened 2023-04-14
install: add directory permissions check before building binaries

### LINKED ISSUE #413: Update logic of tt install with system cfg
Now on macOS, the `tt install tarantool` command with the system config, done without `sudo`, builds tarantool and failed when copying the binaries into `/opt/tarantool/bin` because don't have an access rights.
One or both must be done:
- Check for access to install the binary before building  the tarantool
- offer to enter the password and with updated access rights copy the binaries after a successful build

vr009:
Closes #413

### DIFF
--- a/cli/install/install.go
+++ b/cli/install/install.go
@@ -24,6 +24,7 @@ import (
 	"github.com/tarantool/tt/cli/templates"
 	"github.com/tarantool/tt/cli/util"
 	"github.com/tarantool/tt/cli/version"
+	"golang.org/x/sys/unix"
 )
 
 // Backported cmake rules for static build.
@@ -899,6 +900,10 @@ func installTarantool(binDir string, incDir string, installCtx InstallCtx,
 	return nil
 }
 
+func dirsIsWriteable(dir string) bool {
+	return unix.Access(dir, unix.W_OK) == nil
+}
+
 // installTarantoolEE installs selected version of tarantool-ee.
 func installTarantoolEE(binDir string, includeDir string, installCtx InstallCtx,
 	distfiles string, cliOpts *config.CliOpts) error {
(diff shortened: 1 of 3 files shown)


cli/install/install.go
```diff
@@ -899,6 +899,19 @@ func installTarantool(binDir string, incDir string, installCtx InstallCtx,
 	return nil
 }
 
+func dirsAreWriteable(dirs []string) bool {
+	for _, dir := range dirs {
+		tempFileName := filepath.Join(dir, "tempfile")
+		err := os.WriteFile(tempFileName, []byte{}, 0)
```
LeonidVas, line 905: Why don't you use `os.Stat()`?

cli/install/install.go
```diff
@@ -1061,6 +1074,13 @@ func Install(binDir string, includeDir string, installCtx InstallCtx,
 	local string, cliOpts *config.CliOpts) error {
 	var err error
 
+	// This check is needed for knowing that we will be able to copy
+	// recently built binaries to the corresponding bin and include directories.
+	if !dirsAreWriteable([]string{binDir, includeDir}) {
+		return fmt.Errorf("some of directories are not writeable for the current user. " +
+			"Please, use sudo for the successful install")
```
LeonidVas, line 1081: You don't have to use `sudo`, you can just update rights (the user have a choice)?

cli/install/install.go
vr009: I use the unix syscall `access` for finding out if the current user has write access.

cli/install/install.go
```diff
@@ -1061,6 +1074,13 @@ func Install(binDir string, includeDir string, installCtx InstallCtx,
 	local string, cliOpts *config.CliOpts) error {
 	var err error
 
+	// This check is needed for knowing that we will be able to copy
+	// recently built binaries to the corresponding bin and include directories.
+	if !dirsAreWriteable([]string{binDir, includeDir}) {
+		return fmt.Errorf("some of directories are not writeable for the current user. " +
```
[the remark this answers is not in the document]
vr009: Fixed.

cli/install/install.go
vr009: Fixed the output, thank you.

LeonidVas (approved):

PR #431, opened 2023-04-18
install: check include directory permission before it's mutation

vr009:
This patch fixes the behavior by permissions checking, now we ignore the directory check if it doesn't exist.

Follow up #413

### DIFF
--- a/cli/install/install.go
+++ b/cli/install/install.go
@@ -900,10 +900,6 @@ func installTarantool(binDir string, incDir string, installCtx InstallCtx,
 	return nil
 }
 
-func dirsIsWriteable(dir string) bool {
-	return unix.Access(dir, unix.W_OK) == nil
-}
-
 // installTarantoolEE installs selected version of tarantool-ee.
 func installTarantoolEE(binDir string, includeDir string, installCtx InstallCtx,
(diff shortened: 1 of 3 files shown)


LeonidVas (changes requested):
Hi! Thank you for the patch!
We discussed in a voice that we should also check if we can create a directory (e.g. bin) if it is absent.

cli/install/install.go
```diff
@@ -1061,6 +1057,24 @@ func installTarantoolEE(binDir string, includeDir string, installCtx InstallCtx,
...
+// subDirIsWritable checks if the passed dir doesn't exist but can be created.
+func subDirIsWritable(dir string) bool {
+	var err error
+	for {
+		_, err = os.Stat(dir)
+		if os.IsNotExist(err) {
+			dir = filepath.Dir(dir)
+			continue
+		}
+		return dirsIsWriteable(dir)
```
0x501D, line 1074: Why not using `filepath.Walk` ?

vr009: Cause we need to walk up to the root, AFAIU walkdir works in another direction.

0x501D: accepted

LeonidVas (approved):

PR #454, opened 2023-05-12
status: make the output of the command as a table

### LINKED ISSUE #197: Make the output of the `status` command as a table
Make the output of the `status` command in the form of a table for a more comfortable perception.

askalt:
Made the output of the status command as a table.
Have done it using tabwriter from the standard library. Since tabwriter does not support ANSI escape codes, I need to remove extra spaces from the table header to use colored text in the second column of the table.

Closes #197

### DIFF
--- a/cli/status/status.go
+++ b/cli/status/status.go
@@ -1,0 +1,65 @@
+package status
+
+import (
+	"fmt"
+	"strings"
+	"text/tabwriter"
+
+	"github.com/fatih/color"
+	"github.com/tarantool/tt/cli/process_utils"
+	"github.com/tarantool/tt/cli/running"
+)
+
+const (
+	// colorBytesNumber is the number of bytes added to the colorized string.
+	colorBytesNumber = 9
+
+	// padding is the padding between two columns of the table.
+	padding = 5
+)
+
+var header = []string{"INSTANCE", "STATUS", "PID"}
+
+// Status writes the status as a table.
+func Status(runningCtx running.RunningCtx) error {
+	instColWidth := 0
+	sb := strings.Builder{}
+	tw := tabwriter.NewWriter(&sb, 0, 1, padding, ' ', 0)
+
+	fmt.Fprintln(tw, strings.Join(header, "\t"))
+	for _, run := range runningCtx.Instances {
+		fullInstanceName := running.GetAppInstanceName(run)
+		procStatus := running.Status(&run)
+		if len(fullInstanceName) > instColWidth {
+			instColWidth = len(fullInstanceName)
+		}
+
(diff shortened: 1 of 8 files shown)


cli/status/status.go
```diff
@@ -0,0 +1,62 @@
...
+			fmt.Fprintf(tw, "%d", procStatus.PID)
+		}
+		fmt.Fprintf(tw, "\n")
+	}
+
+	if err := tw.Flush(); err != nil {
+		return err
+	}
+
+	instColWidth += RowIndent
```
psergee, line 48: Please, add a comment about header manipulations below.

cli/status/status.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/fatih/color"
+	"github.com/tarantool/tt/cli/process_utils"
+	"github.com/tarantool/tt/cli/running"
+)
+
+const (
+	ColorSeqLen = 9
+	RowIndent   = 5
+	kInstance   = "INSTANCE"
```
psergee, line 16: No need "k" prefix for consts.
I'd like to see from these three var names that they are related to the table header.
Alternatively, we can create 'columns' or 'header' array. So the purpose of these names and supposed columns order will be more clear. And use `Join` with tabs for this array instead of printf with constant names.

psergee (commented):
Please attach two screen shots of terminal output from `tt status` with and without colors.

cli/status/status.go
```diff
@@ -0,0 +1,62 @@
...
+	"strings"
+	"text/tabwriter"
+
+	"github.com/fatih/color"
+	"github.com/tarantool/tt/cli/process_utils"
+	"github.com/tarantool/tt/cli/running"
+)
+
+const (
+	ColorSeqLen = 9
```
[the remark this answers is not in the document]
askalt: I added comments about these consts and, in addition, chose much more understandable names.

cli/status/status.go
askalt: Added.

cli/status/status.go
askalt: Replaced these consts with a 'header' array and used `Join` to format.

cli/status/status.go
```diff
@@ -0,0 +1,65 @@
...
+
+const (
+	// colorBytesNumber is the number of bytes added to the colorized string.
+	colorBytesNumber = 9
+
+	// padding is the padding between two columns of the table.
+	padding = 5
+)
+
+var header = []string{"INSTANCE", "STATUS", "PID"}
```
psergee, line 21:
```suggestion
var header = [...]string{"INSTANCE", "STATUS", "PID"}
```

psergee (approved):

PR #460, opened 2023-05-18
build: hide building output if there are no errors.

### LINKED ISSUE #450: tt build app must hide building output if there is no errors
`tt install <app>` does not show compilation output if there is no error. 
`tt build app` must behave the same way.

askalt:
Hide building output if there are no errors.

tt build re-worked to redirect stdout to /dev/null in case of no -V.

Closes #450.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+        
+### Changed
+- `tt build` now hides building output if `-V` is not provided.
 
 ## [1.1.0] - 2023-05-02
 
--- a/cli/build/local.go
+++ b/cli/build/local.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"syscall"
 
 	"github.com/apex/log"
 	"github.com/tarantool/tt/cli/cmdcontext"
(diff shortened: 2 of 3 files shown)


cli/build/local.go
```diff
@@ -68,11 +78,32 @@ func buildLocal(cmdCtx *cmdcontext.CmdCtx, cliOpts *config.CliOpts, buildCtx *Bu
...
 
 	// Run rocks make.
-	rocksMakeCmd := []string{"make"}
+	log.Infof("Running rocks make")
+	logOutput := os.Stdout
+	if !cmdCtx.Cli.Verbose {
+		if logOutput, err = os.CreateTemp("", "app_build"); err != nil {
+			return err
+		}
+		defer logOutput.Close()
```
psergee, line 87: The file is not removed, but it should be.

cli/build/local.go
```diff
@@ -68,11 +78,32 @@ func buildLocal(cmdCtx *cmdcontext.CmdCtx, cliOpts *config.CliOpts, buildCtx *Bu
...
 	}
-	if err := rocks.Exec(cmdCtx, cliOpts, rocksMakeCmd); err != nil {
+	selfPath, err := os.Executable()
+	if err != nil {
+		return err
+	}
+
+	rocksMakeCmd := exec.Command(selfPath, rocksMakeArgs...)
+	rocksMakeCmd.Stdout = logOutput
+	rocksMakeCmd.Stderr = logOutput
```
psergee, line 101: Looks like, there is an issues with current "hide build logs" `tt install` implementation and therefore `tt build`: it prints build logs to stdout, and will hide all errors from me, if I redirect stdout to dev null. But I must be able to see the error messages from stderr. And must be able to skip all non-error messages. Also, I think we must print only stderr output in case of error if `-V` is not set. If `-V` option is set - print everything even if the error is not encountered. So, let's avoid stderr redirection and do not print something from stdout if `-V` is not set (ioutil.Discard instead of a log file) . Please check that rocks make error are printed to stderr.
I'd like to do the same (not under this task) for `tt install`. @LeonidVas what do you think?

psergee (commented):
This is user visible change, update changelog.

cli/build/local.go
psergee: Based on this, we can try to redirect or own stdout to null and do not run child tt process. So, luarocks make errors will be printed to stderr, stdout will be discarded. And if someone wants to see the full output, the `--verbose` option is available.

LeonidVas: > Looks like, there is an issues with current "hide build logs" `tt install` implementation and therefore `tt build`: it prints build logs to stdout, and will hide all errors from me, if I redirect stdout to dev null. But I must be able to see the error messages from stderr. And must be able to skip all non-error messages. Also, I think we must print only stderr output in case of error if `-V` is not set. If `-V` option is set - print everything even if the error is not encountered. So, let's avoid stderr redirection and do not print something from stdout if `-V` is not set (ioutil.Discard instead of a log file) . Please check that rocks make error are printed to stderr. I'd like to do the same (not under this task) for `tt install`. @LeonidVas what do you think?

This sounds good to me.

cli/build/local.go
```diff
@@ -47,6 +47,16 @@ func runBuildHook(buildCtx *BuildCtx, hookNames []string) error {
 	return nil
 }
 
+// printLog prints logfile to stdout.
+func printLog(logName string) error {
```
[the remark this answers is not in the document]
askalt: This function was removed as unnecessary.

cli/build/local.go
askalt: Implemented in this way.

psergee:
Please add notes to changelog.

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+        
+### Changed
+- `tt build` now hides building output if `-V` not provided.
```
psergee, line 11:
```suggestion
- `tt build` now hides building output if `-V` is not provided.
```

psergee (approved):

PR #461, opened 2023-05-18
doc: convert README to md

LeonidVas:


psergee (approved):

doc/examples.md
```diff
@@ -0,0 +1,635 @@
...
+tt:
+    app:
+```
+
+For packing it into tarball, call:
+
+``` bash
+$ tt pack tgz
+   • Apps to pack: single_environment
+   • Generating new tt.yaml for the new package.
```
psergee, line 225: Is not it better to use ``` without `bash` for commands and output? Because `bash` highlights keywords such as "for" in command's output.

LeonidVas: Replaced with `console`.

PR #481, opened 2023-06-01
rocks: consider environment for some paths

### LINKED ISSUE #471: Support overriding of the tt environment configuration with environment variables.
tt should consider environment, when working with some directories. Environment variables to support:
- `TT_CLI_REPO_ROCKS_PATH`: check directory `tt.app.repo.rocks` for rocks repo manifest (`manifest` file). If it exists, use this dir as rock repo path. If not, check for the env var. If it is set, use it as a rocks repo path.

- `TT_CLI_TARANTOOL_INCLUDE_PATH`: If tarantool executable is system, and this variable is set, use it instead of `tt.app.inc_dir` for the following commands: build, rocks, pack, cfg. `install` and `uninstall` commands should ignore this variable and install/install tarantool headers to/from `tt.app.inc_dir`. If tarantool is installed by `tt`, ignore this var and use `tt.app.inc_dir`. If the variable is not set, also use `tt.app.inc_dir`.

psergee:
Rocks local repository path and tarantool installation prefix path can be specified using environment variables. TT_CLI_REPO_ROCKS variable is used if it is set and there is no repo.rocks in tt config or repo.rocks directory does not include repository manifest.
TT_CLI_TARANTOOL_PREFIX variable is used for build, rocks commands if it is set and tarantool executable is found in PATH.

Closes #471

### DIFF
--- a/cli/rocks/rocks.go
+++ b/cli/rocks/rocks.go
@@ -4,9 +4,12 @@ import (
 	"embed"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strings"
 
+	"github.com/apex/log"
 	"github.com/tarantool/tt/cli/cmdcontext"
 	"github.com/tarantool/tt/cli/config"
 	"github.com/tarantool/tt/cli/util"
@@ -16,6 +19,12 @@ import (
 //go:embed extra/*
 //go:embed third_party/luarocks/src/*
 var luarocks embed.FS
+
+const (
+	rocksRepoManifestName     = "manifest"
+	repoRocksPathEnvVarName   = "TT_CLI_REPO_ROCKS"
+	tarantoolPrefixEnvVarName = "TT_CLI_TARANTOOL_PREFIX"
+)
 
 // addLuarocksRepoOpts adds --server option to luarocks command line if rocks repository
 // info is specified in tt config. Return updated args slice.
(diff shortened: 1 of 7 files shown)


0x501D (approved):

opomuc:
Technically, this won't close #471, because the ticket has 2 points: rocks repo and include dir. This PR closes only part for rocks repo.

opomuc:
#471 has a point abour include dir. Does this PR closes it as well? `TT_CLI_TARANTOOL_PREFIX` is supposed to be in place of `TT_CLI_TARANTOOL_INCLUDE_PATH`?

cli/rocks/rocks.go
```diff
@@ -44,6 +104,8 @@ func Exec(cmdCtx *cmdcontext.CmdCtx, cliOpts *config.CliOpts, args []string) err
 		cmd = "help"
 	}
 
+	cliOpts.Repo.Rocks = getRocksRepoPath(cliOpts.Repo.Rocks)
```
opomuc, line 107: `tt.yaml.default` example in repository does not have `tt.repo.rocks` configuration

psergee: This is not an example config. This config is used for tt packages creation. It does not include rocks repository.

psergee:
> #471 has a point about include dir. Does this PR closes it as well? `TT_CLI_TARANTOOL_PREFIX` is supposed to be in place of `TT_CLI_TARANTOOL_INCLUDE_PATH`?
Does this PR closes it as well?. Yes.
Is prefix var supposed to be used instead of INCLUDE? Not exactly.
Include path can be obtained from prefix dir: include path = <prefix_dir>/include/tarantool. So, I use prefix dir to do not specify both. 2 env vars is better than 3. In `env.sh` in SDK I'm going to set prefix to the directory where SDK was extracted to.

opomuc:
> TT_CLI_TARANTOOL_PREFIX variable is used for build, rocks commands
if it is set and tarantool executable is found in PATH.

`tt build` should work with SDK as well. Is that correct? will it still try to build from source, even if we have rocks locally?

Issue #426, opened 2023-04-13
tt connect adds new item with command to the latest item in a history


ligurio:
### Steps to reproduce:

Terminal 1:
```
$ tarantool
tarantool> box.cfg{listen=3301}
tarantool> box.schema.user.create("user", { password = "password", if_not_exists = true })
tarantool> box.schema.user.grant("user", "super", nil, nil, { if_not_exists = true })
```

Terminal 2:

```
[0] ~$ tt connect 127.0.0.1:3301 -u user -p password
   • Connecting to the instance...
   • Connected to 127.0.0.1:3301

127.0.0.1:3301> os.exit()
```

Terminal 1:

tarantool finished interactive session:

```
2023-04-13 20:47:25.636 [174909] main/117/iproto.shutdown I> tx_binary: stopped
2023-04-13 20:47:28.637 [174909] main/102/on_shutdown fiber.c:659 E> TimedOut: timed out
2023-04-13 20:47:28.637 [174909] main/102/on_shutdown main.cc:167 E> on_shutdown triggers failed
2023-04-13 20:47:28.637 [174909] main/102/on_shutdown trigger.cc:251 E> TimedOut: timed out
[0] ~/sources/MRG/tarantool$ 
```

Run it again and press Up button for the latest command in a history.

```
[0] ~/sources/MRG/tarantool$ tarantool
Tarantool 2.11.0-rc2-0-g2ae0c94a2
type 'help' for interactive help
tarantool> box.schema.user.grant("user", "super", nil, nil, { if_not_exists = true })
os.exit()
```

### Actual results

latest command entered in `tt connect` session was added to the item with last command in a history entered in tarantool session:

```
[0] ~/sources/MRG/tarantool$ tarantool
Tarantool 2.11.0-rc2-0-g2ae0c94a2
type 'help' for interactive help
tarantool> box.schema.user.grant("user", "super", nil, nil, { if_not_exists = true })
os.exit()
```

Content of `~/.tarantool_history`:
```
<trimmed>
#1681407773
box.schema.user.grant("user", "super", nil, nil, { if_not_exists = true })
#1681407780
box.cfg{listen=3301}
#1681407787

#1681407791
box.schema.user.grant("user", "super", nil, nil, { if_not_exists = true })
os.exit()
```

### Expected results

One command per item in a history.

### Environment

Tarantool 2.11.0-rc2-0-g2ae0c94a2
[...]

PR #482, opened 2023-06-05
connect: make console history consistent

### LINKED ISSUE #478: repl: doesn't support ~/.tarantool_history with timestamps
Tarantool writes `~/.tarantool_history` in the readlines's format with timestamps since 2.10.2.

It looks so:

```
#1677097730
a
#1677097730
b
#1677097730
c
```

See https://github.com/tarantool/tarantool/pull/7525 for the details.

It is important to note that if readline detects the old one-line-per-command format, it automatically converts it to the new one. Seamless upgrading.

What is expected from the `tt` side:

1. `tt` should parse the new format correctly.
2. `tt` should write the history in the new format, when the file in the new format.

The latter point is important, because otherwise after some commands from `tt`'s REPL, tarantool's REPL interprets them as one multiline command.

askalt:
Reworked operations with console history in `tt connect`
to be consistently with tarantool console history.
i.e.:
1) When a new command appears, it completely
rewrites the history file.
2) When writes history, it uses format with
timestamps.
3) Added parsing of timestamps history format.
Parsing mode (with timestamps or not) switches
in dependency of the history file content.

Closes https://github.com/tarantool/tt/issues/478, https://github.com/tarantool/tt/issues/426

### DIFF
--- a/cli/connect/history.go
+++ b/cli/connect/history.go
@@ -1,0 +1,119 @@
+package connect
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/tarantool/tt/cli/util"
+)
+
+// commandHistory stores console command history.
+type commandHistory struct {
+	filepath    string
+	commands    []string
+	timestamps  []int64
+	maxCommands int
+}
+
+// parseHistoryCells extracts timestamps and commands from history lines.
+// If format doesn't contain timestamps, it sets the current timestamp
+// for each command.
+func parseHistoryCells(
+	lines []string) (commands []string, timestamps []int64) {
+	timestampRegex := regexp.MustCompile(`^#\d+$`)
+
+	// startPos is the first position of a timestamp.
+	startPos := -1
+	for i, line := range lines {
+		if timestampRegex.MatchString(line) {
+			startPos = i
+			break
+		}
+	}
+	timestamps = make([]int64, 0)
+	if startPos == -1 {
+		// Read one line per command.
+		// Set the current timestamp for each command.
+		commands = lines
+		timestamp := time.Now().Unix()
+		for range lines {
+			timestamps = append(timestamps, timestamp)
+		}
+		return
+	}
+
+	commands = make([]string, 0)
+	for startPos < len(lines) {
+		j := startPos + 1
+
+		// Move pointer to the next timestamp.
+		for j < len(lines) && !timestampRegex.MatchString(lines[j]) {
+			j++
+		}
+
+		// Extract the current timestamp.
+		timestamp, err := strconv.ParseInt(lines[startPos][1:], 10, 0)
+
+		if j != startPos+1 && err == nil {
+			timestamps = append(timestamps, timestamp)
+			commands = append(commands, strings.Join(lines[startPos+1:j], "\n"))
+		}
+		startPos = j
+	}
+
+	return
+}
+
+// load loads console history from the history file.
+func (history *commandHistory) load() error {
+	rawLines, err := util.GetLastNLines(history.filepath, history.maxCommands)
+	if err != nil {
+		return err
+	}
+
+	history.commands, history.timestamps = parseHistoryCells(rawLines)
+	return nil
+}
+
+// appendCommand appends new command to the history.
+func (history *commandHistory) appendCommand(command string) {
+	history.commands = append(history.commands, command)
+	history.timestamps = append(history.timestamps, time.Now().Unix())
+	if len(history.commands) > history.maxCommands {
+		history.commands = history.commands[1:]
+		history.timestamps = history.timestamps[1:]
+	}
+}
+
+// writeToFile writes console history to the file.
+func (history *commandHistory) writeToFile() error {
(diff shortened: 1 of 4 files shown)


cli/connect/history.go
```diff
@@ -0,0 +1,133 @@
...
+	historyContent := bytes.Buffer{}
+	for i, command := range history.commands {
+		historyContent.WriteString(fmt.Sprintf("#%d\n%s\n", history.timestamps[i], command))
+	}
+	if err := os.WriteFile(history.filepath, historyContent.Bytes(), 0644); err != nil {
+		return fmt.Errorf("failed to write to history file: %s", err)
+	}
+
+	return nil
+}
```
Totktonada, line 119: Let's consider the following scenario.

Console 1:

```sh
$ tt connect uri1
uri1> bear
uri1> wolf
uri1> hippo
uri1>
```

Console 2:

```sh
$ tt connect uri2
uri2> banana
uri2> orange
uri2> lemon
uri2>
```

Swich to console 1:

```sh
uri1> Ctrl+D
```

Switch to console 2:

```sh
uri2> Ctrl+D
```

Will animal and fruit commands saved both?

askalt: No, the history of only one connection will be saved. Tarantool acts the same way.

Totktonada: Tarantool drops the history to the file after each command. Nope, everything is OK there with several simultaneously opened consoles.

Totktonada: A little confusion occured here.

First, the scenario works differently if we open two tarantool/tt sessions first and if we type the commands in the first session and then open the second one. The reason is that the history is loaded at the process start.

From Albert's words -- the file is written per command, just like in tarantool. So the behavior should be the same.

cli/connect/history.go
```diff
@@ -0,0 +1,133 @@
...
+	return
+}
+
+// load loads console history from the history file.
+func (history *commandHistory) load() error {
+	if history == nil {
+		return nil
+	}
+
+	rawLines, err := util.GetLastNLines(history.filepath, history.maxCommands)
```
psergee, line 81: What if we have multiline commands in the history file? Latest 100 lines may be only 10 commands.

askalt: Yes, that's true. But here i used `maxCommands` as the upper bound. Probably user don't need 10000 commands exactly.

cli/connect/history.go
```diff
@@ -0,0 +1,133 @@
...
+	history.timestamps = append(history.timestamps, time.Now().Unix())
+	if len(history.commands) > history.maxCommands {
+		history.commands = history.commands[1:]
+		history.timestamps = history.timestamps[1:]
+	}
+}
+
+// writeToFile writes console history to the file.
+func (history *commandHistory) writeToFile() error {
+	if history == nil {
```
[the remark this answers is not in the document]
askalt: Here i did it for the next reason: if `newCommandHistory` returns an error, application should work correctly even without recording the console history. Because of this, ignore nil here.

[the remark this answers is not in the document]
psergee: I understand it. I consider this functionality (history saving) as a potentially separate API, which does not know who calls it. And this is our console responsibility to decide if history saving is necessary. Yes, we can leave this check in our internal history saving implementation for now.

cli/connect/history.go
psergee: > Probably user don't need 10000 commands exactly.
Probably yes.

cli/connect/history.go
```diff
@@ -0,0 +1,133 @@
...
+	timestamps  []int64
+	maxCommands int
+}
+
+// parseHistoryCells extracts timestamps and commands from history lines.
+// If format doesn't contain timestamps, it sets the current timestamp
+// for each command.
+func parseHistoryCells(
+	lines []string) (commands []string, timestamps []int64) {
+	regex := regexp.MustCompile(`^#\d+$`)
```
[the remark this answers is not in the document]
askalt: Fixed.

cli/connect/history.go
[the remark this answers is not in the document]
askalt: Nice point, logic is separated now.

cli/connect/history.go
```diff
@@ -0,0 +1,133 @@
...
+func (history *commandHistory) writeToFile() error {
+	if history == nil {
+		return nil
+	}
+
+	historyContent := bytes.Buffer{}
+	for i, command := range history.commands {
+		historyContent.WriteString(fmt.Sprintf("#%d\n%s\n", history.timestamps[i], command))
+	}
+	if err := os.WriteFile(history.filepath, historyContent.Bytes(), 0644); err != nil {
```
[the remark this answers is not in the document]
askalt: Fixed.

cli/connect/history.go
```diff
@@ -0,0 +1,133 @@
...
+	for startPos < len(lines) {
+		j := startPos + 1
+
+		// Move pointer to the next timestamp.
+		for j < len(lines) && !regex.MatchString(lines[j]) {
+			j++
+		}
+
+		// Extract the current timestamp.
+		timestamp, err := strconv.Atoi(lines[startPos][1:])
```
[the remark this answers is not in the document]
askalt: Fixed.

PR #490, opened 2023-06-12
connect: support for more flexible URIs

### LINKED ISSUE #462: connect: synchronize uri with Tarantool implementation
Users complain that the `connect` does not behave as they expect:
```
$ tt connect client:secret@~/p/t/tarantool/instance_001.iproto 
   ⨯ tt.yaml not found
$ tt connect client:secret@.//p/t/tarantool/instance_001.iproto 
   ⨯ tt.yaml not found
$ tt connect client:secret@./p/t/tarantool/instance_001.iproto 
<..success..>
```

We need to synchronize uri regular expressions with Tarantool(console.connect):

https://github.com/tarantool/tt/blob/f0d3d04c446f777d268169cbaba46b46ae7abd74/cli/cmd/connect.go#L108-L117

https://github.com/tarantool/tt/blob/f0d3d04c446f777d268169cbaba46b46ae7abd74/cli/cmd/connect.go#L90-L99

askalt:
Reworked regexps for URI in the case of using an unix socket.
Now it supports `../path` and `~/path` patterns.

Besides, this patch moves all URI parsing logic to
`cli/cmd/connect`.

Closes https://github.com/tarantool/tt/issues/462


psergee:
`tt connect` `unix:./socket.control` URIs must be prioritized over `app:instance` format. Currently `tt` treats all "one:two" URIs like app + instance, which is not the case for most of `unix:./path` invocations. So the result of unix socket URI passing is:
```
tt connect unix:./data/router_1.control
   ⨯ tt.yaml not found
```
My proposal is if `tt` detects correct URI format, it does not try to handle it as `app:instance` case and does not require `tt.yaml`. This can either be done under this PR or create a new one (issue + PR).

cli/connector/opts.go
```diff
@@ -66,7 +68,13 @@ func MakeConnectOpts(connString, username, password string,
 
 	addrLen := len(address)
 	switch {
-	case addrLen > 0 && (address[0] == '.' || address[0] == '/'):
+	case addrLen > 0 && (address[0] == '.' || address[0] == '/' || address[0] == '~'):
```
psergee, line 71: Extend unit tests for MakeConnectOpts to check these changes.

cli/cmd/connect.go
```diff
@@ -104,12 +104,15 @@ func isBaseURI(str string) bool {
 	// host:port
 	tcpReStr := `(tcp://)?([\w\\.-]+:\d+)`
 	// unix://../path
+	// unix://~/path
 	// unix:///path
 	// unix://path
-	unixReStr := `unix://[./]*[^\./@]+[^@]*`
-	// ./path
+	unixReStr := `unix://[~./]*[^\./@]+[^@]*`
```
psergee, line 110: Looks like this regexp does not support `unix/:` format. But according to unit tests and [cli/connector/opts.go:86](https://github.com/tarantool/tt/pull/490/files#diff-7b842aea35da08fdf0c8cf553dd425f22826b430dce39fe3839974443085c2f4R86)
it should.
```
$ tarantoolctl connect unix/:./data/router_1.control 
connected to unix/:./data/router_1.control
unix/:./data/router_1.control> 

$ tt connect unix/:./data/router_1.control 
   ⨯ tt.yaml not found
```

psergee: > This is expected. Expanded `URI` formats are more strict: `[sheme://][user:pass@]URI_PATH` .

It's ok. But the code and tests for `unix/:` must be removed.

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- `tt connect`: support for the `../` and `~/` at the beginning of the uri, when using unix sockets.
```
oleg-jukovec, line 12:
```suggestion
- `tt connect`: support for the `../` and `~/` at the beginning of the URI, when using unix sockets.
```

cli/connector/opts.go
askalt: Unit tests extended.

cli/cmd/connect.go
```diff
@@ -104,12 +104,15 @@ func isBaseURI(str string) bool {
...
+	// ../path
+	// ~/path
 	// /path
-	pathReStr := `\.?/[^\./].*`
+	// ./path
+	pathReStr := `[\.~]*/[\./]*[^\./].*`
```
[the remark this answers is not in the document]
askalt: Added more strictly conditions for paths, tests extended.

cli/cmd/connect.go
askalt: Unused parse logic  and tests were removed.

cli/connector/opts.go
```diff
@@ -66,28 +68,27 @@ func MakeConnectOpts(connString, username, password string,
...
```
oleg-jukovec, line 71: In general it looks like violating the Single Responsibility Principle: we have 2 places where we process URI.

I suggest to do it at `cli/cmd/connect.go` only and don't try to parse URI at `MakeConnectOpts`.

1. We need to move the parse logic into `cli/cmd/connect.go`.
2. We could remove `MakeConnectOpts` and initialize the struct object like this:
```Go
ConnectOpts{
    Network: UnixNetwork,
    Address: address,
    User: user,
    Password: pass,
    SslOpts: sslOpts,
}
``` 
It is just a suggestion. Up to you.

cli/cmd/connect.go
psergee: > This is expected. Expanded `URI` formats are more strict: `[sheme://][user:pass@]URI_PATH` .

As I can see, this format is supported by tarantool `console`:
`console.listen('unix/:/tmp/X.sock')`
Taking this into account, should we prohibit the user from using this format in `tt`? @LeonidVas what do you think?

cli/connector/opts_test.go
askalt, line 1: Tests, which check parsing logic moved to `cli/cmd/connect_test.go`.

cli/connector/opts.go
askalt: It's nice point, parsing logic and tests were moved to `cli/cmd/connect.go`. 
Also, `resolveConnectOpts` now returns `ConnectOpts` and the remaining args (instead of replacing first argument to the URI).

cli/cmd/connect.go
```diff
@@ -122,18 +137,53 @@ func isCredentialsURI(str string) bool {
...
+		address = URI[6:]
+	default:
+		network = connector.TCPNetwork
+		address = URI
+	}
+
+	// In the case of a complex uri, shell expansion does not occur, so do it manually.
+	if network == connector.UnixNetwork &&
+		strings.HasPrefix(address, "~/") {
+		if homeDir, err := os.UserHomeDir(); err == nil {
```
psergee, line 179: I'd like to see a warning at least if something goes wrong. Unable to get user home dir in this case. But this one can be a debug type of message, so it will be printed only if verbose mode is enabled.

cli/cmd/connect.go
```diff
@@ -155,46 +205,75 @@ func parseCredentialsURI(str string) (string, string, string) {
...
-		uri, user, pass := parseCredentialsURI(newArgs[0])
-		newArgs[0] = uri
+		newURI, user, pass := parseCredentialsURI(args[0])
+		network, address := parseBaseURI(newURI)
 		connectCtx.Username = user
 		connectCtx.Password = pass
-		return newArgs, nil
-	} else if isBaseURI(newArgs[0]) {
+		connOpts = makeConnOpts(
+			network, address, connectCtx)
```
psergee, line 260: Please, be consistent with existing code style in `tt` codebase: break line before 100-th character. If I see `(` at the end of the line, I expect to see closing `)` on a separate line also.

psergee (approved):
lgtm

cli/cmd/connect.go
```diff
@@ -122,18 +137,53 @@ func isCredentialsURI(str string) bool {
...
 	uriReStr := "^((" + tcpReStr + ")|(" + unixReStr + ")|(" + pathReStr + "))$"
 	uriRe := regexp.MustCompile(uriReStr)
 	return uriRe.MatchString(str)
 }
 
+// parseBaseURI parses an URI and returns:
+// (network, address)
+func parseBaseURI(URI string) (string, string) {
+	var network, address string
+	URILen := len(URI)
```
oleg-jukovec, line 159: A variables that starts with a capital letter may be confusing in Go. Up to you.

```suggestion
// parseBaseURI parses an URIi and returns:
// (network, address)
func parseBaseURI(uri string) (string, string) {
	var network, address string
	uriLen := len(uri)
```

oleg-jukovec (approved):

PR #496, opened 2023-06-15
util: check for instances file in app dir

psergee:
It is possible to have an application without init.lua script. Script names can be specified using instances file. So if the directory contains instances file, it must be treated as a multi-instance application.

### DIFF
--- a/cli/util/util.go
+++ b/cli/util/util.go
@@ -691,19 +691,22 @@ func IsApp(path string) bool {
 		return false
 	}
 
-	if !entry.IsDir() && filepath.Ext(entry.Name()) != ".lua" {
-		return false
-	} else if !entry.IsDir() && filepath.Ext(entry.Name()) == ".lua" {
-		return true
+	if entry.IsDir() {
+		// Check if the directory contains init.lua script or instances.yml file.
+		for _, fileTocheck := range [...]string{"init.lua", "instances.yml", "instances.yaml"} {
+			if fileInfo, err := os.Stat(filepath.Join(path, fileTocheck)); err == nil {
+				if !fileInfo.IsDir() {
+					return true
+				}
+			}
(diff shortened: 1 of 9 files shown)


cli/util/util.go
```diff
@@ -691,19 +691,22 @@ func IsApp(path string) bool {
...
-		return true
-	}
-
-	if _, err = os.Stat(filepath.Join(path, "init.lua")); err == nil && entry.IsDir() {
-		return true
-	} else if entry.IsDir() {
-		return false
+	if entry.IsDir() {
+		// Check if the directory contains init.lua script or instances.yml file.
+		for _, fileTocheck := range [...]string{"init.lua", "instances.yml"} {
```
LeonidVas, line 696: Maybe also include `instances.yaml`?

test/integration/running/multi_inst_app_no_init/ie/mi_app/instances.yml
```diff
@@ -0,0 +1,4 @@
+router:
```
LeonidVas, line 1: AFAIU "mi" == "multi instalce", but what is "ie"?

LeonidVas (approved):
Hi! Thank you for the patch. Generally I'm ok.
If I understand correctly, this change affects the customers, so needs to add a note to CHANGELOG.

cli/util/util.go
psergee: `.yaml` extension support for instances file is added.

test/integration/running/multi_inst_app_no_init/ie/mi_app/instances.yml
psergee: Instances enabled. Or Internet Explorer O_o
Renamed.

PR #503, opened 2023-06-20
install: add ability to install tarantool from the local build directory

### LINKED ISSUE #429: Add the ability to install tarantool from local source tree
As a developer, I don't use released Tarantool versions. Instead I compile Tarantool by myself from a source tree. It'd be nice if `tt` had the ability to a add pseudo Tarantool installation with an arbitrary name that would point to a local source tree (via symlink) so that when I recompiled it, `tt run` would run the new binary without re-installation. I imagine something like this:

```
tt install tarantool-dev --src ~/src/tarantool/build/debug
make -j16 -C ~/src/tarantool/build/debug
tt run tarantool-dev  # runs the binary compiled above
```

This should work both for CE and EE Tarantool repositories (they store compiled binaries in different sub-directories).

Related issue: #292

askalt:
Made it via symlink to an existing tarantool binary. The binary file is searched in 2 paths taken from
`tarantool` and `tarantool-ee`.

Closes #429

@TarantoolBot document
Title: abitilty to install tarantool from the local build directory

Now, one can add a pseudo-installation of tarantool, which will point to the executable file in the local build directory, and then use this installation with `tt run` command. For example:
```
tt install tarantool-dev ~/src/tarantool/build
make -j16 -C ~/src/tarantool/build
tt run # runs the binary compiled above
```


cli/cmd/install.go
```diff
@@ -65,6 +65,29 @@ func newInstallTarantoolEeCmd() *cobra.Command {
...
 
+// newInstallTarantoolDevCmd creates a command to install tarantool
+// from the local build directory.
+func newInstallTarantoolDevCmd() *cobra.Command {
+	tntCmd := &cobra.Command{
+		Use:   "tarantool-dev dir",
+		Short: "Install tarantool from the local build directory",
+		Example: "Assume, tarantool build directory is ~/src/tarantool/build\n" +
+			"  Consider the following use case:\n\n" +
+			"  tt install tarantool-dev ~/src/tarantool/build\n" +
```
better0fdead, line 76: Will that scenario work? 
As i understand, compiled binary should be already placed at ~/src/tarantool/build.
Like :
 1. build binary
 2. tt install tarantool-dev

cli/install/install.go
```diff
@@ -1109,6 +1112,25 @@ func installTarantoolEE(binDir string, includeDir string, installCtx InstallCtx,
...
+// installTarantoolDev creates a symlink to the tarantool executable,
+// which is searched in buildDir.
+func installTarantoolDev(binDir string, buildDir string) error {
+	checkedPaths := make([]string, 0)
+	for _, binaryRelPath := range [...]string{"src/tarantool", "tarantool/src/tarantool"} {
+		binaryPath := filepath.Join(buildDir, binaryRelPath)
+		if util.IsRegularFile(binaryPath) {
+			err := util.CreateSymlink(binaryPath, filepath.Join(binDir, "tarantool"), true)
+			if err == nil {
+				log.Infof("Done.")
```
better0fdead, line 1124: Please make comment more informative. 
Or add a comment above "Searching for tarantool binary...."
"Tarantool binary found, creating symlink..."

better0fdead (changes requested):
Hi, thx for patch. Have some comments.
Pls, add a check that directory exists.
Pls, add an integration test for "tt install tarantool-dev".

cli/cmd/install.go
```diff
@@ -65,6 +65,29 @@ func newInstallTarantoolEeCmd() *cobra.Command {
 	return tntCmd
 }
 
+// newInstallTarantoolDevCmd creates a command to install tarantool
+// from the local build directory.
+func newInstallTarantoolDevCmd() *cobra.Command {
+	tntCmd := &cobra.Command{
+		Use:   "tarantool-dev dir",
```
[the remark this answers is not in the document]
askalt: Fixed.

cli/cmd/install.go
askalt: Yes, you're right, it was implied. Now the sample has been changed to be more clear.

psergee:
Thanks for the patch.
Please, check how `tt binaries` and `tt build`/`tt rocks install` (no system-wide tarantool installed) commands work with tarantool-dev binary "installed".

askalt:
Sample scenario with `tarantool-dev`:

Sample pipeline with `tarantool-dev`

cli/install/install.go
```diff
@@ -1109,6 +1114,88 @@ func installTarantoolEE(binDir string, includeDir string, installCtx InstallCtx,
...
+	} else {
+		// Check for default.
+		if util.IsDir(defaultIncPath) {
+			log.Infof("tarantool headers directory set as %v", defaultIncPath)
+			includeDir = defaultIncPath
+		}
+	}
+
+	if includeDir == "" {
+		log.Warn("Tarantool headers directory wasn't specified and" +
```
better0fdead, line 1149: need to add space between “and” && “was”

better0fdead: also u use "wasn't" and then "was not"
i think it would be prettier to use one variant

test/integration/binaries/test_binaries.py
```diff
@@ -59,7 +59,7 @@ def test_binaries_no_directory(tt_cmd, tmpdir):
 
 def test_binaries_empty_directory(tt_cmd, tmpdir):
     configPath = os.path.join(tmpdir, config_name)
-    os.mkdir(tmpdir+"/bin")
+    os.mkdir(tmpdir + "/bin")
```
better0fdead, line 62: that change doesn't belong to the commit :)

askalt: Fixed.

cli/install/install.go
askalt: Fixed.

cli/uninstall/uninstall.go
```diff
@@ -79,6 +80,27 @@ func UninstallProgram(program string, programVersion string, binDst string, head
...
+	if program == search.ProgramDev {
+		tarantoolBinary := filepath.Join(binDst, "tarantool")
+		binActive, err := util.ResolveSymlink(tarantoolBinary)
+		if err != nil {
+			return err
+		}
+		if !version.IsTarantoolDev(binActive, binDst) {
+			return fmt.Errorf("%s is no installed", program)
+		}
+		if err := os.Remove(tarantoolBinary); err != nil {
```
psergee, line 92: Can we name it tarantoolBinary -> tarantoolBinarySymlink, because `os.Remove(tarantoolBinary)` looks like we are removing binary, which is not true.

test/integration/install/test_install.py
```diff
@@ -160,3 +160,44 @@ def test_install_tarantool_in_docker(tt_cmd, tmpdir):
...
+
+
+def test_install_tarantool_dev(tt_cmd, tmpdir):
+    build_dirs = [
+        os.path.join(tmpdir, "build1"),
+        os.path.join(tmpdir, "build2")
+    ]
+    build_rel_paths = ['src', 'tarantool/src']
+
+    # Create imitations of tarantool binaries.
```
psergee, line 172: Add checks for include dir.
Check that symlink is updated if already exists for binary, for include dir. Check the scenario if `tarantool` exists in bin dir and is not a symlink. The same is for include dir.
Add test for `--include-dir` option and default include dir search. `--include-dir` pointing to non-existing dir, to the file scenarios.
Add test for uninstall. Cover different scenarios additionally: broken symlink for binary, for include dir, tarantool symlink pointing to binary in `bin` (should not be removed).

cli/uninstall/uninstall.go
```diff
@@ -79,6 +81,27 @@ func UninstallProgram(program string, programVersion string, binDst string, head
...
 	var err error
 
+	if program == search.ProgramDev {
+		tarantoolBinarySymlink := filepath.Join(binDst, "tarantool")
+		_, isTarantoolDevInstalled, err := install.IsTarantoolDev(tarantoolBinarySymlink, binDst)
+		if err != nil {
+			return err
+		}
+		if !isTarantoolDevInstalled {
+			return fmt.Errorf("%s is no installed", program)
```
psergee, line 91:
```suggestion
			return fmt.Errorf("%s is not installed", program)
```

psergee (approved):
lgtm

PR #519, opened 2023-07-06
ci: switch macOS tests fully to selfhosted

### LINKED ISSUE #517: investigate how to start docker on macOS self hosted runners.
Our macOS full-ci tests are spliced into 2 jobs. (tests w/wo docker)
Tests with docker run on GitHub CI.
Need to find a way to start docker on self hosted runners.

Main problem - https://github.com/docker/for-mac/issues/2359, https://github.com/docker/roadmap/issues/172 . Docker needs ui interface (accept license, click start...) to start.
That issue is not yet resolved.
Workaround that is mentioned in that issue is not working on our macOS self hosted runners. (e.g brew is not installing progs in default macOS folder.....)
Maybe we can try to use analog of docker for macOS. (e.g Colima)
That way causes some problems.
1 - need to replace docker.sock file to symlink to colima socket.
2 - log messages are different. (tests check logs and fails)
3 - also errors with permissions. (need investigation)

better0fdead:
Since docker daemon is now running on selfhosted
there is no need to use github runners anymore.
Switched full-ci macOS to selfhosted.
Closes #517

### DIFF
--- a/.github/workflows/full-ci.yml
+++ b/.github/workflows/full-ci.yml
@@ -117,7 +117,7 @@ jobs:
       (github.event_name == 'pull_request' &&
         github.event.action == 'synchronize' &&
         contains(github.event.pull_request.labels.*.name, 'full-ci'))
-    runs-on: [self-hosted, macOS-12-self-hosted, x86_64, regular]
+    runs-on: [self-hosted, macOS-13-self-hosted, x86_64, regular]
     steps:
       - uses: actions/checkout@master
         with:
--- a/.github/workflows/full-ci.yml
+++ b/.github/workflows/full-ci.yml
@@ -108,7 +108,7 @@ jobs:
       - name: Integration tests
         run: mage integrationfull
 
-  full-ci-macOS-no-docker:
+  full-ci-macOS:
     if: |
       (github.event_name == 'push') ||
       (github.event_name == 'pull_request' &&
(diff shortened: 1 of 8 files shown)


cli/running/running_test_helpers.go
```diff
@@ -6,5 +6,5 @@ import "time"
 func waitProcessStart() {
 	// We need to wait for the new process (tarantool instance) to set handlers.
 	// It is necessary to update for more correct synchronization.
-	time.Sleep(100 * time.Millisecond)
+	time.Sleep(1000 * time.Millisecond)
```
psergee, line 9: As I can see, we know the console socket path. Can we wait for the socket to appear instead of increasing timeout?

better0fdead: As far as I remember, socket appearance!=instance start. It is created before instance is fully started. So, I think we could check for socket + some timeout / stay as is.

.github/workflows/full-ci.yml
```diff
@@ -149,55 +149,14 @@ jobs:
...
-        env:
-          TT_CLI_BUILD_SSL: 'static'
-        run: mage build
-
-      - name: Install tarantool
-        run: |
-          brew install tarantool
-
-      - name: Integration tests with docker
-        run: mage integrationdocker
```
psergee, line 200: Do we still need `integrationdocker` in magefile and `docker` mark for tests?

PR #535, opened 2023-07-16
connect: expanded formatting modes

### LINKED ISSUE #162: Expanded formatting modes
The idea is to add an expanded formatting modes, see [psql](https://www.postgresql.org/docs/current/app-psql.html) and [space_object:fselect](https://github.com/tarantool/tarantool/commit/0dc72812fb78a192945612f0e954026a0ffe4053). We need to implement three output modes:

1. YAML (the default and the current one).
2. Table.
3. Table transposed (transpose (rotate) data from rows to columns).

The output of tables with meta information in a table mode should build a normal table.

For the convenience of the users, use a short and a long form of commands to switch modes:

1. `\xy` or `\set expanded yaml`
2. `\xt` or `\set expanded table`
3. `\xT` or `\set expanded ttable`

Additional commands:

1. `\xg` or `\set graphics off` disables the output of pseudographic characters (sometimes they interfere a
lot if long lines are output).
2. `\xG` or `\set graphics on` enable the output of pseudographic characters.
3. `\xw number` or `\set table_width number` sets up a max width of output string in a table mode.

Output of various entities in table modes:

* For a scalar value (number, string, or boolean value) consider it as a table with one row and one element in it. Column names do `column1`, `column2` etc.
* Display Lua-tables with names of positional columns as `column1`, `column2` etc. The rest by taking names from the table itself.
* A tuple should be processed as a regular table, if they do not have a format.
* A tuple with format should be processed as a table with column names instead of `column1`, `column2` etc.
* Decimal, datetime, and other values are reduced to a string representation. For datetime, we use the default format.

Additional cases:

* It is required to count the lengths of unicode strings in characters, not bytes.
* If a tuple or a table is displayed inside one cell of the table, then we output it all as a single-line YAML.

The task can be performed in the following sequence:

1. Add console commands and enums for it.
2. Add table mode support.
3. Add transpose table mode support.

GRISHNOV:
The patch introduces formatting modes for the interactive console [[1](https://www.notion.so/tarantool/TT-connect-table-output-format-like-fselect-70bd44d1f77b40cf87eddb357adaf37a)].

Closes #162

@TarantoolBot document
Title: Expanded formatting modes

The patchset introduces addition commands for the `tt connect` console.

It adds several common commands that relate to a general functionality:

* `\help`, `?` - to show help message with additional information and a list of all commands.
* `\quit`, `\q` - quit from the console. Previously this command worked on the `tarantool` side, but now it works on the `tt` side and does not require execution rights for a user.

In addition, the patch adds commands for extended output modes:
    
* `\set output <format>` - set format lua, table, ttable or yaml (default)
* `\set table_format <format>` - set table format default, jira or markdown
* `\set graphics <false/true>` - disables/enables (default)  pseudographics for table modes
* `\set table_column_width <width>` - set max column width for table/ttable
* `\xw <width>` - set max column width for table/ttable
* `\x` - switches output format cyclically
* `\x[l,t,T,y]` - set output format lua, table, ttable or yaml
* `\x[g,G]` - disables/enables pseudographics for table modes
    
Supported formats:
    
1. `yaml` - shows data as a YAML document. The format is used now by `tarantool` and `tt`.
2. `lua` - shows data as Lua-compatible structs.
3. `table` - shows data as a pseudo graphical table.
4. `ttable` - the same as the table format, but a result table is transposed (rotate data from rows to columns).

Supported table formats (for `table` or `ttable` formats):

1. `default` - shows data as a pseudo graphical table.
2. `markdown` - shows data as a Markdown-compatible table.
3. `jira` - shows data as a Jira-compatible table.

Example:

```
> \set output table
> box.space.customers:select()
+------+-----------+------+
| col1 | col2      | col3 |
+------+-----------+------+
| 1    | Elizabeth | 12   |
+------+-----------+------+
> \xg
> box.space.customers:select()
 col1  col2       col3 
 1     Elizabeth  12   
 2     Mary       46   
 3     David      33
```


cli/connect/interactive.go
```diff
@@ -0,0 +1,244 @@
...
+  \set language <language>   -- set language lua or sql
+  \set output <format>       -- set output format yaml, lua, table or ttable
+  \x[y,l,t,T]                -- set output format yaml, lua, table or ttable
+  \x                         -- switches output format cyclically
+  \x[g,G]                    -- disables/enables pseudographics for tables
+  \set table_column_width    -- max column width for table/ttable
+  \xw                        -- max column width for table/ttable
+  \table_format              -- tables format markdown, jira or default
+  \help                      -- show this screen
+  \quit                      -- quit interactive console
```
oleg-jukovec, line 55: It will be clearer:
```suggestion
  \set language <language>   -- set language lua or sql
  \set output <format>       -- set output format yaml, lua, table or ttable
  \x[y,l,t,T]                -- set output format yaml, lua, table or ttable
  \x                         -- switches output format cyclically
  \x[g,G]                    -- disables/enables pseudographics for tables
  \set table_column_width <width>    -- max column width for table/ttable
  \xw                        -- max column width for table/ttable
  \table_format <format>             -- tables format markdown, jira or default
  \help                      -- show this screen
  \quit                      -- quit interactive console
```

cli/connect/output.go
```diff
@@ -0,0 +1,110 @@
...
+	luaOutputFormatStr         = "lua"
+	luaOutputFormatShortStr    = "l"
+	tableOutputFormatStr       = "table"
+	tableOutputFormatShortStr  = "t"
+	ttableOutputFormatStr      = "ttable"
+	ttableOutputFormatShortStr = "T"
+)
+
+// OutputFormat defines a set of supported output format.
+type OutputFormat int
```
oleg-jukovec, line 22:
```suggestion
// Format defines a set of supported output format.
type Format int
```
+ rename the file to `format.go`. We don't lose anything and the code will become compact.

cli/formatter/formatter.go
```diff
@@ -0,0 +1,778 @@
...
+)
+
+const (
+	defaultTableDialectStr  = "default"
+	markdownTableDialectStr = "markdown"
+	jiraTableDialectStr     = "jira"
+)
+
+// String returns a string representation of the table dialect.
+func (f TableDialect) String() string {
```
oleg-jukovec, line 42: There are no tessts for the public API.

cli/formatter/formatter.go
```diff
@@ -0,0 +1,778 @@
...
+	case JiraTableDialect:
+		return jiraTableDialectStr
+	default:
+		panic("Unknown table dialect")
+	}
+}
+
+// ParseTableDialect parses a table dialect string representation. It supports
+// mixed case letters.
+func ParseTableDialect(str string) (TableDialect, bool) {
```
oleg-jukovec, line 57: There are no tessts for the public API.

cli/formatter/formatter.go
```diff
@@ -0,0 +1,778 @@
...
+// Opts contains formatting options.
+type Opts struct {
+	TransposeMode bool
+	NoGraphics    bool
+	ColWidthMax   int
+	TableDialect  TableDialect
+}
+
+// TableDialect defines a set of supported table dialect.
+type TableDialect int
```
oleg-jukovec, line 27: It's confusing to see `OutputFormat` in the `connect/output.go`, but `TableDialect` at `formatter/formatter.go` strange to see here. I see it as abstractions of the same level and the same responsibility. So I suggest to move `OutputFormat` to `formatter/format.go`.

oleg-jukovec, line 27: We could move the type into a separate file.

cli/formatter/formatter.go
```diff
@@ -0,0 +1,778 @@
...
+	"gopkg.in/yaml.v2"
+)
+
+// Opts contains formatting options.
+type Opts struct {
+	TransposeMode bool
+	NoGraphics    bool
+	ColWidthMax   int
+	TableDialect  TableDialect
+}
```
oleg-jukovec, line 24: We could move the file into a separate file.

cli/formatter/formatter.go
```diff
@@ -0,0 +1,778 @@
...
+	default:
+		if elem == nil {
+			return "nil"
+		}
+		return fmt.Sprintf("%v", elem)
+	}
+}
+
+// RenderLua renders lua string from yaml string.
+func RenderLua(inputYaml string) {
```
oleg-jukovec, line 757: There are no tests for the public API.

cli/formatter/formatter.go
```diff
@@ -0,0 +1,778 @@
...
+		}
+
+		return nil
+	}
+
+	return nil
+}
+
+// RenderTables renders tables for table/ttable output formats.
+func RenderTables(inputYaml string, formatterOpts *Opts) {
```
oleg-jukovec, line 674: There are no tests for the public API.

oleg-jukovec, line 674: I suggest do not print the result to the `stdout`, but return as a string. It's easier to test and there are fewer potential problems in the future.
```suggestion
// RenderTables renders tables for table/ttable output formats.
func RenderTables(inputYaml string, formatterOpts *Opts) string {
```

cli/formatter/formatter.go
oleg-jukovec, line 757: I suggest do not print the result to the stdout, but return as a string [2]
```suggestion
func RenderLua(inputYaml string) string {
```

cli/formatter/table/README.md
```diff
@@ -0,0 +1,5 @@
+# formatter/table
+
+Based on github.com/jedib0t/go-pretty
```
oleg-jukovec, line 3: `formatter/table`&`formatter/text`:

After discussion with @LeonidVas and @psergee we decided that we need to do it in a different way:

1. Create a patch for the latest release of the `go-pretty`. The patch should be as small as possible.
2. Add `go-pretty` as a git submodule.
3. Patch the submodule with the patch on the build step.

See how it is already done with cartridge-cli in the repo.

cli/connect/interactive.go
```diff
@@ -0,0 +1,244 @@
...
+  Available backslash commands:
+
+  \set language <language>        -- set language lua or sql
+  \set output <format>            -- set output format yaml, lua, table, ttable
+  \x[y,l,t,T]                     -- set output format yaml, lua, table, ttable
+  \x                              -- switches output format cyclically
+  \x[g,G]                         -- disables/enables pseudographics for tables
+  \set table_column_width <width> -- max column width for table/ttable
+  \xw                             -- max column width for table/ttable
+  \table_format <format>          -- tables format markdown, jira or default
```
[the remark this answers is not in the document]
GRISHNOV: Current implementation correspond to approved RFC

cli/connect/interactive.go
GRISHNOV: Done

cli/connect/interactive.go
```diff
@@ -0,0 +1,244 @@
...
+  \help                           -- show this screen
+  \quit                           -- quit interactive console
+  `
+	fmt.Println(help)
+}
+
+// languageOptionHeandler handles language interactive option.
+func languageOptionHeandler(trimmed string, console *Console) {
```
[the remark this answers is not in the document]
GRISHNOV: Done

cli/connect/interactive.go
```diff
@@ -0,0 +1,244 @@
...
+const setPseudoGraphicsDisable = "\\xg"
+
+// setTableColWidthShortPrefix is a short prefix for table column width interactive option.
+const setTableColWidthShortPrefix = "\\xw "
+
+// setTableColWidthShortPrefix is a prefix for table column width interactive option.
+const setTableColWidthPrefix = "\\set table_column_width "
+
+// helpOptionHeandler prints help of tt connect interactive option.
+func helpOptionHeandler() {
```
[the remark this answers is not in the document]
GRISHNOV: Done

cli/connect/console.go
```diff
@@ -167,20 +171,18 @@ func (console *Console) Close() {
 }
 
 func getExecutor(console *Console) prompt.Executor {
+	var executerEval string = getExecutorEval(console.outputFormat)
```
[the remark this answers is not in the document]
GRISHNOV: Done

cli/connect/interactive.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+// interactiveOptionDriver handles slash options for interactive console and returns
+// true if slash options detected in user input.
+func interactiveOptionDriver(in string, executerEval *string,
+	console *Console, formatterOpts *formatter.Opts) bool {
+	trimmed := strings.TrimSpace(in)
+
+	// helper case handling
+	if trimmed == "\\help" || trimmed == "help" ||
```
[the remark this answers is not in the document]
GRISHNOV: Done. Cannot be fully implemented, but this moment is not critical

```
localhost:3333> help = 123
+------+
| col1 |
+------+
|      |
+------+
localhost:3333> help + help
+------+
| col1 |
+------+
| 246  |
+------+
```

cli/connect/interactive.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+}
+
+// interactiveOptionDriver handles slash options for interactive console and returns
+// true if slash options detected in user input.
+func interactiveOptionDriver(in string, executerEval *string,
+	console *Console, formatterOpts *formatter.Opts) bool {
+	trimmed := strings.TrimSpace(in)
+
+	// helper case handling
```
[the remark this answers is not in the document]
GRISHNOV: Done

cli/connect/output.go
```diff
@@ -0,0 +1,110 @@
...
+// ParseOutputFormat parses a output format string representation. It supports
+// mixed case letters.
+func ParseOutputFormat(str string) (OutputFormat, bool) {
+	switch str {
+	case defaultOutputFormatStr:
+		return DefaultOutputFormat, true
+	case yamlOutputFormatStr:
+		return YamlOutputFormat, true
+	case yamlOutputFormatShortStr:
+		return YamlOutputFormat, true
```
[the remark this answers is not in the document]
GRISHNOV: Done

cli/connect/output.go
```diff
@@ -0,0 +1,110 @@
...
+	return consoleEvalFuncBody
+}
+
+// renderOutput renders output depending on specified output format and formatting options.
+func renderOutput(data string, outputFormat OutputFormat, formatterOpts *formatter.Opts) {
+	switch outputFormat {
+	case DefaultOutputFormat:
+		fmt.Printf("%s\n", data)
+	case YamlOutputFormat:
+		fmt.Printf("%s\n", data)
```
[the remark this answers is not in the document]
GRISHNOV: Done

cli/connect/output.go
GRISHNOV: Done

cli/formatter/formatter.go
GRISHNOV: Done

GRISHNOV: Done

cli/formatter/formatter.go
GRISHNOV: > We could move the file into a separate file.

Impossible. Done

cli/formatter/formatter.go
GRISHNOV: Done

cli/formatter/formatter.go
GRISHNOV: Done

cli/connect/output.go
```diff
@@ -0,0 +1,110 @@
...
+	DefaultOutputFormat OutputFormat = iota
+	YamlOutputFormat
+	LuaOutputFormat
+	TableOutputFormat
+	TTableOutputFormat
+)
+
+// ParseOutputFormat parses a output format string representation. It supports
+// mixed case letters.
+func ParseOutputFormat(str string) (OutputFormat, bool) {
```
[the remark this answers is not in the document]
GRISHNOV: Now there are unit-tests for the public API

cli/connect/output.go
```diff
@@ -0,0 +1,110 @@
...
+	case ttableOutputFormatStr:
+		return TTableOutputFormat, true
+	case ttableOutputFormatShortStr:
+		return TTableOutputFormat, true
+	}
+	return DefaultOutputFormat, false
+}
+
+// String returns a string representation of the output format.
+func (l OutputFormat) String() string {
```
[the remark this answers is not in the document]
GRISHNOV: Now there are tests for the public API

cli/formatter/formatter.go
GRISHNOV: Now there are tessts for the public API

cli/formatter/formatter.go
GRISHNOV: Now there are tessts for the public API

cli/formatter/formatter.go
GRISHNOV: Made private

cli/formatter/formatter.go
GRISHNOV: Made private

cli/formatter/table/README.md
GRISHNOV: Done, implemented by analogy with the `cartridge-cli`

cli/formatter/format.go
```diff
@@ -0,0 +1,752 @@
...
+			columnConfigs = append(columnConfigs,
+				table.ColumnConfig{
+					Number:      i,
+					Transformer: colWidthTransformer,
+					WidthMax:    opts.ColWidthMax,
+				},
+			)
+		}
+		t.SetColumnConfigs(columnConfigs)
+	}
```
oleg-jukovec, line 525: Could you please explain the code with `WidthMax`. Does `t.GetColumnConfig()[0].WidthMax` == `opts.ColWidthMax`?
If this is not the case, then what values can be `t.GetColumnConfig()[0].WidthMax`?

cli/formatter/go-pretty/extra/003_add_no_graphics_style.patch
```diff
@@ -0,0 +1,39 @@
...
++		PaddingLeft:      " ",
++		PaddingRight:     " ",
++		PageSeparator:    "\n",
++		Right:            " ",
++		RightSeparator:   " ",
++		TopLeft:          " ",
++		TopRight:         " ",
++		TopSeparator:     " ",
++		UnfinishedRow:    "  ",
++	}
```
oleg-jukovec, line 35: Can we define `StyleWithoutGraphics` in client code without the patch? Why not?

cli/formatter/go-pretty/extra/002_add_interface_for_001.patch
```diff
@@ -0,0 +1,20 @@
+diff --git a/table/writer.go b/table/writer.go
```
oleg-jukovec, line 1: Let's merge the patch with `001.patch`. It looks like a single feature.

cli/formatter/go-pretty/extra/001_add_transposing_and_colconf.patch
```diff
@@ -0,0 +1,53 @@
...
++	for i := 0; i < rowsRawTransposedCap; i++ {
++		var rowTransposed Row
++		for j := 0; j < len(t.rowsRaw); j++ {
++			rowTransposed = append(rowTransposed, t.rowsRaw[j][i])
++		}
++		rowsRawTransposed = append(rowsRawTransposed, rowTransposed)
++	}
++
++	t.rowsRaw = rowsRawTransposed
++}
```
oleg-jukovec, line 37: Can we transpose lines in the client code without patching the library before appending it into a table? Why can't this be done? I don't see the reason at first glance.

oleg-jukovec (approved):
Thank you for the update, I have a couple of questions about the `go-pretty` patching.

oleg-jukovec (commented):
Approved by mistake.

cli/formatter/format.go
GRISHNOV: Simplified, now using just `opts.ColWidthMax`, it is the same

cli/formatter/go-pretty/extra/003_add_no_graphics_style.patch
GRISHNOV: Moved to `formatter/style.go`

cli/formatter/go-pretty/extra/002_add_interface_for_001.patch
GRISHNOV: Merged

cli/formatter/go-pretty/extra/001_add_transposing_and_colconf.patch
GRISHNOV: Unfortunately not, this method works with private entities (`t.rowsRaw`)

oleg-jukovec: If we have already appended rows. But what about transpose before appending? The idea in pseudo-code: 

```Go
var []rows table.Row

// Build rows slice code.
// ...
// Done.

if opts.TransposeMode {
    rows := someTransposeRowsMethod(rows)
}

t := table.NewWriter()
for _, r := range rows {
    t.AppendRow(r)
}
```
or
```Go
var []srcRows someType

// Build srcRows slice code.
// ...
// Done.

if opts.TransposeMode {
    srcRows := someTransposeRowsMethod(srcRows)
}

t := table.NewWriter()
for _, r := range rows {
    var row table.Row
    row := srcToRow(r)   
    t.AppendRow(row)
}
```

GRISHNOV: I thought about it, unfortunately it is impossible: the logic of vertical tables merging and the design of filling the table with rows does not allow this operation to be performed earlier

We can try to upstream this method later

oleg-jukovec: > the logic of vertical tables merging and the design of filling the table with rows does not allow this operation to be performed earlier

Could you please give an example with the problem?

GRISHNOV: When the table is transposed, the rows become columns. We need to have a complete set of rows to perform the transposition. And the addition to the table is performed not by columns, but by rows (otherwise it would really be simple)

oleg-jukovec: What wrong with the patch?
```patch
--- a/cli/formatter/format.go
+++ b/cli/formatter/format.go
@@ -214,7 +214,9 @@ func renderMapsWithEqualKeys(maps []map[string]renderNode, opts *Opts) (string,
     sort.Strings(commonKeys)
 
     t := table.NewWriter()
-    t.AppendRow(constructHeaderRowForMap(commonKeys))
+    var rows []table.Row
+
+    rows = append(rows, constructHeaderRowForMap(commonKeys))
 
     for _, mapVal := range maps {
         var rowVals table.Row
@@ -238,14 +240,16 @@ func renderMapsWithEqualKeys(maps []map[string]renderNode, opts *Opts) (string,
                 rowVals = append(rowVals, string(jsonVal))
             }
         }
-        t.AppendRow(rowVals)
+        rows = append(rows, rowVals)
     }
 
     t.Style().Options.SeparateRows = true
-    var rowsAmount = t.Length()
+    var rowsAmount = len(rows)
     if opts.TransposeMode {
-        t.TransposeAppendedRows()
+        rows = transposeRows(rows)
     }
+    t.AppendRows(rows)
+
     if opts.NoGraphics {
         t.SetStyle(table.Style{Box: StyleWithoutGraphics})
     }
@@ -363,8 +367,9 @@ func constructMarkdownEmptyRow(totalRowAmount int) string {
 // renderSingleColumnTable returns table with single column as string.
 func renderSingleColumnTable(renderBatch []renderNode, opts *Opts) string {
     t := table.NewWriter()
+
     var rows []table.Row
-    t.AppendRow(table.Row{"col1"})
+    rows = append(rows, table.Row{"col1"})
     for _, node := range renderBatch {
         switch node.(type) {
         case float64, float32:
@@ -375,12 +380,14 @@ func renderSingleColumnTable(renderBatch []renderNode, opts *Opts) string {
             rows = append(rows, []interface{}{fmt.Sprintf("%v", node)})
         }
     }
-    t.AppendRows(rows)
-    var rowsAmount = t.Length()
-    t.Style().Options.SeparateRows = true
+
+    var rowsAmount = len(rows)
     if opts.TransposeMode {
-        t.TransposeAppendedRows()
+        rows = transposeRows(rows)
     }
+    t.AppendRows(rows)
+
+    t.Style().Options.SeparateRows = true
     if opts.NoGraphics {
         t.SetStyle(table.Style{Box: StyleWithoutGraphics})
     }
@@ -452,12 +459,41 @@ func castRenderNodeToMap(node renderNode) (map[string]renderNode, error) {
     return castedRes, nil
 }
 
+func transposeRows(rowsRaw []table.Row) []table.Row {
+    var rowsRawTransposedCap = 0
+    for _, row := range rowsRaw {
+        if len(row) > rowsRawTransposedCap {
+            rowsRawTransposedCap = len(row)
+        }
+    }
+
+    for rowNum, row := range rowsRaw {
+        if len(row) < rowsRawTransposedCap {
+            for i := 0; i < rowsRawTransposedCap-len(row); i++ {
+                rowsRaw[rowNum] = append(rowsRaw[rowNum], "")
+            }
+        }
+    }
+
+    var rowsRawTransposed []table.Row
+
+    for i := 0; i < rowsRawTransposedCap; i++ {
+        var rowTransposed table.Row
+        for j := 0; j < len(rowsRaw); j++ {
+            rowTransposed = append(rowTransposed, rowsRaw[j][i])
+        }
+        rowsRawTransposed = append(rowsRawTransposed, rowTransposed)
+    }
+
+    return rowsRawTransposed
+}
+
 // renderArrs returns table as string for arrays case.
 func renderArrs(renderBatch []renderNode, opts *Opts) (string, error) {
     t := table.NewWriter()
-    t.AppendRow(constructHeaderRowForArrs(renderBatch))
-
     var rows []table.Row
+
+    rows = append(rows, constructHeaderRowForArrs(renderBatch))
     for _, node := range renderBatch {
         nodeBytes, err := yaml.Marshal(node)
         if err != nil {
@@ -489,12 +525,14 @@ func renderArrs(renderBatch []renderNode, opts *Opts) (string, error) {
         }
         rows = append(rows, row)
     }
-    t.AppendRows(rows)
+
     t.Style().Options.SeparateRows = true
-    var rowsAmount = t.Length()
+    var rowsAmount = len(rows)
     if opts.TransposeMode {
-        t.TransposeAppendedRows()
+        rows = transposeRows(rows)
     }
+    t.AppendRows(rows)
+
     if opts.NoGraphics {
         t.SetStyle(table.Style{Box: StyleWithoutGraphics})
     }
```

GRISHNOV: Done, thanks

cli/connect/console.go
```diff
@@ -233,7 +244,7 @@ func getExecutor(console *Console) prompt.Executor {
 			data = results[0]
 		}
 
-		fmt.Printf("%s\n", data)
+		fmt.Printf("%s", formatter.MakeOutput(data, console.outputFormat, formatterOpts))
```
oleg-jukovec, line 247:
```suggestion
		fmt.Print(formatter.MakeOutput(data, console.outputFormat, formatterOpts))
```

cli/connect/interactive.go
```diff
@@ -0,0 +1,245 @@
...
+	if strings.HasPrefix(trimmed, setFormatPrefix) {
+		newFormat = strings.TrimPrefix(trimmed, setFormatPrefix)
+	}
+	if strings.HasPrefix(trimmed, setFormatShortPrefix) {
+		if trimmed == setFormatShortPrefix {
+			if console.outputFormat == formatter.DefaultFormat {
+				console.outputFormat = formatter.YamlFormat
+				formatterOpts.TransposeMode = formatter.IsTTableFormat(console.outputFormat)
+			}
+			console.outputFormat = 1 + (console.outputFormat % 4)
```
oleg-jukovec, line 132: Please, change `4` here to a constant in the `formatter` module.
```suggestion
			console.outputFormat = 1 + (console.outputFormat % 4)
```

cli/connect/interactive.go
```diff
@@ -0,0 +1,245 @@
...
+		}
+		log.Infof("Selected table dialect: %s", formatterOpts.TableDialect.String())
+	} else {
+		log.Warnf("Unsupported table dialect: %s", newTableDialect)
+	}
+}
+
+// handleInteractiveOption handles slash options for interactive console and returns
+// true if slash options detected in user input.
+func handleInteractiveOption(in string, executorEval *string,
```
oleg-jukovec, line 174: This code can be organized more simply. The idea:

1. Create a map: `option(s)_name[handle_function]`.
2.
```Go
for option, fun := range option_map {
     if strings.HasPrefix(trimmed, option) {
           fun(trimmied, console, opts)
     }
}
```

cli/formatter/format.go
```diff
@@ -0,0 +1,788 @@
...
+	"unicode/utf8"
+
+	"github.com/apex/log"
+	"github.com/jedib0t/go-pretty/v6/table"
+	"github.com/jedib0t/go-pretty/v6/text"
+	"gopkg.in/yaml.v2"
+)
+
+// MakeOutput returns formatted output depending on specified output format and formatting options.
+func MakeOutput(data string, outputFormat Format, opts *Opts) string {
```
oleg-jukovec, line 19: Does `MakeOutput` works in fact with a different input data format? We have `console_eval_func_body_tbls_fmt.lua` and `console_eval_func_body_fmt.lua`. Or the input data format the same?

cli/formatter/format.go
```diff
@@ -0,0 +1,788 @@
...
+	switch outputFormat {
+	case DefaultFormat, YamlFormat:
+		return fmt.Sprintf("%s\n", data)
```
oleg-jukovec, line 22:
```suggestion
		return fmt.Sprintln(data) // or data + '\n'
```

cli/formatter/format.go
```diff
@@ -0,0 +1,788 @@
...
+	case LuaFormat:
+		return fmt.Sprintf("%s\n", getRenderedLua(data))
```
oleg-jukovec, line 24:
```suggestion
		return fmt.Sprintln(getRenderedLua(data)) // or getRenderedLua(data) + '\n'
```

cli/formatter/format.go
```diff
@@ -0,0 +1,788 @@
...
+	case TableFormat, TTableFormat:
+		return fmt.Sprintf("%s", getRenderedTables(data, opts))
```
oleg-jukovec, line 26:
```suggestion
		return getRenderedTables(data, opts)
```

cli/formatter/format.go
oleg-jukovec, line 19: We don't have tests for the public API.

cli/formatter/format.go
```diff
@@ -0,0 +1,788 @@
...
+	default:
+		panic("Unknown render case")
+	}
+}
+
+// IsTTableFormat returns true if outputFormat is TTableFormat.
+func IsTTableFormat(outputFormat Format) bool {
```
oleg-jukovec, line 33: We don't have tests for the public API.

cli/formatter/format.go
```diff
@@ -0,0 +1,788 @@
...
+		for k, v2 := range x {
+			x[k] = convertMapIntrfToMapStr(v2)
+		}
+	}
+
+	return v
+}
+
+// getRenderNodeType returns renderNode type.
+func getRenderNodeType(node renderNode) int {
```
oleg-jukovec, line 171:
```suggestion
func getRenderNodeType(node renderNode) nodeType {
```

cli/formatter/format.go
```diff
@@ -0,0 +1,788 @@
...
+		for i := 1; i <= rowsAmount; i++ {
+			columnConfigs = append(columnConfigs,
+				table.ColumnConfig{
+					Number:      i,
+					Transformer: colWidthTransformer,
+					WidthMax:    opts.ColWidthMax,
+				},
+			)
+		}
+		t.SetColumnConfigs(columnConfigs)
```
oleg-jukovec, line 445: It seems that the code used in 3 places, please, extract it into a separate function.

cli/formatter/format.go
```diff
@@ -0,0 +1,788 @@
...
+			markdownTable = insertStrToStrSlice(markdownTable, 0,
+				constructMarkdownEmptyRow(rowsAmount))
+			markdownTable = insertStrToStrSlice(markdownTable, 1,
+				constructMarkdownHeaderSeparatorRow(rowsAmount))
+		}
+		var res string
+		for _, v := range markdownTable {
+			res = res + v + "\n"
+		}
+		return res + "\n"
```
oleg-jukovec, line 465: It seems that the code used in 3 places, please, extract it into a separate function.

cli/formatter/format.go
```diff
@@ -0,0 +1,788 @@
...
+			} else {
+				// Encoding as json value for map/array table cell cases.
+				jsonVal, err := json.Marshal(convertMapIntrfToMapStr(rowColVal))
+				if err != nil {
+					return "", err
+				}
+				row = append(row, string(jsonVal))
+			}
+		}
+		rows = append(rows, row)
```
oleg-jukovec, line 525: It seems that the same code used in 2 places, please, extract it into a separate function.

cli/formatter/types.go
```diff
@@ -0,0 +1,120 @@
...
+	mapRenderNodeType
+)
+
+// Opts contains formatting options.
+type Opts struct {
+	TransposeMode bool
+	NoGraphics    bool
+	ColWidthMax   int
+	TableDialect  TableDialect
+}
```
oleg-jukovec, line 18: It would be easier to view the package if each type will be in a separate file + a separate file for tests.
To `opts.go`.

cli/formatter/types.go
```diff
@@ -0,0 +1,120 @@
...
+	case LuaFormat:
+		return luaFormatStr
+	case TableFormat:
+		return tableFormatStr
+	case TTableFormat:
+		return ttableFormatStr
+	default:
+		panic("Unknown output format")
+	}
+}
```
oleg-jukovec, line 77: To `format.go`.

cli/formatter/types.go
```diff
@@ -0,0 +1,120 @@
...
+	case DefaultTableDialect:
+		return defaultTableDialectStr
+	case MarkdownTableDialect:
+		return markdownTableDialectStr
+	case JiraTableDialect:
+		return jiraTableDialectStr
+	default:
+		panic("Unknown table dialect")
+	}
+}
```
oleg-jukovec, line 120: To `dialect.go`.

cli/formatter/types.go
```diff
@@ -0,0 +1,120 @@
...
+	case tableFormatStr, tableFormatShortStr:
+		return TableFormat, true
+	case ttableFormatStr, ttableFormatShortStr:
+		return TTableFormat, true
+	}
+	return DefaultFormat, false
+}
+
+// String returns a string representation of the output format.
+func (l Format) String() string {
```
oleg-jukovec, line 62: https://github.com/golang/go/wiki/CodeReviewComments#receiver-names
```suggestion
func (f Format) String() string {
```

cli/formatter/types.go
```diff
@@ -0,0 +1,120 @@
...
+	case markdownTableDialectStr:
+		return MarkdownTableDialect, true
+	case jiraTableDialectStr:
+		return JiraTableDialect, true
+	}
+	return DefaultTableDialect, false
+}
+
+// String returns a string representation of the table dialect.
+func (f TableDialect) String() string {
```
oleg-jukovec, line 109: https://github.com/golang/go/wiki/CodeReviewComments#receiver-names
```suggestion
func (d TableDialect) String() string {
```

oleg-jukovec (commented):
I'm glad to see that we could resume without the `go-pretty` patching. Let's now focus on code style.

cli/connect/console.go
GRISHNOV: Done

cli/connect/interactive.go
GRISHNOV: Done

cli/connect/interactive.go
GRISHNOV: Hmm, unfortunately, the detection of several options is not performed only through `HasPrefix`, plus, some options have a check for emptiness. Maybe it's worth to split `handleInteractiveOption` to smaller functions?

cli/formatter/format.go
GRISHNOV: We get yaml from the server all the time (server works only with `\set output yaml` all the time).
Then, depending on the `outputFormat Format`, a specific render is made by MakeOutput. 

Formats can use their own logic in files like `console_eval_func_body_tbls_fmt.lua` or `console_eval_func_body`, all this points is agreed with the product team.
Through such files, among other things, a compatibility layer will be implemented later for named headers for tuples (cdata) when working with tarantool ver < than 3.

cli/formatter/format.go
GRISHNOV: Done

cli/formatter/format.go
GRISHNOV: Done

cli/formatter/format.go
GRISHNOV: Done

cli/formatter/format.go
GRISHNOV: Done

cli/formatter/format.go
GRISHNOV: Done

cli/formatter/format.go
GRISHNOV: Done

cli/formatter/format.go
GRISHNOV: Done

cli/formatter/format.go
GRISHNOV: Done

cli/formatter/types.go
GRISHNOV: Done

cli/formatter/types.go
GRISHNOV: Done

cli/formatter/types.go
GRISHNOV: Done

cli/formatter/types.go
GRISHNOV: Done

cli/connect/interactive.go
oleg-jukovec: > Hmm, unfortunately, the detection of several options is not performed only through `HasPrefix`

As example?

cli/formatter/format_test.go
```diff
@@ -0,0 +1,166 @@
...
+			if c.panic {
+				f := func() { _ = c.tableDialect.String() }
+				assert.PanicsWithValue(t, "Unknown table dialect", f)
+			} else {
+				result := c.tableDialect.String()
+				assert.Equal(t, c.expected, result, "Unexpected result")
+			}
+		})
+	}
+}
```
oleg-jukovec, line 113: -> `dialect_test.go`

cli/formatter/format.go
```diff
@@ -0,0 +1,800 @@
...
+		return tableFormatStr
+	case TTableFormat:
+		return ttableFormatStr
+	default:
+		panic("Unknown output format")
+	}
+}
+
+// MakeOutput returns formatted output depending on specified output format and formatting options.
+func MakeOutput(data string, outputFormat Format, opts *Opts) string {
```
oleg-jukovec, line 78: The `MakeOutput` function does not change opts.

```suggestion
func MakeOutput(data string, outputFormat Format, opts Opts) string {
```

cli/connect/interactive.go
oleg-jukovec: If I add a new format, then it is necessary to change the constant in the different module. This is not good, the changes should be in one place - formatter module.

cli/formatter/format.go
oleg-jukovec: This is a poor test that does not cover all formatting cases.

```Go

func TestFormatter_MakeOutputFormat(t *testing.T) {
	cases := []struct {
		outputFormat formatter.Format
		panic        bool
	}{
		{formatter.DefaultFormat, false},
		{formatter.YamlFormat, false},
		{formatter.LuaFormat, false},
		{formatter.TableFormat, false},
		{formatter.TTableFormat, false},
		{2023, true},
	}

	for _, c := range cases {
		t.Run(fmt.Sprint(c.outputFormat), func(t *testing.T) {
			formatterOpts := &formatter.Opts{
				TransposeMode: false,
				NoGraphics:    false,
				ColWidthMax:   0,
				TableDialect:  formatter.DefaultTableDialect,
			}
			if c.panic {
				f := func() {
					_ = formatter.MakeOutput("--- ['']\n...\n",
						c.outputFormat, formatterOpts)
				}
				assert.PanicsWithValue(t, "Unknown render case", f)
			} else {
				formatter.MakeOutput("--- ['']\n...\n", c.outputFormat, formatterOpts)
			}
		})
	}
}
```

GRISHNOV: The `MakeOutput` function always accepts a yaml formatted string from the server. But the logic of forming this yaml string depends on lua files above

cli/formatter/format.go
```diff
@@ -0,0 +1,800 @@
...
+					rowVals = append(rowVals, fmt.Sprintf("%v", mapVal[commonKey]))
+				}
+			} else {
+				// Encoding as json value for map/array table cell cases.
+				jsonVal, err := encodeToCellJsonVal(mapVal[commonKey])
+				if err != nil {
+					return "", err
+				}
+				rowVals = append(rowVals, jsonVal)
+			}
```
oleg-jukovec, line 388: The duplication is still be here.

cli/formatter/format.go
```diff
@@ -0,0 +1,800 @@
...
+				}
+			} else {
+				// Encoding as json value for map/array table cell cases.
+				jsonVal, err := encodeToCellJsonVal(rowColVal)
+				if err != nil {
+					return "", err
+				}
+				row = append(row, jsonVal)
+			}
+		}
```
oleg-jukovec, line 571: The duplication is still be here.

cli/formatter/format.go
oleg-jukovec: Could I use `console_eval_func_body.lua` result for `TableFormat` and `TTableFormat` and `console_eval_func_body_tbls_fmt.lua` for `DefaultFormat`, `YamlFormat` and `LuaFormat`? Why not?

GRISHNOV:
https://github.com/tarantool/tt/pull/535#discussion_r1272342112

But it covered by integration tests, why duplicate?

cli/formatter/format.go
GRISHNOV: Note: At the moment, it has been discussed separately along with the unit test cases to MakeOutput

GRISHNOV: Added

cli/formatter/format_test.go
GRISHNOV: Done, moved

cli/formatter/format.go
GRISHNOV: Done

cli/formatter/format.go
GRISHNOV: Done, moved to `encodeScalarNodeToTableRow`

cli/formatter/format.go
GRISHNOV: Done, moved to `encodeScalarNodeToTableRow`

cli/formatter/format_test.go
```diff
@@ -0,0 +1,277 @@
...
+			"+------+-------+----+\n" +
+				"| col1 | 1     | 10 |\n" +
+				"+------+-------+----+\n" +
+				"| col2 | 2     | 20 |\n" +
+				"+------+-------+----+\n" +
+				"| col3 | hello |    |\n" +
+				"+------+-------+----+\n",
+			false,
+		},
+		{
```
oleg-jukovec, line 163: Could you please add samples for TableFormat and TTableFormat with names?

https://github.com/tarantool/tt/blob/a4647e8ac6e4871f8239dbba4734fd8ccff22e3e/test/integration/connect/test_connect.py#L897-L908

GRISHNOV: Added

cli/connect/interactive.go
GRISHNOV: Done

oleg-jukovec: Why do we could not replace the slice `interactiveOptsCases` to a map of [prefx]handleFunction ?

GRISHNOV: The detecting logic of particular option may differ from others:

* Checking of option argument presence
* Validation of option argument
* Detecting can be based in `trimmed` or `in` value from user
* Checking the status of other options

So, I have to move this logic to `detectInteractiveOption` functions

oleg-jukovec (changes requested):
The critical lack of the implementation is the input data for the `MakeOutput` function inconsistent with all formats.
```Go
{ // Ok.
	formatter.DefaultFormat,
	"---\n- 1\n- 2\n- 3\n...",
	"---\n- 1\n- 2\n- 3\n...\n",
	false,
},
{ // Ok.
	formatter.TableFormat,
	"---\n- 1\n- 2\n- 3\n...",
	"+------+\n" +
		"| col1 |\n" +
		"+------+\n" +
		"| 1    |\n" +
		"+------+\n" +
		"| 2    |\n" +
		"+------+\n" +
		"| 3    |\n" +
		"+------+\n",
	false,
},
{ // The problem. It does not print a tuple.
	formatter.DefaultFormat,
	"---\n- DATA: 8\n  COLUMN_1: 30\n  FOO: 1\n  COLUMN_2: 50\n...",
	"---\n- DATA: 8\n  COLUMN_1: 30\n  FOO: 1\n  COLUMN_2: 50\n...\n",
	false,
},
{ // Ok.
	formatter.TableFormat,
	"---\n- DATA: 8\n  COLUMN_1: 30\n  FOO: 1\n  COLUMN_2: 50\n...\n",
	"+----------+----------+------+-----+\n" +
		"| COLUMN_1 | COLUMN_2 | DATA | FOO |\n" +
		"+----------+----------+------+-----+\n" +
		"| 30       | 50       | 8    | 1   |\n" +
		"+----------+----------+------+-----+\n",
	false,
},
```
As a user, I expect to pass a valid data to it, I will get a valid result as a tuples for all file formats. But in practice, not all formats can work with the same input data. As example, `DefaultFormat`/`YamlFormat` could not work with a data with fields names as expected (return tuples without field names in the format).

This will сomplicate development in the future. We will need to build two different YAML documents for export tuples from IPROTO.

So we need to make sure that the same valid input is valid for all formats, and formatted correctly by them.

cli/connect/interactive.go
oleg-jukovec: All detect* calls begins with:

```Go
if opts.trimmed == 'some' {
}
```
or
```Go
if strings.HasPrefix(opts.trimmed, 'some') {
}
```

It could be possible to use the map for it:

```Go
for opt, fun := range map {
    if strings.HasPrefix(trimmed, opt) {
         fun(data_without_trimmed)
    }
}
```

GRISHNOV: Sorry, but not all:
* `detectHelperCase`
* `detectTableGraphicsCase`
* `detectOutputFormatCase`
* `detectTableDialectCase`

And some have some preprocessing logic (described above)

Could we use current implementation? The current solution is extensible via `[...]detectInteractiveOption` and flexible in terms of adding custom logic for processing options via `type detectInteractiveOption`

oleg-jukovec: The current approach certainly works, but it is not extensible in any way (for example, if we need to add autocompletion). It's hard to maintain and easy to make mistakes.

In addition, you are trying to implement different logic for the long and short forms in the same handler. This complicates the logic of potentially simple functions.

```Go
type optHandler struct { // or even an interface
    handler                func(postfix string, opts handlerOpts) error
    // autocompleter  func(postfix string, opts handlerOpts) ([]string, error)
}

commands := map[string]optHandler {
    "\\help": helpHandler,
    "\\":        helpHandler,
    "\\?":      helpHandler,
    setPseudoGraphicsEnable: handlePseudoGraphicsEnable,
    setPseudoGraphicsDisable: handlePseudoGraphicsDisable,
    setFormatPrefix: handleFormatLong,
    setFormatShortPrefix: handleFormatShort,
    setTableDialectPrefix: handleTableDialect,
}
```

It will be much easier to understand and maintain the list of options with such mapping. In addition, you won't need to manually strip the prefix/postfix in every function. Just check it and implement the logic of a handler.

GRISHNOV: Done

oleg-jukovec:
Please, rebase the pull request.

cli/connect/interactive.go
```diff
@@ -0,0 +1,360 @@
...
+		if token != "" {
+			tokens = append(tokens, token)
+		}
+	}
+
+	for i := len(tokens); i > 0; i-- {
+		sub := tokens[:i]
+		key := strings.Join(sub, " ")
+		if handler, ok := interactiveOptionHandlers[key]; ok {
+			handler(handlerOpts)
```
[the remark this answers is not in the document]
GRISHNOV: Done

cli/connect/interactive.go
```diff
@@ -0,0 +1,360 @@
...
+  \table_format <format>          -- tables format markdown, jira or default
+  \help                           -- show this screen
+  \quit                           -- quit interactive console
+  `
+	fmt.Println(help)
+}
+
+// handleLanguageOption handles language interactive option.
+func handleLanguageOption(handlerOpts handlerOpts) {
+	newLang := strings.TrimPrefix(handlerOpts.trimmed, setLanguagePrefix)
```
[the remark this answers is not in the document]
GRISHNOV: Done

cli/connect/interactive.go
```diff
@@ -0,0 +1,360 @@
...
+	}
+}
+
+// handleWidthShortOption handles width interactive short option for table/ttable output formats.
+func handleWidthShortOption(handlerOpts handlerOpts) {
+	if !validateWidthOption(handlerOpts) {
+		return
+	}
+
+	maxWidthStr := strings.TrimPrefix(handlerOpts.trimmed, setTableColWidthShortPrefix)
```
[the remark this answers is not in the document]
GRISHNOV: Done

cli/connect/interactive.go
```diff
@@ -0,0 +1,360 @@
...
+	setMaxWidthOption(maxWidthStr, handlerOpts)
+}
+
+// handleWidthLongOption handles width interactive long option for table/ttable output formats.
+func handleWidthLongOption(handlerOpts handlerOpts) {
+	if !validateWidthOption(handlerOpts) {
+		return
+	}
+
+	maxWidthStr := strings.TrimPrefix(handlerOpts.trimmed, setTableColWidthLongPrefix)
```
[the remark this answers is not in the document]
GRISHNOV: Done

cli/connect/interactive.go
```diff
@@ -0,0 +1,360 @@
...
+			log.Infof("selected output format: %s", handlerOpts.console.outputFormat.String())
+		}
+	} else {
+		log.Warnf("unsupported output format: %s", newFormat)
+	}
+}
+
+// validateFormatOption validates format option and returns false if there is any problems.
+func validateFormatOption(handlerOpts handlerOpts) bool {
+	if handlerOpts.trimmed == strings.TrimSpace(setFormatLongPrefix) {
```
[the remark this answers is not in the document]
GRISHNOV: Done

cli/connect/interactive.go
```diff
@@ -0,0 +1,360 @@
...
+	return true
+}
+
+// handleFormatLongOption handles output format long interactive option.
+func handleFormatLongOption(handlerOpts handlerOpts) {
+	if !validateFormatOption(handlerOpts) {
+		return
+	}
+
+	newFormat := strings.TrimPrefix(handlerOpts.trimmed, setFormatLongPrefix)
```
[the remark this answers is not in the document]
GRISHNOV: Done

cli/formatter/format.go
```diff
@@ -0,0 +1,805 @@
...
+// isTTableFormat returns true if output format is TTableFormat and false if TableFormat.
+func isTTableFormat(outputFormat Format) bool {
+	switch outputFormat {
+	case TTableFormat:
+		return true
+	case TableFormat:
+		return false
+	default:
+		panic("unexpected format")
+	}
```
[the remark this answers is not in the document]
GRISHNOV: Done

cli/formatter/format.go
```diff
@@ -0,0 +1,805 @@
...
+		return arrayRenderNodeType
+	default:
+		return scalarRenderNodeType
+	}
+}
+
+// isRenderNodesTypesEqual checks renderNode equal.
+func isRenderNodesTypesEqual(x renderNode, y renderNode) bool {
+	return getRenderNodeType(x) == getRenderNodeType(y)
+}
```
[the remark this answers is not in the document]
GRISHNOV: Done

cli/formatter/types.go
```diff
@@ -0,0 +1,13 @@
...
+type renderNode interface{}
+
+// nodeType defines a set of supported render node types.
+type nodeType int
+
+const (
+	scalarRenderNodeType = iota
+	arrayRenderNodeType
+	mapRenderNodeType
+)
```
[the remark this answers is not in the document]
GRISHNOV: Done

cli/connect/interactive.go
```diff
@@ -0,0 +1,360 @@
...
+		log.Error("max width option only supports for default table dialect, " +
+			"not for jira or markdown")
+		return false
+	}
+	if handlerOpts.console.outputFormat != formatter.TableFormat &&
+		handlerOpts.console.outputFormat != formatter.TTableFormat {
+		log.Error("max width option only supports for " +
+			"table and ttable output formats")
+		return false
+	}
```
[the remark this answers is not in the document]
GRISHNOV: Will be discussed today

cli/connect/interactive.go
```diff
@@ -0,0 +1,372 @@
...
+	} else {
+		log.Warnf("unsupported table dialect: %s", newTableDialect)
+	}
+}
+
+// handlerOpts contains params for interactive option handler and its detector.
+type handlerOpts struct {
+	in            string
+	trimmedSpace  string
+	trimmedPrefix string
```
oleg-jukovec, line 313: Actually it is not a prefix + It's a little strange to see it here.
```suggestion
	trimmedPostfix string
```

cli/connect/interactive.go
```diff
@@ -0,0 +1,372 @@
...
+		"\\help":                             handleHelpOption,
+		"help":                               handleHelpOption,
+		"\\":                                 handleHelpOption,
+		"\\?":                                handleHelpOption,
+		strings.TrimSpace(setLanguagePrefix): handleLanguageOption,
+		strings.TrimSpace(setTableColWidthLongPrefix):     handleWidthLongOption,
+		strings.TrimSpace(setTableColWidthShortPrefix):    handleWidthShortOption,
+		strings.TrimSpace(setPseudoGraphicsEnablePrefix):  handlePseudoGraphicsEnableOption,
+		strings.TrimSpace(setPseudoGraphicsDisablePrefix): handlePseudoGraphicsDisableOption,
+		strings.TrimSpace(setFormatLongPrefix):            handleFormatLongOption,
```
oleg-jukovec, line 334: There may be even fewer manual checks here:
```suggestion
		strings.TrimSpace(setFormatLongPrefix + " yaml"):            handleFormatYamlOption,
		//...
		strings.TrimSpace(setFormatLongPrefix):            handleFormatInvalidOption,
```

cli/connect/interactive.go
```diff
@@ -0,0 +1,372 @@
...
+}
+
+// handleFormatYamlShortOption handles yaml output format short interactive option.
+func handleFormatYamlShortOption(opts handlerOpts) {
+	if !validateFormatOption(opts) {
+		return
+	}
+
+	newFormat := strings.TrimPrefix(opts.trimmedSpace, setFormatShortPrefix)
+	setFormatOption(newFormat, opts)
```
oleg-jukovec, line 239:
```suggestion
	setFormatOption(YamlFormat.String(), opts)
```

oleg-jukovec (changes requested):
At that moment `console_eval_func_body_tbls_fmt.lua` and `console_eval_func_body.lua` return a different YAML data with a different information for the same request. This data is passed to the input of the `MakeOutput` function. The problem is the lack of symmetry, not all incoming data can be displayed as intended by all formats.

I still think that we need to come to one input format for `MakeOutput` function. In order to achieve the goal, we can return the same YAML from: `console_eval_func_body_tbls_fmt.lua` and `console_eval_func_body.lua`.

It could be:

```YAML
meta:
  # Some help metadata. I expect that it could be empty for `console_eval_func_body.lua`.
  # is_table: false
  # is_tuple: true
  # columns: ["foo", "bar", "car", "col4"]

data:
   # The original data format as it returns from the `console.eval(...)` or
   #`console_eval_func_body.lua` at that moment.
```

So instead of modifying the original tuples in `console_eval_func_body_tbls_fmt.lua`, let's go a slightly different way: we will add the necessary metadata near the original content.

As a result, we can parse this inside the Go code and use it in the same way for all formats in `MakeOutput`:

```Go
output, err := MakeOutput(format, meta, data) // Some formats could use the meta, some do not.
```

I don't mind changing the structure of the document if another is useful. But the idea must be saved: the same data must be displayed in the same way for different formats.

cli/connect/interactive.go
GRISHNOV: Done

cli/connect/interactive.go
GRISHNOV: Renamed

cli/formatter/format.go
```diff
@@ -0,0 +1,794 @@
...
+	t.AppendRows(rows)
+
+	if opts.NoGraphics {
+		t.SetStyle(table.Style{Box: StyleWithoutGraphics})
+	}
+	if opts.ColWidthMax > 0 {
+		handleColWidth(len(commonKeys), t, opts)
+	}
+
+	if opts.TableDialect == MarkdownTableDialect {
```
psergee, line 408: To avoid such conditions, TableDialect must be an interface with concrete implementations for Markdown, Jira, etc..

oleg-jukovec:
I suggest to forget about everything that happened before and starting the review all over again.

cli/formatter/table.go
```diff
@@ -0,0 +1,492 @@
...
+	colWidthTransformer := text.Transformer(func(val interface{}) string {
+		str := fmt.Sprintf("%v", val)
+		widthMax := opts.ColumnWidthMax
+		if utf8.RuneCountInString(str) > widthMax {
+			firstLine := string([]rune(str)[:widthMax])
+			remainingLines := string([]rune(str)[widthMax:])
+			firstLine = firstLine + "+"
+			remainingLines = text.InsertEveryN(remainingLines, '+', widthMax-1)
+			return firstLine + remainingLines
+		}
```
DerekBum, line 210: Why do we need this, if we can just use `return text.InsertEveryN(str, '+', widthMax-1)`?

cli/formatter/table.go
```diff
@@ -0,0 +1,492 @@
...
+			anyMaps = append(anyMaps, castedMap)
+		}
+
+		var mapsBatchs = make([][]map[string]any, len(anyMaps))
+		var batchPointer = 0
+		mapsBatchs[batchPointer] = append(mapsBatchs[batchPointer], anyMaps[0])
+
+		for i := 0; i < len(anyMaps)-1; i++ {
+			if !isMapKeysEqual(anyMaps[i], anyMaps[i+1]) {
+				batchPointer = batchPointer + 1
```
DerekBum, line 378:
```suggestion
				batchPointer++
```

cli/formatter/table.go
```diff
@@ -0,0 +1,492 @@
...
+		var res, batchRes string
+		var err error
+		for _, batch := range mapsBatchs {
+			if len(batch) != 0 {
+				batchRes, err = renderEqualMaps(batch, transpose, opts)
+				if err != nil {
+					return "", err
+				}
+				if !opts.Graphics {
+					batchRes = batchRes + "\n"
```
DerekBum, line 392:
```suggestion
					batchRes += "\n"
```

cli/formatter/table.go
```diff
@@ -0,0 +1,492 @@
...
+				}
+				res = res + batchRes
```
DerekBum, line 394:
```suggestion
				res += batchRes
```

cli/formatter/table.go
```diff
@@ -0,0 +1,492 @@
...
+type metadataRows struct {
+	Metadata []metadataField `yaml:"metadata"`
+	Rows     [][]any         `yaml:"rows"`
+}
+
+// remapMetadataRows creates maps from rows with a metainformation.
+func remapMetadataRows(meta metadataRows) []any {
+	var nodes []any
+	for _, row := range meta.Rows {
+		index := 1
```
DerekBum, line 439: Is there any specific reason for `index` to start from 1 and not 0?

cli/formatter/table.go
```diff
@@ -0,0 +1,492 @@
...
+	if input == "---\n- \n...\n" || input == "---\n-\n...\n" {
+		input = "--- ['']\n...\n"
+	}
+
+	var meta []metadataRows
+	var nodes []any
+
+	// First of all try to read it as tuples with metadata (SQL output format).
+	err := yaml.Unmarshal([]byte(input), &meta)
+	if err == nil && len(meta) > 0 && len(meta[0].Rows) > 0 && len(meta[0].Rows) > 0 {
```
DerekBum, line 466:
```suggestion
	if err == nil && len(meta) > 0 && len(meta[0].Rows) > 0 {
```

cli/formatter/table.go
```diff
@@ -0,0 +1,492 @@
...
+		nodes = append(nodes, []any{""})
+	}
+
+	// The code tries to combine multiple values into a one batch by type.
+	var batches = make([][]any, len(nodes))
+	var batchPointer = 0
+	batches[batchPointer] = append(batches[batchPointer], nodes[0])
+	for i := 0; i < len(nodes)-1; i++ {
+		if !isRenderNodesTypesEqual(nodes[i], nodes[i+1]) {
+			batchPointer = batchPointer + 1
```
DerekBum, line 486:
```suggestion
			batchPointer++
```

test/integration/connect/test_output_format_app/test_app.lua
DerekBum, line 1: What does this test check? There is no assertions or checks, and (appears to me) it only sleeps for 5 seconds endlessly.

cli/formatter/lua.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	var decoded []any
+	if err := yaml.Unmarshal([]byte(input), &decoded); err == nil {
+		var res string
+		for i, unpackedVal := range decoded {
+			if i < len(decoded)-1 {
+				res = res + luaEncodeElement(unpackedVal) + ", "
+			} else {
+				res = res + luaEncodeElement(unpackedVal)
```
DerekBum, line 61:
```suggestion
				res += luaEncodeElement(unpackedVal)
```

cli/formatter/lua.go
```diff
@@ -0,0 +1,68 @@
...
+	if input == "---\n...\n" {
+		return ";\n", nil
+	}
+
+	var decoded []any
+	if err := yaml.Unmarshal([]byte(input), &decoded); err == nil {
+		var res string
+		for i, unpackedVal := range decoded {
+			if i < len(decoded)-1 {
+				res = res + luaEncodeElement(unpackedVal) + ", "
```
DerekBum, line 59:
```suggestion
				res += luaEncodeElement(unpackedVal) + ", "
```

cli/formatter/lua.go
```diff
@@ -0,0 +1,68 @@
...
+	switch t := elem.(type) {
+	case map[any]any:
+		res := "{"
+		first := true
+		for k, v := range t {
+			if !first {
+				res += ", "
+			}
+			if str, ok := k.(string); ok {
+				res = res + fmt.Sprintf("%s = %s", str, luaEncodeElement(v))
```
DerekBum, line 20:
```suggestion
				res  += fmt.Sprintf("%s = %s", str, luaEncodeElement(v))
```

cli/formatter/lua.go
```diff
@@ -0,0 +1,68 @@
...
+			} else {
+				res = res + fmt.Sprintf("[%v] = %s", k, luaEncodeElement(v))
```
DerekBum, line 22:
```suggestion
				res += fmt.Sprintf("[%v] = %s", k, luaEncodeElement(v))
```

cli/formatter/table.go
```diff
@@ -0,0 +1,492 @@
...
+// set of arrays.
+func renderArraysAsTables(batch []any, transpose bool, opts Opts) (string, error) {
+	var result string
+	for _, node := range batch {
+		array := node.([]any)
+		barchRes, err := renderArraysAsTable(array, transpose, opts)
+		if err != nil {
+			return "", err
+		}
+		result = result + barchRes
```
DerekBum, line 183:
```suggestion
		result += barchRes
```

DerekBum (commented):
Thanks for the patch! I've got some questions, please see the comments below.

Also, are changes from `x = x + ...` to `x += ...` and `x++` considered as too sinister nits? I just think that it would be nice to see a uniform code style (and the method of addition differs from line to line).

test/integration/connect/test_connect.py
```diff
@@ -160,11 +190,51 @@ def test_connect_and_get_shortcuts(tt_cmd, tmpdir_with_cfg):
...
+  \\set output <format>            -- set format lua, table, ttable or yaml (default)
+  \\set table_format <format>      -- set table format default, jira or markdown
+  \\set graphics <false/true>      -- disables/enables pseudographics for table modes
+  \\set table_column_width <width> -- set max column width for table/ttable
+  \\xw <width>                     -- set max column width for table/ttable
+  \\x                              -- switches output format cyclically
+  \\x[l,t,T,y]                     -- set output format lua, table, ttable or yaml
+  \\x[g,G]                         -- disables/enables pseudographics for table modes
+  \\shortcuts                      -- show available hotkeys and shortcuts
+  \\quit, \\q                       -- quit from the console
```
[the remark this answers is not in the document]
oleg-jukovec: Everything is ok, there are two escapes `\\` + `\\`, but in other lines there is only one. So this breaks the alignment of the test string (but not the actual output).

cli/connect/const.go
```diff
@@ -0,0 +1,89 @@
+package connect
+
+// setLanguagePrefix is used to set a language in the Tarantool console.
+const setLanguagePrefix = "\\set language"
+
+// setLanguage is a command to set a current language.
+const setLanguage = setLanguagePrefix
```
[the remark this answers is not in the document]
oleg-jukovec: The first is the command to send to `Tarantool` to change the language, the second is the command to change the language to `tt`. In fact it is the same string, but in theory it could be different strings.

cli/formatter/table.go
```diff
@@ -0,0 +1,492 @@
...
+	if array, ok := batch[0].([]any); ok {
+		return isSingleType(array, arrayNodeType)
+	} else {
+		return false
+	}
+}
+
+// renderArrays returns a string representation of arrays.
+func renderArrays(batch []any, transpose bool, opts Opts) (string, error) {
+	if !isSingleArrayOfArrays(batch) {
```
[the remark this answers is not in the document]
oleg-jukovec: In one case the output is just one array table, in the second case the output is number of array tables. We have such test cases:

https://github.com/tarantool/tt/blob/f3d4709c802e5bbdeaad691b8e1e215ad9e7b5aa/test/integration/connect/test_connect.py#L964-L974
https://github.com/tarantool/tt/blob/f3d4709c802e5bbdeaad691b8e1e215ad9e7b5aa/test/integration/connect/test_connect.py#L1027-L1038

cli/formatter/table.go
oleg-jukovec: Yes it is, Lua-tables with "array" parts are marshal-unmarshal with `1` as a key for the first index. So, Lua table `{foo = "bar", [1] = 1, [2] = 2}`  would be after unmarshaling:

`Go
map[any]any{
    "foo": "bar",
    1: "1",
    2: "2",
}
```

The code do the same mapping.

File: test/integration/connect/test_output_format_app/test_app.lua
Comment (oleg-jukovec) [reply test/integration/connect/test_output_format_app/test_app.lua:1]: It is not a test, but a test application with a test data. See other folders with `test/integration/connect/test_output_format_app/`

File: cli/formatter/table.go (hunk 60 above)
Comment (oleg-jukovec) [reply cli/formatter/table.go:210]: It is a bit different output:
```
+-----+-----+-----+
| 7   | Eli | 18  |
|     | +za |     |
|     | +be |     |
|     | +th |     |
+-----+-----+-----+
```
vs
```
+-----+-----+-----+
| 7   | El+ | 18  |
|     | iz+ |     |
|     | ab+ |     |
|     | et+ |     |
|     | h   |     |
+-----+-----+-----+
```

File: cli/connect/commands.go [hunk 75]
```diff
@@ -0,0 +1,598 @@
...
+	_ cmd = argUnsignedCmdDecorator{}
+	_ cmd = argBooleanCmdDecorator{}
+)
+
+var (
+	errNotUnsigned = errors.New("the command expects one unsigned number")
+	errNotBoolean  = errors.New("the command expects one boolean")
+)
+
+// find returns true if the string is founded in the sorted slice.
```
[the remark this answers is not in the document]
oleg-jukovec: Thank you, fixed.

cli/connect/commands.go
```diff
@@ -0,0 +1,598 @@
...
+func (command noArgsCmdDecorator) Run(console *Console,
+	cmd string, args []string) (string, error) {
+	if len(args) != 0 {
+		return "", fmt.Errorf("the command does not expect arguments")
+	}
+	return command.base.Run(console, cmd, args)
+}
+
+// argSetCmdDecorator is a decorator for a command that checks that a
+// first argument in the set of allowed arguments.
```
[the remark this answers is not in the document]
oleg-jukovec: Updated.

cli/connect/commands.go
```diff
@@ -0,0 +1,598 @@
...
+	if len(args) != 1 || !find(command.sorted, args[0]) {
+		return "", fmt.Errorf("the command expects one of: %s",
+			strings.Join(command.sorted, ", "))
+	}
+
+	return command.base.Run(console, cmd, args)
+}
+
+// argUnsignedDecorator is a decorator for a command that checks that a
+// first argument is a positive integer.
```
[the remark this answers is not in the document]
oleg-jukovec: Updated.

cli/connect/commands.go
```diff
@@ -0,0 +1,598 @@
...
+	}
+	if _, err := strconv.ParseUint(args[0], 10, 64); err != nil {
+		return "", errNotUnsigned
+	}
+
+	return command.base.Run(console, cmd, args)
+}
+
+// argBooleanDecorator is a decorator for a command that checks that a
+// first argument is a boolean.
```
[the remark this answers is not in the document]
oleg-jukovec: Updated.

cli/connect/const.go
```diff
@@ -0,0 +1,89 @@
...
+// setFormatTable is a short command to set the ttable format.
+const setFormatTTable = "\\xT"
+
+// setFormatTableShortPrefix is a short command to set the table output
+// format.
+const setFormatTableShortPrefix = "\\xt"
+
+// setFormatTTableShortPrefix is a short command to set the ttable output
+// format.
+const setFormatTTableShortPrefix = "\\xT"
```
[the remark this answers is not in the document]
oleg-jukovec: It is a double copy-paste issue, the second ones removed.

cli/formatter/lua.go
```diff
@@ -0,0 +1,68 @@
...
+				res += ", "
+			}
+		}
+		return res + "}"
+	default:
+		if elem == nil {
+			return "nil"
+		}
+		if str, ok := elem.(string); ok {
+			return fmt.Sprintf("\"%v\"", str)
```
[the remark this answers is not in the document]
oleg-jukovec: Updated.

cli/formatter/style.go
```diff
@@ -0,0 +1,34 @@
...
+	MiddleVertical:   " ",
+	PaddingLeft:      " ",
+	PaddingRight:     " ",
+	PageSeparator:    "\n",
+	Right:            " ",
+	RightSeparator:   " ",
+	TopLeft:          " ",
+	TopRight:         " ",
+	TopSeparator:     " ",
+	UnfinishedRow:    "  ",
```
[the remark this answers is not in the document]
oleg-jukovec: Thank you, fixed.

cli/formatter/table.go
```diff
@@ -0,0 +1,492 @@
...
+			dst[strKey] = v
+		} else {
+			dst[fmt.Sprint(k)] = v
+		}
+	}
+
+	return dst
+}
+
+// deepCastAnyMapToStringMap casts all map[map]any to map[string]any deeply.
```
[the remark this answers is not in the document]
oleg-jukovec: Fixed.

cli/formatter/table.go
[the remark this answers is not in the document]
oleg-jukovec: Reverted to `if (true) else` condition instead of if (false) else`.

cli/formatter/table.go
oleg-jukovec: Fixed.

cli/formatter/table.go
oleg-jukovec: Updated.

cli/formatter/table.go
oleg-jukovec: Fixed.

cli/formatter/table.go
oleg-jukovec: Fixed.

cli/formatter/table.go
oleg-jukovec: Updated

cli/formatter/lua.go
oleg-jukovec: Updated

cli/formatter/lua.go
oleg-jukovec: Updated

cli/formatter/lua.go
oleg-jukovec: Updated

cli/formatter/lua.go
oleg-jukovec: Updated

cli/formatter/table.go
oleg-jukovec: Updated

cli/connect/commands.go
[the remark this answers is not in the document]
oleg-jukovec: Updated

oleg-jukovec (approved):
LGTM =)

cli/formatter/table.go
DerekBum: Is there any test for checking the correctness of handling the `ColumnWidthMax` argument? Maybe we should add one.

DerekBum: Nevermind, here's this test https://github.com/tarantool/tt/blob/f3d4709c802e5bbdeaad691b8e1e215ad9e7b5aa/test/integration/connect/test_connect.py#L1471-L1502

cli/formatter/table.go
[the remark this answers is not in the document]
oleg-jukovec: The code updated. It should be clear now.

DerekBum (approved):

PR #549, opened 2023-07-27
cluster: add cluster configuration file parsing

### LINKED ISSUE #543: cluster: implement a package for reading/writing tarantool cluster configuration
Implement `cluster` package in `cli` with an API for reading and writing local cluster configuration.

oleg-jukovec:
The patch introduces four entry points for a configuration collection:

* `cluster.GetClusterConfig(path)` retrieves the cluster configuration from the specified path. It uses a configuration file from the path, etcd and default environment variables [1] as sources.
* `cluster.Instances(cconfig)` returns a list of instance names from a cluster configuration.
* `cluster.HasInstance(cconfig, name)` checks is an instance with the name presents in the cluster configuration.
* `cluster.GetInstanceConfig(cconfig, instanceName)` retrieves the complete configuration for a specific instance by name from a cluster config. It also always checks the same environment variables used by Tarantool [2][3].

1. https://github.com/tarantool/tarantool/pull/8917
2. https://github.com/tarantool/tarantool/blob/532bada7289b4197cc3bc081081257d050030ece/src/box/lua/config/init.lua#L133
3. https://github.com/tarantool/tarantool/blob/532bada7289b4197cc3bc081081257d050030ece/src/box/lua/config/init.lua#L105

Closes #543


cli/cluster/config.go
```diff
@@ -0,0 +1,161 @@
...
+
+// Set sets a value to a configuration path.
+func (c *Config) Set(path []string, value any) error {
+	if path == nil || len(path) == 0 {
+		c.paths = value
+		return nil
+	}
+
+	last := len(path) - 1
+	if err := c.createMaps(path[0:last]); err != nil {
```
psergee, line 79: Isn't it useful for `createMaps` to return the last created map?

cli/cluster/file.go
```diff
@@ -0,0 +1,32 @@
...
+	"os"
+)
+
+// FileCollector collects data from a YAML file.
+type FileCollector struct {
+	path string
+}
+
+// NewFileCollector create a new file collector for a path.
+func NewFileCollector(path string) Collector {
```
psergee, line 14: Do not return interfaces, return concrete implementation.

cli/cluster/file.go
```diff
@@ -0,0 +1,32 @@
...
+	return FileCollector{
+		path: path,
+	}
+}
+
+// Collect collects a configuration from a file located at a specified path.
+func (c FileCollector) Collect() (*Config, error) {
+	data, err := os.ReadFile(c.path)
+	if err != nil {
+		return nil, fmt.Errorf("unable to read a file '%s': %w", c.path, err)
```
psergee, line 24:
```suggestion
		return nil, fmt.Errorf("unable to read a file %q: %w", c.path, err)
```

cli/cluster/file.go
```diff
@@ -0,0 +1,32 @@
...
+	}
+
+	config, err := NewYamlCollector(data).Collect()
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse a file '%s': %w", c.path, err)
```
psergee, line 29:
```suggestion
		return nil, fmt.Errorf("unable to parse a file %q: %w", c.path, err)
```

cli/cluster/generate.go
```diff
@@ -0,0 +1,3 @@
+package cluster
+
+//go:generate ./env_generate.lua
```
psergee, line 3: Is generation supposed to be invoked manually?

cli/cluster/yaml.go
```diff
@@ -0,0 +1,35 @@
...
+	"gopkg.in/yaml.v2"
+)
+
+// YamlCollector collects a configuration from YAML data.
+type YamlCollector struct {
+	data []byte
+}
+
+// NewYamlCollector create a new YAML collector.
+func NewYamlCollector(data []byte) Collector {
```
psergee, line 15: Return structs, not interfaces.

cli/cluster/yaml.go
```diff
@@ -0,0 +1,35 @@
...
+	return YamlCollector{
+		data: data,
+	}
+}
+
+// Collect collects a configuration from YAML data.
+func (c YamlCollector) Collect() (*Config, error) {
+	var t map[any]any
```
psergee, line 23: Please, use meaningful names for variables.

cli/cluster/yaml_test.go
```diff
@@ -0,0 +1,97 @@
...
+		})
+	}
+}
+
+func TestYamlCollector_unique(t *testing.T) {
+	collector := cluster.NewYamlCollector([]byte("config: asd"))
+	require.NotNil(t, collector)
+
+	config1, err := collector.Collect()
+	require.Nil(t, err)
```
psergee, line 89: Use `NoError` for no error cases.

cli/cluster/config.go
```diff
@@ -0,0 +1,161 @@
+package cluster
+
+import (
+	"fmt"
+
+	"gopkg.in/yaml.v2"
+)
+
+const (
+	fmtPathNotMap   = "path %v is not a map"
```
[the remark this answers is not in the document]
oleg-jukovec: Updated.

cli/cluster/config.go
```diff
@@ -0,0 +1,161 @@
...
+type Config struct {
+	paths any
+}
+
+// NewConfig creates a new empty configuration.
+func NewConfig() *Config {
+	return &Config{}
+}
+
+func (c *Config) createMaps(path []string) error {
```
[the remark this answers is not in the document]
oleg-jukovec: Added.

cli/cluster/config.go
```diff
@@ -0,0 +1,161 @@
...
+	var (
+		prev    map[any]any
+		prevKey string
+		cur     any
+	)
+
+	cur = c.paths
```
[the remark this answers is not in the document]
oleg-jukovec: It is a current value for a key in a map =)

Replaced by `currentValue`.

cli/cluster/config.go
```diff
@@ -0,0 +1,161 @@
...
+)
+
+const (
+	fmtPathNotMap   = "path %v is not a map"
+	fmtPathNotExist = "path %v does not exist"
+)
+
+// Config is a container for deserialized configuration.
+type Config struct {
+	paths any
```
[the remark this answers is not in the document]
oleg-jukovec: > Pls, add a comment about this member.

Added:

```Go
	// paths is a container implementation for the deserialized configuration.
	//
	// At this moment it is fully-compatible with yaml.v2 Marshal()/Unmarshal()
	// functions. So it could be marshaled and unmarshaled directly to YAML.
	// We may change it in the future.
```

cli/cluster/file.go
oleg-jukovec: Returns `FileCollector` at now.

cli/cluster/file.go
oleg-jukovec: Updated.

cli/cluster/yaml.go
oleg-jukovec: Returns `YamlCollector` now.

cli/cluster/yaml.go
oleg-jukovec: Replaced by `parser`.

cli/cluster/yaml_test.go
oleg-jukovec: Updated.

cli/cluster/generate.go
oleg-jukovec: At the moment yes.

Doing it automatically before every tt build doesn't seem like a good idea (because the result depends on a Tarantool version). Maybe add an extra target to the magefile for a manual run?

cli/cluster/config.go
oleg-jukovec: Updated.

oleg-jukovec:
@psergee I've introduced a new method `Config.Elems()` and changed `Config.ForEach()`. Please, see an example of usage in `cli/cluster/example_test.yaml` -> `ExampleConfig_ForEach_instance`.

cli/cluster/generate.go
psergee: Yes, I understand that it is not a good idea to run it on each mage build. I meant... What should I run to update the file? What command in which directory?

oleg-jukovec: 1. Install a latest Tarantool.
2. go generate ./... in the root folder of `tt`.

oleg-jukovec: I've added the instruction into the generated file:

```Go
// To update the file:
// 1. Install a latest Tarantool version.
// 2. Run: go generate ./...
```

cli/cluster/cluster.go
```diff
@@ -0,0 +1,171 @@
...
+		}
+	})
+
+	return iconfig
+}
+
+// GetClusterConfig returns a cluster configuration according to options. The
+// function returns a cluster config as is, without merging of settings
+// from scopes: global, group, replicaset, instance.
+func GetClusterConfig(opts PathOpts) (*Config, error) {
```
psergee, line 99: I think we must omit PathOpts, since etcd options are a part of cluster configuration: empty, or got from a file.

cli/cluster/generate.go
oleg-jukovec: I've added `mage generate` target.

cli/cluster/cluster.go
oleg-jukovec: Replaced by `path string`.

cli/cluster/cluster.go
```diff
@@ -0,0 +1,171 @@
...
+			iconfig.Set(path, value)
+		}
+	})
+
+	return iconfig
+}
+
+// GetClusterConfig returns a cluster configuration according to options. The
+// function returns a cluster config as is, without merging of settings
+// from scopes: global, group, replicaset, instance.
```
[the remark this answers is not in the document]
oleg-jukovec: I've update `GetClusterConfig` and `GetInstanceConfig`. I think this is what you need.

```Go
clusterConfig, _ := cluster.GetClusterConfig("some/path/config.yaml")
// clusterConfig is merged cluster configuration TT_*_DEFAULT + cluster config from the file + cluster config from etcd
instanceConfig := cluster.GetInstanceConfig(clusterConfig, "instanceName")
// instanceConfig is a full-merged configuration for the instance (include TT_*).

File: cli/cluster/config.go [hunk 15]
```diff
@@ -0,0 +1,206 @@
...
+)
+
+// Config is a container for deserialized configuration.
+type Config struct {
+	// paths is a container implementation for the deserialized configuration.
+	//
+	// At this moment it is fully-compatible with yaml.v2 Marshal()/Unmarshal()
+	// functions. So it could be marshaled and unmarshaled directly to YAML.
+	// We may change it in the future.
+	paths any
```
psergee, line 21: Does unmarshaling to go struct type have any drawbacks? Probably `findInstance` might be much simpler if we are working with structs and not `map[any]any`.

oleg-jukovec: The `Сonfig` provides a generic interface independent of any structure. As example, we don't have a common structure for a cluster config and an instance config.

In the future, nothing prevents us from marshaling `Config` into a some structure via helper functions.

cli/cluster/yaml_test.go
```diff
@@ -0,0 +1,97 @@
...
+		nil,
+		[]byte(""),
+	}
+
+	for _, d := range data {
+		t.Run(fmt.Sprintf("%v", data), func(t *testing.T) {
+			collector := cluster.NewYamlCollector(d)
+			config, err := collector.Collect()
+
+			assert.NoError(t, err)
```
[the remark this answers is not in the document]
psergee: Config may be not-nil in case of collect error?

cli/cluster/config.go
psergee: 1. But in fact we have... Global, group, replicaset and instance configurations have the same format with minor exceptions.
2. We need to change all updated paths. If I forget to change some path in my code, I will get an error in the middle of the work, not on YAML parsing at the beginning.

cli/cluster/cluster.go
```diff
@@ -0,0 +1,145 @@
...
+// name from a cluster config. It merges the configuration from all configured
+// sources and scopes: environment, global, group, replicaset, instance.
+func GetInstanceConfig(clusterConfig *Config, instance string) (*Config, error) {
+	if !HasInstance(clusterConfig, instance) {
+		return nil, fmt.Errorf("an instance %q not found", instance)
+	}
+
+	mainEnvCollector := NewEnvCollector(func(path []string) string {
+		middle := strings.ToUpper(strings.Join(path, "_"))
+		return fmt.Sprintf("TT_%s", middle)
```
psergee, line 132: Will this merge `TT_CONFIG_*` vars to my instance config? If yes, this is incorrect. `config` section is only a global level section. Yes, I can just ignore this section for instance level, but if I ever want to dump this instance to YAML, it will be incorrect config.

cli/cluster/yaml_test.go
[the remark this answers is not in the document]
oleg-jukovec: So the test ensures in that.

cli/cluster/cluster.go
oleg-jukovec: The same logic in the implementation of Tarantool's cluster/instance config.

There is some logic in this too: a fully complete configuration of an instance is impossible without `config` section. Yes, this section could only be described at the global scope (like `sharding` only on global + group). But `config` like `sharding` is a part of the full configuration of the instance. See:

https://www.notion.so/tarantool/config-v3-0-0-cb0a2e8f1cf94cdf9e19416c5f77618d

It's not difficult for me to make changes in the implementation, up to you.

But it seems to me it would be more logical to nil `config` and `sharding` sections before export to yaml as an instance scope configuration.

cli/cluster/config.go
oleg-jukovec: You also need to take into account that not only YAML could be a source of configuration (at that moment it is environment variables too, but I think this list may expand in the future (as example, JSON from etcd or something else, there is no Tarantool 3.0 release yet)).

If we implement the config as a concrete type, then this will at least complicate extensibility: nothing prevents a user from adding additional sections to the configuration on different scopes and we will need to catch it somehow + merging becomes not such a trivial task.

I like the idea to keep the `Config` like a layer between configuration sources YAML, but it could be a good idea to add functions to transform the `Config` into some structs. In any case, this may be required further for validation: https://www.notion.so/tarantool/tt-3-0-fbd40e724dce4a04919d1a35c506a65b

```
Требования
- Производит валидацию конфига
```

So it will look like:

environment, YAML etc -> Config -> SomeBaseClusterStruct.
environment, YAML etc -> Config -> SomeBaseInstanceStruct.

cli/cluster/cluster.go
psergee: > config and sharding sections before export to yaml as an instance scope configuration.

And 'config' in replicasets and in groups, and 'sharding' in replicasets. Shouldn't this be controlled by schema, not manual setting to nil?

oleg-jukovec: See:

https://github.com/tarantool/tarantool/blob/671db638b8bb155b3329b11a47a90ac44d47138a/src/box/lua/config/instance_config.lua#L246-L250

psergee: This will be restricted as I know.

cli/cluster/config.go
psergee: I'm not talking about full structuring of the config. I would like to see the global structure:
 config -> group -> replicaset -> instance
which will make the code of the `findInstance` and other config traverse code much easier. I'm not going to play with paths in forEach function. I find it much easier to make a simple loop and `get` call on the level I'm interested in.

oleg-jukovec: > I find it much easier to make a simple loop and `get` call on the level I'm interested in.

Do we really need in practice levels other than `cluster` and `instance`? For example, why do you need the `groups`, `replicasets` levels, what do you want to do with it?

If we do not need other levels, then from the user's point of view it is easier to get a list of instances from a cluster config and iterate over it:

```range
cconfig, _ := GetClusterConfig(path)
for _, instanceName := range GetClusterIntances(cconfig) {
    instances := GetInstanceConfig(cconfig, instanceName)
}
```

psergee: > If we do not need other levels, then from the user's point of view it is easier to get a list of instances from a cluster config and iterate over it:

From user point of view - yes.
But we must maintain the code within. I think it will be much easier to maintain triple `for` loop instead of `forEach` and path manipulations. And if I ever want to write something new, it will be easier, because an editor will show me possible variations after typing '.'. Hope, I will never need to use `forEach` to iterate over a config elements :)

oleg-jukovec: Updated.

psergee (approved):
lgtm

oleg-jukovec:
@psergee configuration structures updated - added the ability to deserialize from YAML.
See changes at `ClusterConfig` at #560 to get it in action.

psergee (approved):

PR #559, opened 2023-08-04
connect: fix nil pointer dereference

askalt:
`console.prompt` can be nill in the case of calling the executor without a prompt, so a check has been added.

Also, the return type of `getExecutor` has been changed to to make it clearer, that the executor can be used not only with a prompt.

### DIFF
--- a/cli/connect/console.go
+++ b/cli/connect/console.go
@@ -159,7 +159,8 @@ func (console *Console) Close() {
 	}
 }
 
-func getExecutor(console *Console) prompt.Executor {
+// getExecutor returns command executor.
+func getExecutor(console *Console) func(string) {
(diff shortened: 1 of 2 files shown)


oleg-jukovec (commented):
Don't we need history in these cases?

askalt:
> Don't we need history in these cases?

No, because if we don't have a prompt, then we no need to push history entries into it.
For example, in this case: https://github.com/tarantool/tt/blob/5fea6e16448598c67afa534e93010f4b1d2f70ae/cli/connect/console.go#L124-L131

oleg-jukovec (approved):
Thank you for the patch!

psergee (approved):

PR #564, opened 2023-08-15
tt: fix version parsing with GC suffix

psergee:
Tarantool Enterprise version includes gc suffix in version string. Version parse fails to parse such versions. Fix it by adding supported suffixes in regexp.


askalt (approved):
Thanks for the patch!

PR #568, opened 2023-08-16
pack: fix package version in cartridge-compact mode

DerekBum:
Version from git was only taken if packing an application from the current directory. But since in `compat` mode we can pack only one application, we can use this application name to get git version from the application directory.

### DIFF
--- a/cli/pack/common.go
+++ b/cli/pack/common.go
@@ -458,10 +458,15 @@ func prepareDefaultPackagePaths(opts *config.CliOpts, packagePath string) {
 func getVersion(packCtx *PackCtx, opts *config.CliOpts, defaultVersion string) string {
 	packageVersion := defaultVersion
 	if packCtx.Version == "" {
-		// Get version from git only if packing an application from the current directory.
-		if opts.App.InstancesEnabled == "." {
-			version, err := util.CheckVersionFromGit(opts.App.InstancesEnabled)
-			if err == nil || version != "" {
+		// Get version from git only if packing an application from the current directory,
+		// or packing with cartridge-compat enabled.
+		var appPath = opts.App.InstancesEnabled
+		if opts.App.InstancesEnabled != "." && packCtx.CartridgeCompat {
+			appPath = filepath.Join(appPath, packCtx.Name)
+		}
+		if opts.App.InstancesEnabled == "." || packCtx.CartridgeCompat {
+			version, err := util.CheckVersionFromGit(appPath)
+			if err == nil {
 				packageVersion = version
 				if packCtx.CartridgeCompat {
 					if normalVersion, err := normalizeGitVersion(packageVersion); err == nil {
--- a/test/integration/pack/test_pack.py
+++ b/test/integration/pack/test_pack.py
@@ -578,6 +578,55 @@ def test_pack_tgz_git_version_compat(tt_cmd, tmpdir):
     assert os.path.isfile(package_file)
 
 
+@pytest.mark.slow
+def test_pack_tgz_git_version_compat_with_instances(tt_cmd, tmpdir):
+    tmpdir = os.path.join(tmpdir, "bundle1")
+    shutil.copytree(os.path.join(os.path.dirname(__file__), "test_bundles", "bundle1"),
+                    tmpdir, symlinks=True, ignore=None,
+                    copy_function=shutil.copy2, ignore_dangling_symlinks=True,
+                    dirs_exist_ok=True)
+
+    base_dir = tmpdir
+    app_dir = os.path.join(base_dir, "app2")
+
+    rc, output = run_command_and_get_output(
+        ["git", "init"],
+        cwd=app_dir, env=dict(os.environ, PWD=app_dir))
+    assert rc == 0
+
+    rc, output = run_command_and_get_output(
+        ["git", "add", "*"],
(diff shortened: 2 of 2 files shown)


cli/pack/common.go
```diff
@@ -458,9 +458,14 @@ func prepareDefaultPackagePaths(opts *config.CliOpts, packagePath string) {
...
-			version, err := util.CheckVersionFromGit(opts.App.InstancesEnabled)
+		// Get version from git only if packing an application from the current directory,
+		// or packing with cartridge-compat enabled.
+		var appPath = opts.App.InstancesEnabled
+		if packCtx.CartridgeCompat {
+			appPath = filepath.Join(appPath, packCtx.Name)
+		}
+		if opts.App.InstancesEnabled == "." || packCtx.CartridgeCompat {
+			version, err := util.CheckVersionFromGit(appPath)
 			if err == nil || version != "" {
```
psergee, line 469: Should't it be && instead of ||?

cli/pack/common.go
```diff
@@ -458,9 +458,14 @@ func prepareDefaultPackagePaths(opts *config.CliOpts, packagePath string) {
...
 	packageVersion := defaultVersion
 	if packCtx.Version == "" {
-		// Get version from git only if packing an application from the current directory.
-		if opts.App.InstancesEnabled == "." {
-			version, err := util.CheckVersionFromGit(opts.App.InstancesEnabled)
+		// Get version from git only if packing an application from the current directory,
+		// or packing with cartridge-compat enabled.
+		var appPath = opts.App.InstancesEnabled
+		if packCtx.CartridgeCompat {
+			appPath = filepath.Join(appPath, packCtx.Name)
```
psergee, line 465: If instances enabled == "." what will be in appPath? ./app_name? How does this work?

test/integration/pack/test_pack.py
```diff
@@ -578,6 +578,54 @@ def test_pack_tgz_git_version_compat(tt_cmd, tmpdir):
     assert os.path.isfile(package_file)
 
 
+@pytest.mark.slow
+def test_pack_tgz_git_version_compat_with_instances(tt_cmd, tmpdir):
+    shutil.copytree(os.path.join(os.path.dirname(__file__), "test_bundles"),
```
psergee, line 583: Can we copy only the required bundle, not all bundles?

test/integration/pack/test_pack.py
```diff
@@ -578,6 +578,54 @@ def test_pack_tgz_git_version_compat(tt_cmd, tmpdir):
...
+        ["git", "tag", "1.2.3"],
+        cwd=app_dir, env=dict(os.environ, PWD=app_dir))
+    assert rc == 0
+
+    rc, output = run_command_and_get_output(
+        [tt_cmd, "pack", "tgz", "--app-list", "app2", "--cartridge-compat"],
+        cwd=base_dir, env=dict(os.environ, PWD=base_dir))
+    assert rc == 0
+
+    package_file = os.path.join(base_dir, "app2-1.2.3.0." + get_arch() + ".tar.gz")
```
psergee, line 625: Have you checked that cartridge-cli generates the same 1.2.3.0 version for the 1.2.3 app tag?

psergee (commented):
Thanks for the patch!
Just minor comments.

cli/pack/common.go
DerekBum: We still need to get version even without `cartridge-compat` flag enabled. In this case we can only get correct version if and only if `InstancesEnabled` is set to `.` (in case we might be packing several applications).

But if `cartridge-compat` is enabled, even if `InstancesEnabled` is not a `.`, we are sure that we are packing only one application (it was checked previously).

cli/pack/common.go
DerekBum: If `cartridge-compat` will be enabled, appPath will be equal to `instances_enabled_dir/app_name`.
Otherwise, it will be equal to `instances_enabled_dir`.

cli/pack/common.go
DerekBum: Seems like it was left aster some rewriting of the `CheckVersionFromGit` function. Now, `err != nil` means `version == ""`. so we can change the condition to just `if err == nil` I think.

test/integration/pack/test_pack.py
DerekBum: Yes, the name will contain `1.2.3.0` version in cartridge too.

test/integration/pack/test_pack.py
DerekBum: Thanks for the comment, yes, we can copy only the required bundle. Fixed.

cli/pack/common.go
DerekBum: > If cartridge-compat will be enabled, appPath will be equal to `instances_enabled_dir/app_name`.

Note that it will be true if `instances_enabled_dir` is not a `.`.

test/integration/pack/test_pack.py
DerekBum: This issue was present in some of the other `pack` tests too. Fixed it in the additional commit.

cli/pack/common.go
psergee: > Note that it will be true if `instances_enabled_dir` is not a `.`. Otherwise, appName will be equal to `.`.
Now - yes :)

psergee (approved):

PR #580, opened 2023-08-28
Fix restart test race

psergee:
Sometimes instance restart test fails on the system with a loaded CPU.
This failure is due to timeout waiting for the instance to restart.
Wait time is 5 second. Watchdog already has 5 seconds timeout before
restarting the instance. so the whole start time becomes 5s watchdog
timeout + instance start time. So test's 5 seconds is not enough.
Increased wait time. Added log message about waiting for restart.
Added check for app status in test.


LeonidVas (approved):
Hi! Thank you for the patch.

PR #592, opened 2023-09-01
cluster: fix merging configs with empty instances

oleg-jukovec:
Before the patch Config.Merge() lost empty maps values.


psergee (approved):
Thanks for the patch.

PR #596, opened 2023-09-07
rocks: replace $PWD with getwd call

psergee:
Relying on PWD var means that tt invocation from  other code require setting of the PWD variable, making chdir is not enogh. Call getwd instead of using getring PWD. getwd call internally also checks PWD environment variable.


0x501D (approved):
LGTM

askalt (approved):
LGTM, thank you, it works!

PR #627, opened 2023-10-03
cli: fix flag parsing issues

### LINKED ISSUE #521: tt rocks --local somehow requires option
When calling `tt rocks --local confg` flag --local somehow digests value `config`.
This is not true, but flag `--local` does not require argument.

```
❯ tt rocks --local  config
   ⨯ Failed to configure Tarantool CLI: failed to change working directory: failed to change directory: chdir /Users/v.grubov/personal/orm/config: no such file or directory
```

Version: `Tarantool CLI version 1.1.0, darwin/amd64. commit: df4e356`

0x501D:
This patch fixes incorrect behavior of parsing global flags of the tt
command.
    
E.g: `tt --cfg some_file install tarantool --cfg other_file`
    
In this case, the result of the second cfg option was taken and the
error was thrown that there was no such option in install.
This behavior is incorrect and has been fixed. The value of the overridden
option will now be ignored.
    
Now global flags are required to be positioned only before child
commands:
    
OK:
`tt --cfg tt.yaml install ...`
    
Not OK:
`tt install ... --cfg tt.yaml`
    
Closes #521

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -111,8 +111,8 @@ func NewCmdRoot() *cobra.Command {
 		Use:   "tt",
 		Short: "Tarantool CLI",
 		Long:  "Utility for managing Tarantool packages and Tarantool-based applications",
-		Example: `$ tt version -L /path/to/local/dir
-  $ tt help -S -I
+		Example: `$ tt -L /path/to/local/dir version
+  $ tt -S -I help
   $ tt completion`,
 		Run: func(cmd *cobra.Command, args []string) {
 			cmd.Help()
@@ -131,6 +131,8 @@ func NewCmdRoot() *cobra.Command {
 		"", "Path to configuration file")
 	rootCmd.Flags().BoolVarP(&cmdCtx.Cli.Verbose, "verbose", "V",
 		false, "Verbose output")
+
+	rootCmd.Flags().SetInterspersed(false)
 
 	rootCmd.AddCommand(
 		NewVersionCmd(),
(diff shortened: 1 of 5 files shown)


test/integration/cli/test_launch.py
```diff
@@ -18,7 +18,7 @@
 # ##### #
 def test_local_launch(tt_cmd, tmpdir):
     module = "version"
-    cmd = [tt_cmd, module, "-L", tmpdir]
+    cmd = [tt_cmd, "-L", tmpdir, module]
```
psergee, line 21: Please check that `--help` option after some sub-command, still prints this sub-command help.

psergee:
Thanks for the patch. Since this is user visible change, please add a note to the change log.

test/integration/cli/test_launch.py
0x501D: Add tests

cli/cmd/root.go
```diff
@@ -118,7 +118,7 @@ func NewCmdRoot() *cobra.Command {
 		Short: "Tarantool CLI",
 		Long:  "Utility for managing Tarantool packages and Tarantool-based applications",
 		Example: `$ tt version -L /path/to/local/dir
```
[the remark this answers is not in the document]
0x501D: Fixed

0x501D:
> Thanks for the patch. Since this is user visible change, please add a note to the change log.

Note to change log added.

psergee (approved):

PR #632, opened 2023-10-06
doc: add examples for tt cluster

oleg-jukovec:
It also fixes help messages for `tt cluster show` and `tt cluster publish`.

### DIFF
--- a/doc/examples.md
+++ b/doc/examples.md
@@ -7,6 +7,7 @@ This file contains various examples of working with tt.
 * [Working with a set of instances](#working-with-a-set-of-instances)
 * [Creating Cartridge application](#creating-cartridge-application)
 * [Working with application templates](#working-with-application-templates)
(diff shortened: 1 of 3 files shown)


doc/examples.md
```diff
@@ -200,6 +201,146 @@ local login = user1
...
+
+-   `tt cluster show SOURCE` - to show a cluster configuration from the
+    `SOURCE`.
+-   `tt cluster publish SOURCE config.yaml` - to publish a cluster
+    configuration to the `SOURCE`.
+
+The `SOURCE` could be:
+
+- An application name or application:instance name.
+- An etcd URI. In the cause you could specify an instance name as an URI
```
psergee, line 220:
```suggestion
- An etcd URI. In this case you could specify an instance name as an URI
```

psergee (approved):
Thanks for the patch! lgtm.

doc/examples.md
oleg-jukovec: Updated.

PR #633, opened 2023-10-06
tt: change working directory

### LINKED ISSUE #570: Change `work dir`
The `work dir` should be moved to directory with the application sources. 
This change should include the following effects:
* by default, artifacts from the application will be created in the source directory
* a directory tree (`var/log`, `var/run`, `var/lib`) will be created in the source directory (now it's created in the root of the env)
* in `tt,yaml` will separate the common section and the application section. In the first case, the relative path will refer to the env root, and in the second case it will refer to the application source directory

psergee:
- Change work dir to application source location. In case of
single instance application, separate directory will be created
in instances enabled directory.
- Re-work application files layout and naming. `app` section 
relative paths are considered relative to working directory, 
which is an application source directory. Application name
sub-directory no longer used for relative paths.
- Support new layout in pack.

Closes #570


cli/pack/common.go
```diff
@@ -63,6 +51,63 @@ var (
...
+	}
+
+	if strings.HasPrefix(src, ".git") ||
+		strings.Contains(src, "/.git") {
+		return true, nil
+	}
+	return false, nil
+}
+
+// prepeareArtifactFilters prepare a slice of filter for runtime artifacts.
```
better0fdead, line 72:
```suggestion
// prepeareArtifactFilters prepares a slice of filter for runtime artifacts.
```

better0fdead (changes requested):
Hi, thx for patch. Please add a Changelog entry and description to the second commit,

LeonidVas (approved):
LGTM!

cli/pack/common.go
psergee: Fixed

psergee:
> Hi, thx for patch. Please add a Changelog entry and description to the second commit,

The change log notes are about the changes, not commits. `pack` command changes are a part of layout and naming update. Since there are a lot of changes for the `pack`, they are in a separate commit.

better0fdead (approved):

Issue #44, opened 2022-04-07
Implement `tt completion` for luarocks


LeonidVas:
Now `tt completion` is implemented for internal modules. The `rocks` module case is where a certain external module is plugged in at compile time. It is necessary to implement `tt completion` for this case as well. The connection mechanism must be generalized, because in the future we will be able to connect more modules in the same way.

psergee:
Closed by #635

PR #635, opened 2023-10-11
completion: added luarocks completions

### LINKED ISSUE #616: tt rocks: fix completion commands
Luarocks version 3.9.2 added built-in commands: `completion` and `admin completion`

Not this commands fails:
```sh
./tt rocks completion bash
   ⨯ <string>:1316: assertion failed!
stack traceback:
	[G]: in function 'assert'
	<string>:1316: in function <<string>:1314>
	<string>:1196: in function 'action'
	<string>:1793: in function 'close'
	<string>:1747: in function 'pass'
	<string>:1919: in function 'pass'
	<string>:2045: in function <<string>:1990>
	(tailcall): ?
	(tailcall): ?
	<string>:573: in function 'run_command'
	<string>:29: in function 'exec'
	<string>:1: in main chunk
	[G]: ?
```
It is necessary to fix these commands or disable them (it won’t work normally using the wrapper.lua wrapper)

0x501D:
added luarocks completions. The `tt rocks completion` and `tt rock admin completion` commands are disabled.

Closes #616 #44

### DIFF
--- a/cli/cmd/completion.go
+++ b/cli/cmd/completion.go
@@ -1,12 +1,14 @@
 package cmd
 
 import (
+	"bytes"
 	"fmt"
-	"os"
 
 	"github.com/spf13/cobra"
 	"github.com/tarantool/tt/cli/cmdcontext"
 	"github.com/tarantool/tt/cli/modules"
+	"github.com/tarantool/tt/cli/rocks"
+	"github.com/tarantool/tt/cli/util"
 )
 
 // NewCompletionCmd creates a new completion command.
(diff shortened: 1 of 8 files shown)


cli/cmd/completion.go
```diff
@@ -46,17 +48,54 @@ func RootShellCompletionCommands(cmd *cobra.Command, args []string,
...
+	idx := bytes.Index(completion, label)
+	if idx == -1 {
+		return nil, fmt.Errorf("failed to inject LuaRocks completions")
+	}
+	res.Write(completion[:idx])
+	res.Write([]byte(injection))
+	res.Write(completion[idx:])
+	res.Write([]byte(rocks))
+
+	return res.Bytes(), nil
```
psergee, line 74: Let's add a test to check that at least `source` command works fine after this injection.

psergee (approved):
Thanks for the patch.

cli/cmd/completion.go
0x501D: Test added.

Issue #575, opened 2023-08-21
Update the log writing scheme


LeonidVas:
Now tarantool writes logs to stdout / stderr on start, and tt adds its logs and writes everything to a file. Logrotate is also done through tt.
After some discussion, it was decided to replace this scheme with the following:
- tarantool logs are the responsibility of the tarantool instance
- tt logs are the responsibility of the tt

After these changes there will be 2 types of files in `var/log`: `tt.log` and `tarantool.log`.

PR #644, opened 2023-10-19
tt: update logging scheme

psergee:
Enable logging to file by default for `tarantool` cluster instances. Default log file name for an instance is `tarantool.log`. `tarantool`'s stdout/stderr and `tt` logs go to `tt.log` file.

Part of #575


LeonidVas (changes requested):
Hi! Thank you for the patch.

After the commit the behavior became different for tarantool 3.0 (logs will be split into two, `tt.log` and `tarantool.log`) and tarantool 2.x / 1.10 (logs will merge into one as before).
After discussing this situation with Pavel, we decided to make the same behavior for all tarantool versions (split the log into two: `tt.log` and `tarantool.log`)

LeonidVas (approved):
LGTM.

PR #662, opened 2023-10-30
running: changes in `instances.yml` parsing logic

### LINKED ISSUE #565: Make processing of instances.yml config consistent
If application directory contains `config.yml` file all items in `instance.yml` config are considered as instance names. All separators are ignored.

If there is no `config.yml` in application directory, `tt` must use `.` char as a separator for application and instance name. 
Suppose, there is a directory `myapp` containing the following `instances.yml`:
```
myapp.master:

myapp.storage:

app1-router:
```
`tt` must manipulate with this application using it's directory name - `myapp`: `tt start/stop/status myapp`. Everything on the left side of the first `.` in instance name is considered the application name.
When a user run `tt start myapp`, `tt` finds application directory `myapp` in `instances.enabled` dir, load project name from a `.rockspec` file if any. If there is no rockspec file, application name is a directory name. During `instances.yml` parsing:
myapp.master and myapp.storage are splitted by `.` and `myapp` is used as application name with `master` and `storage` instances name.
For `app1-router`: there is no app-instance separator, so the instance will be invoked with `TARANTOOL_APP_NAME` variable set to `app1-router`. `TARANTOOL_INSTANCE_NAME` should not be set. When looking for script, `tt` will search for `app1-router.init.lua` first, then `init.lua`. Still need to discuss what to do with instance names for such instances.
This is not an error if application name in `instance.yml` differs from application directory name. Application directory name is used in `tt` cli everywhere. But application name from a `.rockspec` file has a priority over this name when working with environment required by instance script.

askalt:
This patch adds changes to the `instances.yml` parsing logic:
1. Ignore app-instance delimiters for Tarantool 3.0 instances.
2. Don't use dash as an app-instance delimiter. At the same time, `cartridge_app-stateboard`
treated as a special case.

I didn't forget about:
1. [x] tests
2. [x] changelog
3. [x] docs  

Closes #565

### DIFF
--- a/README.md
+++ b/README.md
@@ -455,8 +455,10 @@ instance_name:
   parameter: value
 ```
 
-The dot and dash characters in instance names are reserved for system
-use. if it is necessary for a certain instance to work on a source file
+The dot character in instance names is reserved for system use.
+Also, if an instance name ends with `-stateboard`, it will be recognized
+as `stateboard`.
+If it is necessary for a certain instance to work on a source file
 other than `init.lua`, then you need to create a script with a name in
 the format: `instance_name.init.lua`.
 
--- a/cli/running/running.go
+++ b/cli/running/running.go
@@ -270,8 +270,12 @@ func collectAppDirFiles(appDir string) (appDirCtx appDirCtx, err error) {
 	return
 }
 
-// getInstanceName gets instance name from app name + instance name
-func getInstanceName(fullInstanceName string) string {
+// getInstanceName gets instance name from app name + instance name.
+func getInstanceName(fullInstanceName string, isClusterInstance bool) string {
+	if isClusterInstance {
+		// If we have a cluster instance, delimiters are ignored.
+		return fullInstanceName
+	}
 	sepIndex := strings.IndexAny(fullInstanceName, ".-")
 	if sepIndex == -1 {
 		return fullInstanceName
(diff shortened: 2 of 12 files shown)


cli/running/running.go
```diff
@@ -270,9 +273,18 @@ func collectAppDirFiles(appDir string) (appDirCtx appDirCtx, err error) {
...
+	if isClusterInstance {
+		// If we have a cluster instance, delimiters are ignored.
+		return fullInstanceName
+	}
+	// Consider `-stateboard` suffix for the cartridge application compatibility.
+	if strings.HasSuffix(fullInstanceName, fmt.Sprintf("-%s", stateBoardInstName)) {
+		return stateBoardInstName
+	}
+
+	sepIndex := strings.IndexAny(fullInstanceName, ".")
```
psergee, line 287:
```suggestion
	sepIndex := strings.Index(fullInstanceName, ".")
```

cli/running/running_test.go
```diff
@@ -331,3 +331,20 @@ func Test_collectInstancesSingleInstanceTntCtlLayout(t *testing.T) {
...
+func Test_getInstanceName(t *testing.T) {
+	for _, tc := range []struct {
+		fullInstanceName  string
+		isClusterInstance bool
+		expected          string
+	}{
+		{"master", false, "master"},
+		{"app.master", false, "master"},
+		{"app-stateboard", false, "stateboard"},
+		{"app-master", true, "app-master"},
```
psergee, line 344: Add a test with the same name, but `false` cluster flag. Add `app.inst-001` test.

README.md
```diff
@@ -455,8 +455,9 @@ instance_name:
   parameter: value
 ```
 
-The dot and dash characters in instance names are reserved for system
-use. if it is necessary for a certain instance to work on a source file
+The dot character in instance names is reserved for system
+use. Also, if an instance name ends with `-stateboard`, it will be recognized as `stateboard`.
```
psergee, line 459: This line length differs from other lines in this paragraph.

psergee (commented):
Thanks for the patch!

psergee (approved):

PR #666, opened 2023-11-02
logrotate: update log rotation scheme

### LINKED ISSUE #575: Update the log writing scheme

psergee:
- remove log rotation configuration
- Log rotation functionality is removed from `tt`. Now `logrotate` command reopens the log file and sends SIGHUP to the child tarantool process. Motivation: tt should not take over the work of standard log rotation tools for non-tt files like tarantool logs.

Closes #575


LeonidVas (approved):
LGTM

better0fdead (approved):

PR #669, opened 2023-11-07
test: enable install tests back

psergee:
Some tests were disabled due to incompatibility changes. Enable them back and make them more robust.


LeonidVas (approved):
Hi! Thank you for the patch. If everything works fine, I'm OK.

PR #674, opened 2023-11-07
test: increase timeout delta

psergee:
Increase timeout delta to fix flaky test.


oleg-jukovec (approved):
Thank you for the patch!

PR #675, opened 2023-11-08
CI: add tarantool 3.0 CI

psergee:


cli/rocks/rocks.go
```diff
@@ -27,6 +27,7 @@ const (
 	rocksRepoManifestName     = "manifest"
 	repoRocksPathEnvVarName   = "TT_CLI_REPO_ROCKS"
 	tarantoolPrefixEnvVarName = "TT_CLI_TARANTOOL_PREFIX"
+	tarantoolDefaultPrefixDir = "/usr"
```
LeonidVas, line 30: Changes related to the `rocks` module are not related to the addition of tarantool 3.0 to CI and can be moved to a separate preparatory commit.

.github/actions/prepare-ce-test-env/action.yml
```diff
@@ -48,11 +48,23 @@ runs:
...
+        !startsWith(inputs.tarantool-version, '3.')
       uses: tarantool/setup-tarantool@v1
-      if: ${{ inputs.skip-tarantool-install == 'false' }}
       with:
         tarantool-version: '${{ inputs.tarantool-version }}'
 
+    - name: Install Tarantool from repo
+      if: |
+        inputs.skip-tarantool-install == 'false' &&
+        startsWith(inputs.tarantool-version, '3.')
```
LeonidVas, line 61: I have two questions/suggestions:
- Do I understand correctly that in case of 3.0 the latest 3.0 release (alpha, beta or ...) will be installed?
- I suggest adding a comment that `curl ...` is a temporary trick until we can't use the `setup-tarantool` action, and it will be replaced in the future

LeonidVas (changes requested):
HI! Thank you for the patch. Generally OK, see several comments bellow.
I also suggest changing the title to : "ci: add tarantool 3.0".

.github/actions/prepare-ce-test-env/action.yml
psergee: Answered in the comment in the action code.

cli/rocks/rocks.go
psergee: Done.

LeonidVas (approved):
LGTM.

PR #680, opened 2023-11-11
connect: fix TLS flags setup

oleg-jukovec:
The patch removes an invalid default value for TLS flags. At now the default value `connect.DefaultLanguage.String()` does not cause any problem because it is "".


psergee (approved):
Thanks for the patch! lgtm.

PR #688, opened 2023-11-15
pack: scaffolding for integrity check file generation

AlgebraicWolf:
This patch provides a noop `integrity` module intended to be used as a scaffolding for generating integrity check files. The logic of `tt pack` command is modified to perform calls to appropriate `integrity` methods wherever it is required. It is designed as a separate module to allow easy replacement via directive in `go.mod` file.


cli/cmd/pack.go
```diff
@@ -88,6 +89,9 @@ The supported types are: tgz, deb, rpm`,
 		packCtx.TarantoolVersion,
 		"Version of the tarantool for pack in docker (only with --use-docker flag).")
 
+	// Integrity
```
oleg-jukovec, line 92: Integrity what?
```suggestion
	// Integrity.
```

AlgebraicWolf: Fixed!

cli/integrity/integrity.go
```diff
@@ -0,0 +1,19 @@
...
+// IntegritySigner is a noop integrity check preparation.
+type IntegritySigner struct{}
+
+func NewIntegritySigner(string) (*IntegritySigner, error) {
+	return &IntegritySigner{}, nil
+}
+
+func RegisterIntegrityFlag(cmd *cobra.Command, dst *string) {}
+
+func (IntegritySigner) PackHashing([]util.AppListEntry, string) error {
```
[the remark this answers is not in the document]
AlgebraicWolf: Fixed

cli/integrity/integrity.go
```diff
@@ -0,0 +1,25 @@
+package integrity
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/tarantool/tt/cli/util"
+)
+
+// IntegritySigner is a noop integrity check preparation.
+type IntegritySigner struct{}
```
oleg-jukovec, line 9: It's suitable to make an interface instead of the concrete implementation.

oleg-jukovec (commented):
There are no unit-tests for public API of the module.

cli/integrity/integrity.go
AlgebraicWolf: I had made an interface at some point, but it feels redundant to me. Having it as a part of `integrity` module is inconvenient, since the whole module gets replaced via `go.mod` file, we will have to duplicate it both here and in actual implementation. Moving it to a separate package, like `integrity_interface` also feels unnecessary -- it seems that there is only one implementation of integrity module per project. Even if we resort to having various hashing and signature algorithms, I would rather prefer to parametrize over them rather than provide several implementations of the module.

oleg-jukovec: Just an interface in the `integrity` module will be enough. This will help make the logic more agile in the future without additional steps:

Create different specific implementations from the `NewIntegritySigner` method depending on something in the tt-ee part.

oleg-jukovec, line 9:
```suggestion
// Signer is a noop integrity check preparation.
type Signer struct{}
```

cli/integrity/integrity.go
```diff
@@ -0,0 +1,25 @@
...
+
+// NewIntegritySigner is a noop constructor of an IntegritySigner based
+// on a private key file.
+func NewIntegritySigner(string) (*IntegritySigner, error) {
+	return &IntegritySigner{}, nil
+}
```
oleg-jukovec, line 15:
```suggestion
func NewSigner(string) (*IntegritySigner, error) {
	return &IntegritySigner{}, nil
}
```

cli/integrity/integrity.go
```diff
@@ -0,0 +1,25 @@
...
+
+// RegisterIntegrityFlag is a noop function that is intended to add
+// flags to `tt pack` command.
+func RegisterIntegrityFlag(cmd *cobra.Command, dst *string) {}
```
oleg-jukovec, line 19:
```suggestion
func RegisterWithIntegrityFlag(cmd *cobra.Command, dst *string) {}
```

cli/integrity/integrity.go
```diff
@@ -0,0 +1,25 @@
...
+
+// PackHashing is a noop method that is intended to prepare hash files
+// for applications and environment and sign them.
+func (IntegritySigner) PackHashing([]util.AppListEntry, string) error {
+	return nil
+}
```
oleg-jukovec, line 25: Pack or Sign?
```suggestion
func (Packer) Pack(paht string, []util.AppListEntry) error {
	return nil
}
```
```suggestion
func (Signer) Sign(path string, apps []util.AppListEntry) error {
	return nil
}
```
What is the second argument actually?

oleg-jukovec (commented):
We already have `integrity` in the module name, so the naming could be shorter.

cli/integrity/integrity.go
AlgebraicWolf: I thought it this over, and I feel like neither fit the purpose nicely. I guess something like `AddIntegrityHashes` fits the function's purpose better

oleg-jukovec: I suggest you think more about universal naming.

cli/integrity/integrity.go
AlgebraicWolf: Fixed

cli/integrity/integrity.go
```diff
@@ -0,0 +1,31 @@
...
+
+// RegisterWithIntegrityFlag is a noop function that is intended to add
+// flags to `tt pack` command.
+func RegisterWithIntegrityFlag(cmd *cobra.Command, dst *string) {}
+
+// Sign is a noop preparation of integrity check files.
+func (*DummySigner) Sign([]util.AppListEntry, string) error {
+	return nil
+}
```
oleg-jukovec, line 31: I don't see a reason to make it public.
```suggestion
// dummySigner is a noop integrity check preparator.
type dummySigner struct{}

// Sign is a noop preparation of integrity check files.
func (dummySigner) Sign([]util.AppListEntry, string) error {
	return nil
}

// NewSigner contructrs a noop Signer.
func NewSigner(string) (Signer, error) {
	return &dummySigner{}, nil
}

// RegisterWithIntegrityFlag is a noop function that is intended to add
// flags to `tt pack` command.
func RegisterWithIntegrityFlag(cmd *cobra.Command, dst *string) {}
```

cli/integrity/integrity.go
AlgebraicWolf: I've created an appropriate interface.

AlgebraicWolf: Done

cli/integrity/integrity.go
AlgebraicWolf: Done

cli/integrity/integrity.go
AlgebraicWolf: I gave it a thought, and I'm sold on `Sign`.

cli/integrity/integrity.go
AlgebraicWolf: Done

cli/integrity/integrity.go
```diff
@@ -0,0 +1,31 @@
...
+	// AddIntegrityHashes generates files with hashes for integrity
+	// checks and signs them.
+	Sign(appList []util.AppListEntry, baseDir string) error
+}
+
+// dummySigner is a noop integrity check preparator.
+type dummySigner struct{}
+
+// Sign is a noop preparation of integrity check files.
+func (dummySigner) Sign([]util.AppListEntry, string) error {
```
oleg-jukovec, line 20:
```suggestion
// Sign is a noop preparation of integrity check files.
func (dummySigner) Sign(basePath string, apps []util.AppListEntry) error {
```
where -> applist seems less confusing than applist -> where.
Please, also name arguments in the interface.

cli/integrity/integrity.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"github.com/spf13/cobra"
+	"github.com/tarantool/tt/cli/util"
+)
+
+// Signer implements high-level API for operations that require
+// a private key.
+type Signer interface {
+	// AddIntegrityHashes generates files with hashes for integrity
+	// checks and signs them.
```
oleg-jukovec, line 12: The comment is outdated.

oleg-jukovec (commented):
All that's left is tests.

cli/integrity/integrity.go
```diff
@@ -0,0 +1,31 @@
...
+// dummySigner is a noop integrity check preparator.
+type dummySigner struct{}
+
+// Sign is a noop preparation of integrity check files.
+func (dummySigner) Sign([]util.AppListEntry, string) error {
+	return nil
+}
+
+// NewSigner contructrs a noop Signer.
+func NewSigner(string) (Signer, error) {
```
oleg-jukovec, line 25:
```suggestion
// NewSigner contructrs a noop Signer.
func NewSigner(path string) (Signer, error) {
```

cli/integrity/integrity.go
```diff
@@ -0,0 +1,31 @@
...
+
+// dummySigner is a noop integrity check preparator.
+type dummySigner struct{}
+
+// Sign is a noop preparation of integrity check files.
+func (dummySigner) Sign(basePath string, apps []util.AppListEntry) error {
+	return nil
+}
+
+// NewSigner contructrs a noop Signer.
```
psergee, line 24:
```suggestion
// NewSigner constructs a noop Signer.
```

AlgebraicWolf: Fixed

cli/pack/common.go
```diff
@@ -257,6 +268,16 @@ func prepareBundle(cmdCtx *cmdcontext.CmdCtx, packCtx *PackCtx,
 	if err != nil {
 		return "", err
 	}
+
+	if !packCtx.CartridgeCompat {
```
psergee, line 272: Why?

psergee (commented):
Thank you for the patch!

cli/pack/common.go
AlgebraicWolf: It looks like that would imply that `cartridge-cli` would then be obligated to be able to perform integrity checks as well, and it is not going to.

I guess that throwing an error when both are specified would be a nicer behaviour.

psergee: So let's return an error in this case. Looks like, now we silently skip signing.

AlgebraicWolf: Now it throws an error. I've also added an integration test on tt-ee part

cli/integrity/integrity.go
AlgebraicWolf: Fixed!

cli/integrity/integrity.go
AlgebraicWolf: Fixed

cli/integrity/integrity.go
AlgebraicWolf: Fixed

cli/integrity/integrity_test.go
```diff
@@ -0,0 +1,97 @@
+package integrity
```
oleg-jukovec, line 1: Please, test only the public API.
```suggestion
package integrity_test
```

cli/integrity/integrity_test.go
```diff
@@ -0,0 +1,97 @@
...
+		{
+			name:           "Arbitrary path",
+			privateKeyPath: "private.pem",
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			_, err := NewSigner(testCase.privateKeyPath)
+			require.Equal(t, err, nil, "did not expect an error")
```
oleg-jukovec, line 28:
```suggestion
			signer, err := NewSigner(testCase.privateKeyPath)
			require.Equal(t, err, nil, "did not expect an error")
			require.NotNil(t, signer)
```

cli/integrity/integrity_test.go
```diff
@@ -0,0 +1,97 @@
...
+			name:    "Nothing empty",
+			command: &cobra.Command{},
+			dst:     &someStr,
+		},
+	}
+
+	for _, testCase := range testCases {
+		// This simply mustn't die.
+		t.Run(testCase.name, func(t *testing.T) {
+			RegisterWithIntegrityFlag(testCase.command, testCase.dst)
```
oleg-jukovec, line 66: So, what actually do you test here?

AlgebraicWolf: The only way the function with such signature can die is by dereferencing the null pointer. So, the idea is to throw various combinations of null pointers at it and see if any would trigger a runtime error

cli/integrity/integrity_test.go
AlgebraicWolf: Fixed

cli/integrity/integrity_test.go
AlgebraicWolf: Fixed

cli/integrity/integrity_test.go
oleg-jukovec: > The only way the function with such signature can die is by dereferencing the null pointer.

It also could add a flag into a command.

cli/integrity/integrity_test.go
```diff
@@ -0,0 +1,103 @@
...
+			command: &cobra.Command{},
+			dst:     &someStr,
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			integrity.RegisterWithIntegrityFlag(testCase.command, testCase.dst)
+			if testCase.command != nil {
+				require.Equal(t, reflect.DeepEqual(*testCase.command, cobra.Command{}), true,
```
oleg-jukovec, line 70: It still does not clear what exactly the test tests.
```suggestion
				require.False(t, testCase.command.Flags().HasFlags())
```

cli/integrity/integrity.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"github.com/spf13/cobra"
+)
+
+// Signer implements high-level API for operations that require
+// a private key.
+type Signer interface {
+	// Sign generates files with hashes for integrity
+	// checks and signs them.
+	Sign(basePath string, appNameList []string) error
```
oleg-jukovec, line 12: Up to you, but it is shorter with the same meaning.
```suggestion
	Sign(basePath string, appNames []string) error
```

cli/integrity/integrity_test.go
AlgebraicWolf: Why not? The intention is to check whether the the call to `RegisterWithIntegrityFlag` mutates the command.

cli/integrity/integrity.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"github.com/spf13/cobra"
+)
+
+// Signer implements high-level API for operations that require
+// a private key.
+type Signer interface {
+	// Sign generates files with hashes for integrity
+	// checks and signs them.
```
oleg-jukovec, line 11:
```suggestion
// Signer implements high-level API for package signing.
type Signer interface {
	// Sign generates data to sign a package.
```

cli/pack/common.go
```diff
@@ -113,6 +115,16 @@ func skipArtifacts(cliOpts *config.CliOpts) func(src string) (bool, error) {
...
+	var signer integrity.Signer = nil
+
+	// If integrity checks are enabled, create an IntegritySigner.
+	if packCtx.IntegrityPrivateKey != "" {
+		signer, err = integrity.NewSigner(packCtx.IntegrityPrivateKey)
+
+		if err != nil {
+			return "", err
+		}
+	}
```
oleg-jukovec, line 127: Otherwise we need to create some kind of a dummy implementation any way.
```suggestion
	signer, err := NewSigner(packCtx.IntegrityPrivateKey)}
	if err != nil {
		return "", err	
	}
```

cli/pack/common.go
```diff
@@ -257,6 +269,23 @@ func prepareBundle(cmdCtx *cmdcontext.CmdCtx, packCtx *PackCtx,
...
+	if packCtx.IntegrityPrivateKey != "" {
+		if !packCtx.CartridgeCompat {
+			err = signer.Sign(basePath, appNameList)
+			if err != nil {
+				return "", err
+			}
+		} else {
+			return "", errors.New("cannot pack with integrity checks in cartridge-compat mode")
+		}
+	}
```
oleg-jukovec, line 287: And use it without any condition.
```suggestion
	if packCtx.IntegrityPrivateKey != "" && packCtx.CartridgeCompat {
		return "", errors.New("cannot pack with integrity checks in cartridge-compat mode")
	}

	err = signer.Sign(basePath, appNameList)
	if err != nil {
		return "", err
	}
```
Up to you. But from the code reader side it's better to choose.

cli/integrity/integrity_test.go
oleg-jukovec: The idea of RegistryWithIntegrityFlag is modify a FlagSet. The argument type is chosen incorrectly if you need to test too much.

AlgebraicWolf: > The idea of RegistryWithIntegrityFlag is modify a FlagSet/add a flag into a FlagSet. So the argument type is chosen incorrectly if you need to test too much.

That's fair. I'll refine the argument domain

cli/pack/common.go
AlgebraicWolf: I don't like the idea of creating and making calls to `signer` regardless of whether the key is specified because, even though it works nicely with dummy implementation, it would require us to handle cases where key is not specified on the `tt-ee` part, where there is a non-trivial signer. But if you are comfortable with this, I add the suggested change.

oleg-jukovec: As an alternative: you could remove the dummy implementation and return an error from NewSingner() bacause it should not be called in the CE part.

AlgebraicWolf: > you could remove the dummy implementation and return an error from NewSingner() 

Yep, that seems to me like a better thing to do. Already in progress

AlgebraicWolf: We will have to ditch the `Sign` tests then, as we won't be able to obtain a signer to call it with

cli/integrity/integrity_test.go
AlgebraicWolf: Fixed

cli/integrity/integrity.go
AlgebraicWolf: I've changed it

cli/integrity/integrity.go
AlgebraicWolf: Fixed

cli/pack/common.go
AlgebraicWolf: Now the error is thrown. The good thing is we can now ditch `dummySigner` altogether!

cli/integrity/integrity_test.go
```diff
@@ -0,0 +1,62 @@
...
+			name:           "Arbitrary path",
+			privateKeyPath: "private.pem",
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			signer, err := integrity.NewSigner(testCase.privateKeyPath)
+			require.Nil(t, signer, "signer must not be created")
+			require.Error(t, err, "an error should be produced")
```
oleg-jukovec, line 30: Let's compare the error message too.

cli/integrity/integrity_test.go
```diff
@@ -0,0 +1,62 @@
...
+		{
+			name:    "Nothing empty",
+			command: &cobra.Command{},
+			dst:     &someStr,
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			integrity.RegisterWithIntegrityFlag(testCase.command.Flags(), testCase.dst)
```
oleg-jukovec, line 57: It seems to me that there is an extra command object here.

oleg-jukovec (approved):
Thank you for the patch! LGTM after resolve all comments.

cli/integrity/integrity_test.go
AlgebraicWolf: Fixed

cli/integrity/integrity_test.go
AlgebraicWolf: Fixed

oleg-jukovec: nit: Errorf asserts that a function returned an error (i.e. not `nil`).

cli/integrity/integrity_test.go
oleg-jukovec: I suppose a FlagSet could be created without command + we could test for a `nil` value.

AlgebraicWolf: Oh, I see what you've meant now. I've updated the tests.

psergee (approved):
Lgtm. Just one small comment.

cli/integrity/integrity_test.go
oleg-jukovec: `Errorf`->`EqualError`.

AlgebraicWolf: Right. Fixed

cli/pack/common.go
```diff
@@ -257,6 +269,23 @@ func prepareBundle(cmdCtx *cmdcontext.CmdCtx, packCtx *PackCtx,
...
+
+	if packCtx.IntegrityPrivateKey != "" {
+		if !packCtx.CartridgeCompat {
+			err = signer.Sign(basePath, appNames)
+			if err != nil {
+				return "", err
+			}
+		} else {
+			return "", errors.New("cannot pack with integrity checks in cartridge-compat mode")
```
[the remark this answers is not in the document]
AlgebraicWolf: That makes sense. I've moved the check to `FillCtx`

PR #693, opened 2023-11-20
test: run tests on tarantool 3.0.0~alpha3

psergee:
Run tests on tarantool 3.0 alpha3 to avoid a failure due to tarantool/tarantool#9275

### DIFF
--- a/.github/actions/prepare-ce-test-env/action.yml
+++ b/.github/actions/prepare-ce-test-env/action.yml
@@ -58,6 +58,9 @@ runs:
     # Here is a workaround for Tarantool 3.0 installation until it is supported
     # by setup-tarantool. This step will install the latest pre-released
     # Tarantool 3.0 from repository.
+    #
+    # Now it installs alpha3 version, because the latest versions fail some tests due to
+    # the issue: https://github.com/tarantool/tarantool/issues/9275
     - name: Install Tarantool from repo
       if: |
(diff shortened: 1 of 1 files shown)


.github/actions/prepare-ce-test-env/action.yml
```diff
@@ -56,16 +56,16 @@ runs:
         tarantool-version: '${{ inputs.tarantool-version }}'
 
     # Here is a workaround for Tarantool 3.0 installation until it is supported
-    # by setup-tarantool. This step will install the latest pre-released
-    # Tarantool 3.0 from repository.
+    # by setup-tarantool. It installs alpha3 version, because the latest
+    # versions fail some tests due to the issue: https://github.com/tarantool/tarantool/issues/9275
```
oleg-jukovec, line 60: It seems to me that this will make it clearer what should be, but why not. Up to you.
```suggestion
    # by setup-tarantool. This step will install the latest pre-released
    # Tarantool 3.0 from repository.
    #
    # Now it installs alpha3 version, because the latest versions fail some tests due to
    # the issue: https://github.com/tarantool/tarantool/issues/9275
```

oleg-jukovec (approved):

.github/actions/prepare-ce-test-env/action.yml
psergee: Done.

PR #694, opened 2023-11-23
connect: change completion behavior

askalt:
This patch changes the behavior of `tt connect` auto-completion, when there are no running apps now. 
In this case `tt connect` completion should show directories and files.


psergee (approved):
Thank you for the patch!

PR #699, opened 2023-11-29
start: separate pgid for watchdog process

psergee:
Set separate process group for watchdog process, so it will not be killed by a closed ssh session.


LeonidVas (approved):
Hi! Thank you for the patch.
I also think it would be nice to add a comment to the code. Up to you.
LGTM.

PR #704, opened 2023-12-05
cluster: add `key` argument to URI

oleg-jukovec:
@TarantoolBot document
Title: new argument for `tt cluster show` and `tt cluster publish`

The patch adds a new argument `key` that could be specified via etcd URI:

* key - a target configuration key in the prefix.

A user can now perform operations (`show` and `publish`) on individual keys in etcd, not just for entire prefixes.

As example:

```
$ tt cluster show https://user:pass@localhost:2379/prefix?key=target
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ running apps.
 ### Added
 
 - `tt env`: add current environment binaries location to the PATH variable.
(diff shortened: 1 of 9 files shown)


askalt (approved):
Thank you for the patch!

LeonidVas (approved):
HI! Thank you for the patchset. LGTM

PR #733, opened 2023-12-28
replicaset: extract discover methods

oleg-jukovec:
The patch extracts two methods into the `replicaset` subpackage:

* `DiscoveryApplication(app RunningCtx, orchestrator Orchestrator)` to discovery replicasets information for an application running context.
* `DiscoveryInstance(evaler Evaler, orchestrator Orchestrator)` to discovery replicasets information for a connection.

An orchestrator type can be forced manually using the second argument or you can use OrchestratorUnknown for automatic mode.


psergee (approved):
Thanks for the patch! lgtm.

PR #736, opened 2023-12-29
create: add built-in vshard cluster template

askalt:
This patch introduces built-in `vshard_cluster` template,
which uses tarantool 3.0 config.


psergee (approved):
Thanks for the patch! lgtm.
Will auto-completion suggest this new template?

askalt:
> Thanks for the patch! lgtm. Will auto-completion suggest this new template?

Yes.

PR #743, opened 2024-01-11
build(deps): bump go-tarantool to v1.12.2

oleg-jukovec:
The bump needed to fix svacer issues.


psergee (approved):
Thanks for the patch!

PR #752, opened 2024-01-18
ci: set tests timeout of self-hosted runners

### LINKED ISSUE #713: ci: add a timeout for pytest tests
Now tests on CI could freeze for a long time. It takes away tarantool's self-hosted macOS runners from a work for other projects.
We need to set not too big timeout for `pytest` at least for macOS jobs.
`go test` has 10 min timeout by default, so the Golang unit-tests should not be a problem

psergee:
Sometimes tests may hang and block other jobs.
Set not too big timeout for the jobs to avoid blocking forever.

Closes #713


oleg-jukovec (approved):
How about a separate timeout for `pytest` test? To avoid freeze on the local machine, as example.

psergee:
> How about a separate timeout for `pytest` test? To avoid freeze on the local machine, as example.

Added.

PR #754, opened 2024-01-19
replicaset: prefer cartridge instances without critical issues during discovery

### LINKED ISSUE #747: replicaset: `status myapp` for cartridge should no discovery expelled instances
At now `tt replicaset status myapp` could retrieve a topology information from expelled instances. This results in out of date topology information being shown to the user. We need to fix the problem.

askalt:
Now, discovery prefers instances without critical issues during search.

Currently, there is no way to determine whether an instance has been expelled from the replicaset. However, critical issues are likely to be the reason for expelling an instance. Also, expelling an instance leads to the critical issues with replication in some cases. Therefore, we can identify expelled instances based on indirect reasons.

Closes https://github.com/tarantool/tt/issues/747

### DIFF
--- a/cli/replicaset/cartridge.go
+++ b/cli/replicaset/cartridge.go
@@ -25,6 +25,10 @@ type cartridgeTopology struct {
 	Provider string
 	// Replicasets is an array of replicasets.
 	Replicasets []Replicaset
+	// IsCritical indicates whether instance has critical issues.
+	IsCritical bool `mapstructure:"is_critical"`
+	// IsBootstrapped indicates whether instance is bootstrapped.
+	IsBootstrapped bool
 }
 
 // CartridgeInstance is an instance with the Cartridge orchestrator.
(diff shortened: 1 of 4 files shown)


cli/replicaset/cartridge.go
```diff
@@ -107,11 +129,15 @@ func (c *CartridgeApplication) GetReplicasets() (Replicasets, error) {
...
-				return true, err
+		err = EvalForeachAlive(c.runningCtx.Instances, InstanceEvalFunc(
+			func(inst running.InstanceCtx, evaler connector.Evaler) (bool, error) {
+				receviedReplicasets, err := NewCartridgeInstance(evaler).GetReplicasets()
+				if err != nil {
+					return true, err
+				}
+				var toStop bool
+				replicasets, toStop = updateCartridgeReplicasets(replicasets, receviedReplicasets)
+				return toStop, nil
```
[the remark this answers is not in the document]
askalt: The idea is applied.

cli/replicaset/cartridge.go
```diff
@@ -25,6 +25,10 @@ type cartridgeTopology struct {
 	Provider string
 	// Replicasets is an array of replicasets.
 	Replicasets []Replicaset
+	// IsCritical indicates whether instance has critical issues.
+	IsCritical bool `mapstructure:"is_critical"`
+	// isBootstrapped indicates whether instance is bootstrapped.
+	isBootstrapped bool
```
oleg-jukovec, line 31: The structure is `private` at all.
```suggestion
	// IsBootstrapped indicates whether instance is bootstrapped.
	IsBootstrapped bool
```

oleg-jukovec (approved):

cli/replicaset/lua/cartridge/get_topology_replicasets_body.lua
```diff
@@ -43,8 +43,18 @@ for _, replicaset in pairs(replicasets) do
 end
 
 local failover_params = require('cartridge').failover_get_params()
+local issues = require('cartridge.issues').list_on_cluster()
+local is_critical = false
+for _, issue in pairs(issues) do
+    if issue.level == 'critical' and issue.instance_uuid == box.info().uuid then
```
psergee, line 49: nitpick: save current instance `uuid` before loop to avoid `box.info` call on every iteration.

cli/replicaset/cartridge.go
```diff
@@ -103,22 +126,30 @@ func (c *CartridgeApplication) GetReplicasets() (Replicasets, error) {
...
+			if topology.isBootstrapped {
+				if newTopology.isBootstrapped {
+					topology = newTopology
+				}
+			} else {
+				topology = newTopology
+			}
+
+			// Stop if we already found a valid topology.
+			return topology.isBootstrapped && !topology.IsCritical, nil
```
psergee, line 145: Looks like, I've missed something. Why do make `topology = newTopology` and only after that check `IsCrtitical`. Do we really need to perform an assignment if new topology is not bootstrapped or has critical issues?

psergee (approved):

cli/replicaset/cartridge.go
askalt: If a bootstrapped instance is found, we never switch it to a non-bootstrapped. And If this instance hasn't critical issues - we stopping the search. But if every instance has critical on it - there's nothing left to us except to take topology from them.

cli/replicaset/lua/cartridge/get_topology_replicasets_body.lua
askalt: Thank you! Changed.

cli/replicaset/cartridge.go
askalt: Thank you! Changed.

cli/replicaset/cartridge.go
psergee: Okay.

psergee:
Please add change log note about logic updates.

PR #758, opened 2024-01-25
test: increase timeouts

askalt:
This patch adds changes to the go integration tests:
* Wait for tarantool start longer: it was noticed that the timeout is too small for macos

* Don't use `log.Fatalf`: deferred code is not executed, which can lead to non stopped instance

### DIFF
--- a/cli/connector/integration_test.go
+++ b/cli/connector/integration_test.go
@@ -3,7 +3,7 @@
 package connector_test
 
 import (
-	"log"
+	"fmt"
 	"net"
 	"os"
 	"testing"
@@ -248,7 +248,7 @@ func TestConnect_textTls(t *testing.T) {
 var poolCases = []struct {
 	Name string
 	Opts []ConnectOpts
-} {
+}{
 	{
 		Name: "single",
 		Opts: []ConnectOpts{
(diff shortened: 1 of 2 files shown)


cli/connector/integration_test.go
```diff
@@ -328,25 +328,28 @@ func runTestMain(m *testing.M) int {
 		WorkDir:      workDir,
 		User:         opts.User,
 		Pass:         opts.Pass,
-		WaitStart:    100 * time.Millisecond,
-		ConnectRetry: 3,
+		WaitStart:    time.Second,
```
psergee, line 331: I'm not sure about a seconds timeout between retries. It seems too big for unit tests (but ok for integration). Can we make it smaller (200 ms maybe) and increase retries count?

cli/connector/integration_test.go
```diff
@@ -328,25 +328,28 @@ func runTestMain(m *testing.M) int {
...
+		ConnectRetry: 5,
 		RetryTimeout: 500 * time.Millisecond,
 	})
 	defer test_helpers.StopTarantoolWithCleanup(inst)
 	if err != nil {
-		log.Fatalf("Failed to prepare test tarantool: %s", err)
+		fmt.Println("Failed to prepare test tarantool:", err)
```
psergee, line 337: Let's replace this err check + print + return code with single `require.NoError` call.

psergee (commented):
Thank you for the patch!

cli/connector/integration_test.go
askalt: `require.NoError` wants `testing.T`, but we have `testing.M` here on setup stage.

cli/connector/integration_test.go
askalt: I have reduced `RetryTimeout` to 200 ms.

askalt: As I see `WaitStart` is waited only one time. And `RetryTimeout` is used after each failure. 
https://github.com/tarantool/go-tarantool/blob/master/test_helpers/main.go#L240-L261

psergee (approved):

PR #764, opened 2024-01-30
install: fix checkout error

psergee:
git checkout fails in tt install due to missing submodule. The solution is to checkout and init submodules separately.


better0fdead (approved):
LGTM

PR #767, opened 2024-01-31
cluster: bump yaml to v3 

askalt:
Since v2 has unexpected artifacts like `on/off` to bool parsing, let's bump this dependency as it was changed in the v3.

Since v3 can parse maps to `map[string|any]string` depending on the weather, let's get rid of ambiguities performing converting to the `map[any]any`.

### DIFF
--- a/cli/cluster/config.go
+++ b/cli/cluster/config.go
@@ -2,8 +2,9 @@ package cluster
 
 import (
 	"fmt"
+	"strings"
 
-	"gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v3"
 )
 
 const (
@@ -14,7 +15,7 @@ const (
 type Config struct {
 	// paths is a container implementation for the deserialized configuration.
 	//
-	// At this moment it is fully-compatible with yaml.v2 Marshal()/Unmarshal()
+	// At this moment it is fully-compatible with yaml.v3 Marshal()/Unmarshal()
 	// functions. So it could be marshaled and unmarshaled directly to YAML.
 	// We may change it in the future.
 	paths any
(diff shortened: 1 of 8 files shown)


cli/cfg/dump.go
```diff
@@ -44,7 +44,9 @@ func dumpConfiguration(writer io.Writer, cmdCtx *cmdcontext.CmdCtx,
 			writer.Write([]byte(cmdCtx.Cli.ConfigPath + ":\n"))
 		}
 	}
-	err := yaml.NewEncoder(writer).Encode(cliOpts)
+	encoder := yaml.NewEncoder(writer)
+	encoder.SetIndent(2)
+	err := encoder.Encode(cliOpts)
 	return err
```
oleg-jukovec, line 50:
```suggestion
	return encoder.Encode(cliOpts)
```

cli/cluster/config.go
```diff
@@ -213,11 +214,36 @@ func (config *Config) Merge(low *Config) {
...
+	case map[any]any:
+		for k, v2 := range x {
+			x[k] = deepCastStringMapToAnyMap(v2)
+		}
+	case []any:
+		for i, v2 := range x {
+			x[i] = deepCastStringMapToAnyMap(v2)
+		}
+	default:
+		break
```
oleg-jukovec, line 235: The same.
```suggestion
```

cli/cluster/config_test.go
```diff
@@ -313,9 +313,9 @@ foo:
   bar: 1
   zoo: 1
 zoo:
-- "1"
-- "2"
-- "3"
+  - "1"
```
oleg-jukovec, line 316: Why are the changes required? As I could see the Tarantool prefers no indentation here in examples:
https://github.com/tarantool/tarantool/blob/master/doc/examples/config/replicaset_manual_failover.yaml#L23

oleg-jukovec (commented):
Let's bump the yaml version only in the cluster module to avoid side-effects.

cli/cluster/config_test.go
askalt: Because of the way how yaml v3 encodes values: https://go.dev/play/p/LaNfVBcRP2J
And here is no way to configure this behavior or I can't see it in the public API: https://pkg.go.dev/gopkg.in/yaml.v3#section-readme

oleg-jukovec: Does Tarantool read updated configs fine?

askalt: Yes, here are no problems.

askalt: We use the same format with indents in some integration tests: https://github.com/tarantool/tt/blob/master/test/integration/running/small_cluster_app/config.yaml

cli/cfg/dump.go
askalt: Now I bump only for the `cluster` as you suggested, so this change is not relevant.

cli/cluster/config.go
askalt: Thank you! Changed.

cli/cluster/config_test.go
oleg-jukovec: https://github.com/tarantool/tarantool/blob/d7b72654799272e541882bea3e16ea79b1b7a36d/doc/examples/config/etcd/local.yaml#L9

oleg-jukovec (approved):
Thank you for the patch.

psergee (approved):
Thank you for the patch! lgtm.

PR #768, opened 2024-02-01
doc: update cluster examples config format

psergee:
`listen` format is changed in latest tarantool version. Update the examples according to the changes.


oleg-jukovec (approved):
Thank you!

PR #776, opened 2024-02-08
start: repairing instance launch on freebsd

0x501D:
This patch fixes the `tt start` bug on FreeBSD:
`LuajitError: stdin:259: Undefined symbol "stdout"`


psergee (approved):

PR #779, opened 2024-02-14
Add examples to tt package

psergee:


LeonidVas (approved):
HI! Thank you for the patchset. Great job.

I have not tested it and believe it works fine.
I have one nit, "I think it is better to rename `replicaset_example` to `replicaset_example_tarantool_3`, because 1 as the default value is ok, but 3..... If we add `replicaset_example_tarantool_4` after that, the order is very strange."

PR #784, opened 2024-02-15
binaries: fix list error if binary is not a link

psergee:
If `tarantool` binary is not a symlink, try to execute it to get version. Use this version in versions list.
Here is an example of output if `tarantool` is a binary file (not a symlink):
```
$ ls -l bin
total 38176
-rwxr-xr-x 1 user user 39089200 фев 15 15:32 tarantool

$ tt binaries list
List of installed binaries:
   • tarantool:               
	3.1.0-entrypoint-83-gcb0264c3c [active]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ empty, for example.
 - `tt pack` with `--use-docker` fails due to incompatible versions of `tt`
 between local system and docker container. Install current `tt` version
 in docker image if possible.
+- `tt binaries list` invalid argument error if `tarantool` is not a symlink.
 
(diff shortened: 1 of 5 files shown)


askalt (approved):
lgtm

askalt:
Is this situation possible if user didn't change tt_env/bin/tarantool by himself?

psergee:
> Is this situation possible if user didn't change tt_env/bin/tarantool by himself?

Yes, it is possible after `tt pack` for example. `tt pack` does not create symlinks, so we will have `tarantool` and `tt` binaries in bin dir in a result package.

PR #788, opened 2024-02-27
ci: remove installing python form ci on macOS

foxzi:
This patch remove installing python on macOS because it broke running tests. On macOS already installed python 3.9.

### DIFF
--- a/.github/workflows/full-ci.yml
+++ b/.github/workflows/full-ci.yml
@@ -185,7 +185,6 @@ jobs:
       - name: Install dependencies
         run: |
           brew install --overwrite go mage
-          brew install python3
           pip3 install -r test/requirements.txt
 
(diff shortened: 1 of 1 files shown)


.github/workflows/full-ci.yml
```diff
@@ -185,7 +185,6 @@ jobs:
       - name: Install dependencies
         run: |
           brew install --overwrite go mage
-          brew install python3
```
psergee, line 188: Do test fail on self-hosted or on github runners?

foxzi: on solf-host, I see this patch can't help

psergee:
This change is for github runners, but the issue with latest python version affects only self-hosted runners. Closing the pull request.

PR #790, opened 2024-02-28
cluster: some improvements 

askalt:
This patch makes the handling of cluster.NotExistError unambiguous. There is one place where we return it by pointer, while in other places it's returned by value. Let's always return it by value.

Also, adds an ability to extract key from the config source. This will be convenient if, after retrieving keys using the DataCollector, we want to upload them back.

### DIFF
--- a/lib/cluster/storages_test.go
+++ b/lib/cluster/storages_test.go
@@ -1,0 +1,33 @@
+package cluster_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/tarantool/tt/lib/cluster"
+)
+
+func TestStorage_GetStorageKey(t *testing.T) {
+	cases := []struct {
+		prefix string
+		source string
+		key    string
+		err    string
+	}{
+		{"/foo", "/foo/config/all", "all", ""},
+		{"/foo/", "/foo/config/all", "all", ""},
+		{"/foo///", "/foo/config/all", "all", ""},
+		{"/foo", "/bar/config/all", "", "source must begin with: /foo/config/"},
+	}
+	for _, cs := range cases {
+		t.Run(cs.source, func(t *testing.T) {
+			key, err := cluster.GetStorageKey(cs.prefix, cs.source)
+			if cs.err != "" {
+				require.EqualError(t, err, cs.err)
+			} else {
+				require.NoError(t, err)
+				require.Equal(t, cs.key, key)
+			}
+		})
+	}
+}
--- a/cli/running/running.go
+++ b/cli/running/running.go
@@ -479,7 +479,7 @@ func mapValuesFromConfig[T any](cfg *libcluster.Config, mapFunc func(val T) (T, 
 	for _, cfgMapping := range maps {
 		value, err := cfg.Get(cfgMapping.path)
 		if err != nil {
-			var eNotExist *libcluster.NotExistError
+			var eNotExist libcluster.NotExistError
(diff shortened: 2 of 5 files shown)


lib/cluster/etcd.go
oleg-jukovec, line 1: Please, move the code to storages.go

lib/cluster/etcd.go
```diff
@@ -805,3 +805,12 @@ func isSameDirSymlink(f fs.DirEntry, dir string) bool {
...
 	return err == nil && !strings.Contains(target, "/")
 }
+
+// GetStorageKey extracts the key from the source that contains the config prefix.
+func GetStorageKey(prefix, source string) (string, error) {
+	cfgPrefix := getConfigPrefix(prefix)
+	if !strings.HasPrefix(source, cfgPrefix) {
+		return "", fmt.Errorf("source must begin with: %s", cfgPrefix)
+	}
+	return source[len(cfgPrefix):], nil
```
oleg-jukovec, line 815: You could use `strings.CutPrefix` to do the thing.

lib/cluster/etcd_test.go
```diff
@@ -666,3 +666,12 @@ func TestEtcdKeyDataPublisher_Publish_timeout(t *testing.T) {
...
 }
+
+func TestEtcd_GetStorageKey(t *testing.T) {
+	key, err := cluster.GetStorageKey("/foo", "/foo/config/all")
+	require.NoError(t, err)
+	require.Equal(t, "all", key)
+
+	_, err = cluster.GetStorageKey("/foo", "/bar/config/all")
+	require.EqualError(t, err, "source must begin with: /foo/config/")
+}
```
oleg-jukovec, line 677: There are two different test cases in fact, not a single test.
```suggestion
func TestGetStorageKey(t *testing.T) {
	key, err := cluster.GetStorageKey("/foo", "/foo/config/all")
	require.NoError(t, err)
	require.Equal(t, "all", key)

	_, err = cluster.GetStorageKey("/foo", "/bar/config/all")
	require.EqualError(t, err, "source must begin with: /foo/config/")
}
```

lib/cluster/etcd.go
askalt: Done, didn't see this new file.

lib/cluster/etcd.go
askalt: Thank you! Applied.

lib/cluster/etcd_test.go
askalt: Separated.

lib/cluster/storages_test.go
```diff
@@ -0,0 +1,30 @@
...
+		source string
+		key    string
+		err    string
+	}{
+		{"/foo/config/all", "all", ""},
+		{"/bar/config/all", "", "source must begin with: /foo/config/"},
+	}
+	for _, cs := range cases {
+		t.Run(cs.source, func(t *testing.T) {
+			key, err := cluster.GetStorageKey("/foo", cs.source)
```
oleg-jukovec, line 21: It would be nice to have tests for prefixes: /foo/, /foo/// too.
```suggestion
			key, err := cluster.GetStorageKey("/foo", cs.source)
```

oleg-jukovec (approved):

psergee (approved):

lib/cluster/storages_test.go
askalt: Added!

PR #795, opened 2024-03-04
cd: add publishing to universal static repo

foxzi:
Because we are using a universal static repo for tarantool since 3.0
version, we need to publish `tt` to simular repo. It helps us avoid
a lot of problems with supporting OS and their versions.

Related https://github.com/tarantool/installer.sh/issues/36


psergee (approved):
Thank you for the patch! LGTM.

ylobankov (approved):

PR #796, opened 2024-03-05
doc: add a note about box.cfg.background usage

### LINKED ISSUE #780: tt watchdog exits in case of tarantool 3.* instance daemonization
How to reproduce: tarantool 3.0.1
Create app.lua:
``` lua
-- /etc/tarantool/instances.enabled/app.lua:
box.cfg{listen=3301,
background=true,
log = 'file:tarantool.log',
pid_file='tnt.pid'}
```
Run `tt start`.
Actual: there is no tt watchdog process. Tarantool instance is running.
Expected: tt watchdog is running and manages child tarantool process.
Tarantool 3.* invocation using a script forks and exits. Which causes `tt watchdog` to exit.

oleg-jukovec:
@TarantoolBot document
Title: User should not to set `box.cfg.background = true` with `tt start`

`tt start` daemonizes a Tarantool process itself. The `box.cfg.background` setting does the same thing from a Tarantool process. These features conflict with each other. As a result, `tt status` shows an invalid status of a Tarantool instance, and it is unable to stop the instance with `tt stop`.

A user should never set `box.cfg.backround = true` in an application code that is started by `tt`.

Closes #780

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ Tarantool-based applications.
   * [Setting Tarantool configuration parameters via environment variables](#setting-tarantool-configuration-parameters-via-environment-variables)
   * [Add current environment binaries location to the PATH variable](#add-current-environment-binaries-location-to-the-path-variable)
 * [Migration from older TT versions](doc/migration_from_older_versions.md)
+* [Known issues](doc/known_issues.md)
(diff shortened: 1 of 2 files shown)


psergee (approved):

PR #838, opened 2024-05-06
start: fix socket path overflow check

### LINKED ISSUE #834: Tarantool 3 config instance fails to use 108 symbols control socket on tt start
The case is follows.

I had created an application based on `tt create vshard_cluster` template. Application path is as follows: `/home/moiseevgeorgy/Development/sandbox/t3-app/cluster-app`. Then I've tried to start an application, and got the following status:
```
INSTANCE                      STATUS          PID
cluster-app:storage-002-a     RUNNING         161460
cluster-app:storage-002-b     RUNNING         161461
cluster-app:router-001-a      NOT RUNNING     
cluster-app:storage-001-a     RUNNING         161458
cluster-app:storage-001-b     RUNNING         161459
```

```
LuajitError: builtin/config/applier/console.lua:40: builtin/box/console.lua:1011: failed to create server unix/:/home/moiseevgeorgy/Development/sandbox/instances.enabled/cluster-app/var/run/router-001-a/tarantool.control: No buffer space available
```

"No buffer space available" is an error often associated with "too long UNIX socket path". At the same time everything is fine with storage instances, even though their socket path is expected to be 1 symbol longer. The difference is as follows.

storage instances had received the following `TT_CONSOLE_SOCKET_DEFAULT`:
```
var/run/storage-001-a/tarantool.control
var/run/storage-001-b/tarantool.control
var/run/storage-002-a/tarantool.control
var/run/storage-002-b/tarantool.control
```
and router instance had received the following one:
```
/home/moiseevgeorgy/Development/sandbox/instances.enabled/cluster-app/var/run/router-001-a/tarantool.control
```

The reason of this difference is as follows. A router full socket path is 108 symbols, which is considered to be ok for Linux, thus it doesn't shortened. A storage full socket path is 109 symbols, which exceeds this limit and is transformed to be relative.
https://github.com/tarantool/tt/blob/a0e5aa8eeab8493cc909093277e3058729e37ea3/cli/running/script_instance.go#L21
https://github.com/tarantool/tt/blob/a0e5aa8eeab8493cc909093277e3058729e37ea3/cli/running/script_instance.go#L115-L117

On the other hand, while working with admin socket, Tarantool adds `unix/:` or `unix/:./` before the path ([see source code](https://github.com/tarantool/tarantool/blob/2467e421cbfd42d0f58c0d27e4e397e730412e6c/src/box/lua/config/applier/console.lua#L6-L11)) and then connect fails  ([see source code](https://github.com/tarantool/tarantool/blob/2467e421cbfd42d0f58c0d27e4e397e730412e6c/src/box/lua/config/applier/console.lua#L38-L41)).

The workaround is rather easy: either make socket path longer or shorter. The proper solution is seems to be "consider additional Tarantool pads while trying to shorten the path", but we may give it a second thought.

psergee:
Closes #834

### DIFF
--- a/cli/running/script_instance_test.go
+++ b/cli/running/script_instance_test.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"testing"
 
@@ -116,6 +117,15 @@ func Test_shortenSocketPath(t *testing.T) {
 		socketPath string
 		basePath   string
 	}
+
+	maxSocketPathLen := maxSocketPathLinux
+	if runtime.GOOS == "darwin" {
+		maxSocketPathLen = maxSocketPathMac
+	}
+	dirLen := maxSocketPathLen - len("/tarantool.control") - 1
+	maxSocketPath := "/" + strings.Repeat("a", dirLen) + "/tarantool.control"
+	require.Equal(t, maxSocketPathLen, len(maxSocketPath))
+
 	tests := []struct {
 		name    string
 		args    args
@@ -165,6 +175,15 @@ func Test_shortenSocketPath(t *testing.T) {
 				basePath:   "",
 			},
 			want:    "../aaaaaaaaaa/tarantool.control",
+			wantErr: true,
+		},
+		{
+			name: "max socket path",
+			args: args{
+				socketPath: maxSocketPath,
+				basePath:   "",
+			},
+			want:    "",
 			wantErr: true,
 		},
 	}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 
 - `tt clean` no longer tries to clean files multiple times.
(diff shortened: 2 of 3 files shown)


cli/running/script_instance_test.go
```diff
@@ -116,6 +117,15 @@ func Test_shortenSocketPath(t *testing.T) {
 		socketPath string
 		basePath   string
 	}
+
+	maxSocketPathLen := maxSocketPathLinux
+	if runtime.GOOS == "darwin" {
+		maxSocketPathLen = maxSocketPathMac
+	}
+	dirLen := maxSocketPathLen - len("/tarantool.control") - 1
+	maxSocketPath := "/" + strings.Repeat("a", dirLen) + "/tarantool.control"
```
oleg-jukovec, line 126: Up to you.
```suggestion
	}

	const socketFile = "tarantool.control"
	dirLen := maxSocketPathLen - len(socketFile) - 2*len(os.PathSeparator)
	maxSocketPath := os.PathSeparator + strings.Repeat("a", dirLen) + os.PathSeparator + socketFile
```

DifferentialOrange (approved):

PR #842, opened 2024-05-15
fix tests on macOS

better0fdead:


.github/workflows/publish.yml
```diff
@@ -138,14 +138,6 @@ jobs:
           cd mage
           go run bootstrap.go
 
-      - name: Build OpenSSL 3.0
```
oleg-jukovec, line 141: So why did we need to not use a custom installation? It was necessary for repeatability of the build.

oleg-jukovec (approved):

.github/workflows/publish.yml
better0fdead: Ok, changed

.github/workflows/full-ci.yml
```diff
@@ -203,6 +203,17 @@ jobs:
...
+      # Etcd can be still running after integration tests when:
+      # 1. pytest recieve SIGALRM (can be caused by pytest-timeout plugin)
+      # 2. when pytest crashes
+      # In both cases finalizers are not invoked, so etcd is not stopped.
+      - name: Kill etcd, if it was left after integration tests
+        if: always()
+        run: |
+          ETCD_TT_BIN=${ETCD_PATH}etcd;
+          ETCD_TT_PID=$( ps aux | grep ${ETCD_TT_BIN} | grep -vw grep | awk '{ print $2 }' );
+          kill -9 ${ETCD_TT_PID} || true
```
psergee, line 215: Can we try to kill it softly at first?

.github/workflows/full-ci.yml
```diff
@@ -203,6 +203,17 @@ jobs:
...
 
+      # Etcd can be still running after integration tests when:
+      # 1. pytest recieve SIGALRM (can be caused by pytest-timeout plugin)
+      # 2. when pytest crashes
+      # In both cases finalizers are not invoked, so etcd is not stopped.
+      - name: Kill etcd, if it was left after integration tests
+        if: always()
+        run: |
+          ETCD_TT_BIN=${ETCD_PATH}etcd;
+          ETCD_TT_PID=$( ps aux | grep ${ETCD_TT_BIN} | grep -vw grep | awk '{ print $2 }' );
```
psergee, line 214: Isn't there something like `pgrep` or `pidof` command?

psergee (approved):

.github/workflows/full-ci.yml
better0fdead: I tested those commands locally on macOs, they doesn't seem to find/kill process by path only by binary name.
I don't want to have a possibility that our tests will kill someone else's 'etcd'.

.github/workflows/full-ci.yml
better0fdead: Yes, tested, changed.

.github/workflows/full-ci.yml
psergee: `pgrep --full` isn't supported?
`pidof` also works with full paths on Linux:
```
$ pidof /opt/tarantool/tarantool
5418
```

> I don't want to have a possibility that our tests will kill 'etcd' that was not started by our job.

Sure.

.github/workflows/full-ci.yml
```diff
@@ -203,6 +203,17 @@ jobs:
...
+          kill -2 ${ETCD_TT_PID} || true
```
psergee, line 215: Please, use a signal name instead of a code.

.github/workflows/full-ci.yml
better0fdead:
```
f.terekhin@f-terekhin sharded-queue-enterprise % pgrep --full
pgrep: illegal option -- -
```
pidof is not pre-installed, needs to be installed separately.

better0fdead: `pidof` is not pre-installed on macOs. However, `pkill -f` should work in our case. Changed

psergee (approved):

PR #845, opened 2024-05-21
Reuse tt in tt-ee

DifferentialOrange:
Minor refactoring for https://github.com/tarantool/tt-ee/issues/171


cli/cluster/cmd/common.go
```diff
@@ -149,10 +149,10 @@ func connectTarantool(uriOpts UriOpts, connOpts connectOpts) (tarantool.Connecto
 		connectorOpts.User = connOpts.Username
 		connectorOpts.Pass = connOpts.Password
 		if connectorOpts.User == "" {
-			connectorOpts.User = os.Getenv(connect.TarantoolUsernameEnv)
+			connectorOpts.User = os.Getenv(libconnect.TarantoolUsernameEnv)
```
psergee, line 152: Why not to use the original package name `connect` instead of `libconnect`?

DifferentialOrange: `"github.com/tarantool/tt/cli/connect"` is imported in `tt/cli/cmd/connect.go`. I think it may be confusing to have different `connect`s inside the different files of the same module `cmd`.

DifferentialOrange: Removed renaming for modules, except for `cmd` where naming clashes.

psergee (approved):
lgtm

PR #859, opened 2024-06-05
connect: support Tarantool tuple format

### LINKED ISSUE #832: Improve data display in table/ttable mode
Improve data display in table/ttable mode by adding a format information (fields name).
Only for tarantool >= 3.0

DerekBum:
Support format for Tarantool tuples. This support only works for Tarantool versions >= 3.2.

Closes https://github.com/tarantool/tt/issues/832

@TarantoolBot document
Title: Support Tarantool tuple format

This patch adds support for formatting Tarantool tuples. Installed Tarantool version must be >= 3.2. Otherwise, there will be no change.

Examples:
```
> f = box.tuple.format.new({{'id', 'number'}, {'name', 'string'}})
---
...

> t = box.tuple.new({1, 'Flint', true}, {format = f})
---
...

> \xt
> t
+----+-------+------+
| id | name  | col1 |
+----+-------+------+
| 1  | Flint | true |
+----+-------+------+
> box.space.customers:select()
+----+-----------+-----+
| id | name      | age |
+----+-----------+-----+
| 1  | Elizabeth | 12  |
+----+-----------+-----+
| 2  | Mary      | 46  |
+----+-----------+-----+
> \xT
> t
+------+-------+
| id   | 1     |
+------+-------+
| name | Flint |
+------+-------+
| col1 | true  |
+------+-------+
> box.space.customers:select()
+------+-----------+------+
| id   | 1         | 2    |
+------+-----------+------+
| name | Elizabeth | Mary |
+------+-----------+------+
| age  | 12        | 46   |
+------+-----------+------+
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   * `--lines` number of lines to print.
   * `--follow` print appended data as log files grow.
 
+### Fixed
+
+- Sorted by name order of columns for `table` and `ttable` formats.
+
 ## [2.3.1] - 2024-06-13
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `tt log`: a module for viewing instances logs. Supported options:
   * `--lines` number of lines to print.
   * `--follow` print appended data as log files grow.
+- `tt connect`: support format for Tarantool tuples for Tarantool
+  versions >= 3.2.
 
 ### Fixed
 
--- a/cli/connect/connect.go
+++ b/cli/connect/connect.go
@@ -97,23 +97,24 @@ func Eval(connectCtx ConnectCtx, connOpts connector.ConnectOpts, args []string) 
 	}
 	defer conn.Close()
 
-	var eval string
-	evalArgs := []interface{}{command}
+	evalArgs := []interface{}{command, connectCtx.Language == SQLLanguage}
 	if connectCtx.Language != DefaultLanguage {
 		// Change a language.
 		if err := ChangeLanguage(conn, connectCtx.Language); err != nil {
 			return nil, fmt.Errorf("unable to change a language: %s", err)
 		}
-		eval = consoleEvalFuncBody
+		evalArgs = append(evalArgs, false)
 	} else {
-		eval = evalFuncBody
+		needMetaInfo := connectCtx.Format == formatter.TableFormat ||
+			connectCtx.Format == formatter.TTableFormat
+		evalArgs = append(evalArgs, needMetaInfo)
 		for i := range args {
 			evalArgs = append(evalArgs, args[i])
 		}
 	}
 
 	// Execution of the command.
-	response, err := conn.Eval(eval, evalArgs, connector.RequestOpts{})
+	response, err := conn.Eval(evalFuncBody, evalArgs, connector.RequestOpts{})
 	if err != nil {
 		return nil, err
 	}
--- a/cli/connect/console.go
+++ b/cli/connect/console.go
@@ -218,7 +218,10 @@ func getExecutor(console *Console) func(string) {
 		}
 
 		var results []string
-		args := []interface{}{console.input}
+		needMetaInfo := console.format == formatter.TableFormat ||
+			console.format == formatter.TTableFormat
+		args := []interface{}{console.input, console.language == SQLLanguage,
+			needMetaInfo}
 		opts := connector.RequestOpts{
 			PushCallback: func(pushedData interface{}) {
 				encodedData, err := yaml.Marshal(pushedData)
@@ -233,7 +236,7 @@ func getExecutor(console *Console) func(string) {
 		}
 
 		var data string
-		if _, err := console.conn.Eval(consoleEvalFuncBody, args, opts); err != nil {
+		if _, err := console.conn.Eval(evalFuncBody, args, opts); err != nil {
 			if err == io.EOF {
 				// We need to call 'console.Close()' here because in some cases (e.g 'os.exit()')
 				// it won't be called from 'defer console.Close' in 'connect.runConsole()'.
--- a/cli/connect/language_test.go
+++ b/cli/connect/language_test.go
@@ -2,6 +2,7 @@ package connect_test
 
 import (
 	"errors"
+	"os"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -82,7 +83,11 @@ func (evaler *inputEvaler) Eval(fun string,
 }
 
 func TestChangeLanguage_requestInputs(t *testing.T) {
-	expectedFun := "return require('console').eval(...)\n"
+	rawFun, err := os.ReadFile("./lua/eval_func_body.lua")
+	if err != nil {
+		t.Fatal("Failed to read lua file:", err)
+	}
+	expectedFun := string(rawFun)
 	expectedOpts := connector.RequestOpts{}
 	cases := []struct {
 		lang Language
--- a/cli/connect/lua/eval_func_body.lua
+++ b/cli/connect/lua/eval_func_body.lua
@@ -1,6 +1,18 @@
 local yaml = require('yaml')
+yaml.cfg{ encode_use_tostring = true }
 local args = {...}
 local cmd = table.remove(args, 1)
+local is_sql_language = table.remove(args, 1)
+local need_metainfo = table.remove(args, 1)
+
+local function is_command(line)
+    return line:sub(1, 1) == '\\'
+end
+
+if is_command(cmd) or is_sql_language == true then
+    return require('console').eval(cmd)
+end
+
 local fun, errmsg = loadstring("return "..cmd)
 if not fun then
     fun, errmsg = loadstring(cmd)
@@ -13,9 +25,26 @@ local function table_pack(...)
     return {n = select('#', ...), ...}
 end
 
+local function format_equal(f1, f2)
+    if #f1 ~= #f2 then
+        return false
+    end
+
+    for i, field in ipairs(f1) do
+        if field.name ~= f2[i].name or field.type ~= f2[i].type then
+            return false
+        end
+    end
+    return true
+end
+
 local ret = table_pack(pcall(fun, unpack(args)))
 if not ret[1] then
-    return yaml.encode({box.NULL})
+    local err = unpack(ret, 2, ret.n)
+    if err == nil then
+        err = box.NULL
+    end
+    return yaml.encode({{error = err}})
 end
 if ret.n == 1 then
     return "---\n...\n"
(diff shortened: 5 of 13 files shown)


cli/connect/console.go
```diff
@@ -233,7 +235,7 @@ func getExecutor(console *Console) func(string) {
 		}
 
 		var data string
-		if _, err := console.conn.Eval(consoleEvalFuncBody, args, opts); err != nil {
+		if _, err := console.conn.Eval(evalFuncBody, args, opts); err != nil {
```
oleg-jukovec, line 238: It seems like with the `evalFuncBody` `\set language sql` and `\set delimiter` won't work anymore. So we need tests for the cases + update `consoleEvalFuncBody` in the `console.eval` implementation manner too.

oleg-jukovec (changes requested):

cli/connect/console.go
DerekBum: Updated.

cli/connect/connect.go
```diff
@@ -97,23 +97,23 @@ func Eval(connectCtx ConnectCtx, connOpts connector.ConnectOpts, args []string)
...
 	defer conn.Close()
 
-	var eval string
-	evalArgs := []interface{}{command}
+	evalArgs := []interface{}{command, connectCtx.Language == SQLLanguage}
 	if connectCtx.Language != DefaultLanguage {
 		// Change a language.
 		if err := ChangeLanguage(conn, connectCtx.Language); err != nil {
 			return nil, fmt.Errorf("unable to change a language: %s", err)
 		}
```
oleg-jukovec, line 105: The change will help to be clear:
```suggestion
		}
		evalArgs = append(evalArgs, false)
```

cli/connect/language_test.go
```diff
@@ -81,29 +81,6 @@ func (evaler *inputEvaler) Eval(fun string,
...
-	for _, c := range cases {
-		t.Run(c.lang.String(), func(t *testing.T) {
-			evaler := &inputEvaler{}
-			ChangeLanguage(evaler, c.lang)
-			assert.Equal(t, expectedFun, evaler.fun)
-			assert.Equal(t, c.arg, evaler.args[0].(string))
-			assert.Equal(t, expectedOpts, evaler.opts)
-		})
-	}
-}
```
oleg-jukovec, line 105: It seems to me that it would be better to update the test according to the new implementation.

cli/connect/lua/eval_func_body.lua
```diff
@@ -1,6 +1,22 @@
...
+    if line:sub(1, 1) == '\\' then
+        return line:sub(2)
+    end
+    return nil
+end
+
+local command = get_command(cmd)
+if command or is_sql_language == true then
+    return require('console').eval(cmd)
+end
```
oleg-jukovec, line 18: It seems to me that this will only make things easier:
```suggestion
local function is_command(line)
    return line:sub(1, 1) == '\\'
end

if is_command(cmd) or is_sql_language == true then
    return require('console').eval(cmd)
end
```

cli/connect/lua/eval_func_body.lua
```diff
@@ -13,9 +29,26 @@ local function table_pack(...)
...
+
 local ret = table_pack(pcall(fun, unpack(args)))
 if not ret[1] then
-    return yaml.encode({box.NULL})
+    local err = unpack(ret, 2, ret.n)
+    if err == nil then
+        err = box.NULL
+    end
+    return yaml.encode({{error = err}})
```
oleg-jukovec, line 51: Is it the same behavior as before? I guess that just {err} returned in the case.

cli/connect/lua/eval_func_body.lua
```diff
@@ -25,4 +58,46 @@ for i=2,ret.n do
...
+
+if ret.n == 2 and type(ret[2]) == 'cdata' and ret[2] ~= nil and ret[2].format then
+    local ret_with_format = {
+        metadata = ret[2]:format(),
+        rows = { ret[2] },
+    }
+    if #ret_with_format.metadata > 0 then
+        ret[2] = ret_with_format
+    end
+end
```
oleg-jukovec, line 73: What if we have ret[3] and ret[4], or ret[6]? It should be showed as two different tables in the current logic.

Please, add a test for it.

cli/connect/lua/eval_func_body.lua
```diff
@@ -25,4 +58,46 @@ for i=2,ret.n do
...
+
+if ret.n == 2 and type(ret[2]) == 'table' and #ret[2] > 0 then
```
oleg-jukovec, line 75: The same about ret[3], ret[4] etc. Here we need to write more general code.

cli/connect/lua/eval_func_body.lua
```diff
@@ -25,4 +58,46 @@ for i=2,ret.n do
...
+    }
+
+    local same_format = true
+    for i, tuple in ipairs(ret[2]) do
+        local is_tuple = type(tuple) == 'cdata' and tuple.format
+
+        if i == 1 then
+            if not is_tuple then
+               same_format = false
+                break
```
oleg-jukovec, line 88:
```suggestion
                same_format = false
                break
```

cli/formatter/unordered_map.go
```diff
@@ -0,0 +1,22 @@
+package formatter
+
+type unorderedMap[T comparable] struct {
```
oleg-jukovec, line 3: It would be nice to have `ForEach()` method here. It could be `orderedMap` than.

cli/formatter/node.go
```diff
@@ -12,6 +12,8 @@ const (
 // getNodeType returns renderNode type.
 func getNodeType(node any) nodeType {
 	switch node.(type) {
+	case unorderedMap[any]:
```
oleg-jukovec, line 15: What about unorderedMap[string]?

cli/formatter/table.go
```diff
@@ -24,15 +24,41 @@ func decodeYamlArr(input string) ([]any, error) {
 	return decoded, nil
 }
 
-// castAnyMapToStringMap casts map[any]any to map[string]any.
-func castAnyMapToStringMap(src map[any]any) map[string]any {
-	dst := make(map[string]any)
+// castAnyMapToStringUMap casts map[any]any to unorderedMap[string].
+// The result will be sorted by keys.
+func castAnyMapToStringUMap(src map[any]any) unorderedMap[string] {
```
oleg-jukovec, line 29: Sorry, but I don't get the idea: why do we need unorderedMap[string] and unorderedMap[any] at the same time? Why do not enough unorderedMap[any]?

oleg-jukovec (changes requested):
Please, mark the pull request as "Ready for review" if it is already ready.

We need more tests for cases:

```
return tuple1, tuple2, a number/table
```
and
```
return tuples1, tuples2, a number/table
```

We the same or mixed format. At now I see tests only for (or I missed something):

```
return tuple
```
```
return {tuple, a number/table}
```
or
```
return tuples
```

cli/connect/lua/eval_func_body.lua
DerekBum: This is the same behaviour as Tarantool `console.eval`. It is made that way to pass the test with `error('blah blah')` console command.

cli/formatter/node.go
DerekBum: It's still `mapNodeType`. unorderedMap[string] satisfies unorderedMap[any] case.

cli/formatter/unordered_map.go
DerekBum: So you suggesting something like that?
```go
func (u *unorderedMap[T]) ForEach() map[T]any {
	return u.innerMap
}
```
Or like
```go
func (u *unorderedMap[T]) ForEach(f func(key T, value any)) {
	for _, key := range u.keys {
		f(key, u.innerMap[key])
	}
}
```

oleg-jukovec: The second.

cli/connect/lua/eval_func_body.lua
oleg-jukovec: Please, add the test with `Lua` output format type.

We also need some tests with tuples + format and Lua/YAML output type.

DerekBum: There are already test added for Lua type:
https://github.com/tarantool/tt/blob/8f4d3182f21d2b8ce663c8bd207c3a7e80b969a4/test/integration/connect/test_connect.py#L949-L956
(I will add some tests for yaml).

Tests for errors for table and lua already exist:
https://github.com/tarantool/tt/blob/8f4d3182f21d2b8ce663c8bd207c3a7e80b969a4/test/integration/connect/test_connect.py#L1224-L1231

https://github.com/tarantool/tt/blob/8f4d3182f21d2b8ce663c8bd207c3a7e80b969a4/test/integration/connect/test_connect.py#L908-L914

oleg-jukovec: Thank you, great!

cli/formatter/unordered_map.go
DerekBum: I don't know if this is necessary since unorderedMap is also private. Or do you want to make it public?

cli/connect/connect.go
DerekBum: Updated.

cli/connect/language_test.go
DerekBum: Restored and updated this test.

cli/connect/lua/eval_func_body.lua
DerekBum: Updated.

cli/connect/lua/eval_func_body.lua
DerekBum: Updated. New tests added.

cli/connect/lua/eval_func_body.lua
DerekBum: Updated.

cli/connect/lua/eval_func_body.lua
DerekBum: Updated.

cli/formatter/table.go
DerekBum: Removed all unorderedMap[string] (unorderedMap[any] is used).

cli/formatter/node.go
DerekBum: Anyway, because of the fix for comment below, unorderedMap[string] is no longer present int the code.

cli/formatter/unordered_map.go
DerekBum: Ok, added this method. Also used it in some `table.go` functions.

cli/formatter/yaml.go
```diff
@@ -0,0 +1,20 @@
+package formatter
+
+// LazyMessage is used for lazy Unmarshal.
+type LazyMessage struct {
+	unmarshal func(any) error
```
oleg-jukovec, line 5: Why is the type public?

oleg-jukovec (approved):
Thank you for the patch!

CHANGELOG.md
```diff
@@ -28,13 +28,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `tt pack`: systemd unit parameterizing support.
 - `tt replicaset vshard`: module to manage vshard in the tarantool replicaset.
   * `tt replicaset vshard bootstrap`: command to bootstrap vshard.
+- `tt connect`: support format for Tarantool tuples for a Tarantool
```
better0fdead, line 31: move changelog entries to unreleased section

CHANGELOG.md
```diff
@@ -28,13 +28,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+  version >= 3.2.
 
 ### Fixed
 
 - `tt clean` no longer tries to clean files multiple times.
 - Tarantool 3 config instance fails to use 108 symbols control socket on tt start.
 - Incorrect application name in case of explicit providing config path without directories.
 - Application build failure during pack with Tarantool from the current environment.
+- Sorted by name order of columns for `table` and `ttable` formats.
```
better0fdead, line 40: move changelog entries to unreleased section

better0fdead (approved):
Hi thx for patch

CHANGELOG.md
DerekBum: Oops, updated.

CHANGELOG.md
DerekBum: Oops, updated.

oleg-jukovec:
Please, rebase to the `master` branch.

> Conflicting files
> CHANGELOG.md

DerekBum:
Rebased this branch.

cli/connect/lua/eval_func_body.lua
```diff
@@ -25,4 +54,49 @@ for i=2,ret.n do
...
+
+    if type(ret[i]) == 'table' and #ret[i] > 0 then
+        local ret_with_format = {
+            metadata = {},
+            rows = ret[i],
+        }
+
+        local same_format = true
+        for tuple_ind, tuple in ipairs(ret[i]) do
+            local is_tuple = type(tuple) == 'cdata' and tuple ~= nil and tuple.format
```
oleg-jukovec, line 80: The same.

cli/connect/lua/eval_func_body.lua
```diff
@@ -25,4 +54,49 @@ for i=2,ret.n do
         ret[i] = box.NULL
     end
 end
+if not need_metainfo then
+    return yaml.encode({unpack(ret, 2, ret.n)})
+end
+
+for i=2,ret.n do
+    if type(ret[i]) == 'cdata' and ret[i] ~= nil and ret[i].format then
```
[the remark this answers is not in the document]
DerekBum: Thanks, updated.

cli/connect/lua/eval_func_body.lua
DerekBum: Thanks, updated.

PR #892, opened 2024-07-04
uninstall: fix removing tt by version

### LINKED ISSUE #583: tt install and uninstall version format is inconsistent
Uninstall does not work with the version specified during installation.
```
$ tt install tt 1.2.0
   • Checking existing...     
   • Installing tt=v1.2.0     
   • Checking dependencies... 
   • Downloading tt...        
   • Building tt...           
   • Copying executable...    
   • Done.                    
$ ls -l bin
total 13232
lrwxrwxrwx 1 user user        9 Aug 29 17:34 tt -> tt_v1.2.0
-rwxrwxr-x 1 user user 13549568 Aug 29 17:34 tt_v1.2.0

$ tt uninstall tt 1.2.0
   • Removing binary...       
   ⨯ there is no tt installed

$ tt uninstall tt v1.2.0
   • Removing binary...       
   • tt=v1.2.0 is uninstalled.
```

themilchenko:
It was impossible to uninstall tt by version without 'v' prefix using `tt unistall tt 1.2.0` for example because program checks version in binary file with prefix.

After the patch `tt uninstall` checks 'v' prefix before version and if it didn't find, adds for strings matching.

Closes #583

### DIFF
--- a/cli/install/install.go
+++ b/cli/install/install.go
@@ -68,11 +68,15 @@ var unwindPatch []byte
 //go:embed extra/bump-libunwind-new.patch
 var unwindPatchNew []byte
 
-// defaultDirPermissions is rights used to create folders.
-// 0755 - drwxr-xr-x
-// We need to give permission for all to execute
-// read,write for user and only read for others.
-const defaultDirPermissions = 0755
+const (
+	// defaultDirPermissions is rights used to create folders.
+	// 0755 - drwxr-xr-x
+	// We need to give permission for all to execute
+	// read,write for user and only read for others.
+	defaultDirPermissions = 0755
+
+	MajorMinorPatchRegexp = `^[0-9]+\.[0-9]+\.[0-9]+`
+)
 
 // programGitRepoUrls contains URLs of programs git repositories.
 var programGitRepoUrls = map[string]string{
@@ -494,8 +498,13 @@ func installTt(binDir string, installCtx InstallCtx, distfiles string) error {
 			if err != nil {
 				return err
 			}
+
+			versionMatches, err := regexp.Match(MajorMinorPatchRegexp, []byte(ttVersion))
+			if err != nil {
+				return err
+			}
 			for _, ver := range versions {
-				if ttVersion == ver.Str || (ttVersion[:1] != "v" && "v"+ttVersion == ver.Str) {
+				if ttVersion == ver.Str || (versionMatches && "v"+ttVersion == ver.Str) {
 					versionFound = true
 					ttVersion = ver.Str
 					break
--- a/cli/uninstall/uninstall.go
+++ b/cli/uninstall/uninstall.go
@@ -26,6 +26,8 @@ const (
 	verRegexp = "(?P<ver>.*)"
 )
 
+var errNotInstalled = errors.New("program is not installed")
+
 // remove removes binary/directory and symlinks from directory.
 // It returns true if symlink was removed, error.
 func remove(program string, programVersion string, directory string,
@@ -53,7 +55,7 @@ func remove(program string, programVersion string, directory string,
 	path := filepath.Join(directory, fileName)
 
 	if _, err := os.Stat(path); os.IsNotExist(err) {
-		return false, fmt.Errorf("there is no %s installed", program)
+		return false, errNotInstalled
 	} else if err != nil {
 		return false, fmt.Errorf("there was some problem locating %s", path)
 	}
@@ -113,11 +115,25 @@ func UninstallProgram(program string, programVersion string, binDst string, head
 		}
 	}
 
-	var isSymlinkRemoved bool
-	isSymlinkRemoved, err = remove(program, programVersion, binDst, cmdCtx)
+	versionsToDelete, err := getAllTtVersionFormats(program, programVersion)
 	if err != nil {
 		return err
 	}
+
+	var isSymlinkRemoved bool
+	for _, verToDel := range versionsToDelete {
+		isSymlinkRemoved, err = remove(program, verToDel, binDst, cmdCtx)
+		if err != nil && !errors.Is(err, errNotInstalled) {
+			return err
+		}
+		if err == nil {
+			break
+		}
+	}
+	if err != nil {
+		return err
+	}
+
 	if strings.Contains(program, "tarantool") {
 		log.Infof("Removing headers...")
 		_, err = remove(program, programVersion, headerDst, cmdCtx)
@@ -131,6 +147,26 @@ func UninstallProgram(program string, programVersion string, binDst string, head
 		err = switchProgramToLatestVersion(program, binDst, headerDst)
 	}
 	return err
+}
+
+// getAllTtVersionFormats returns all version formats with 'v' prefix and
+// without it before x.y.z version.
+func getAllTtVersionFormats(programName, ttVersion string) ([]string, error) {
+	versionsToDelete := []string{ttVersion}
+
+	if programName == search.ProgramTt {
+		// Need to determine if we have x.y.z format in tt uninstall argument
+		// to make sure we add version prefix.
+		versionMatches, err := regexp.Match(install.MajorMinorPatchRegexp, []byte(ttVersion))
+		if err != nil {
+			return versionsToDelete, err
+		}
+		if versionMatches {
+			versionsToDelete = append(versionsToDelete, "v"+ttVersion)
+		}
+	}
+
+	return versionsToDelete, nil
 }
 
 // getDefault returns a default version of an installed program.
--- a/cli/uninstall/uninstall_test.go
+++ b/cli/uninstall/uninstall_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/tarantool/tt/cli/configure"
+	"github.com/tarantool/tt/cli/search"
 	"github.com/tarantool/tt/cli/version"
 )
 
@@ -140,6 +141,55 @@ func TestSearchLatestVersion(t *testing.T) {
 			} else {
 				assert.Error(t, err)
 			}
+		})
+	}
+}
+
+func TestGetAllVersionFormats(t *testing.T) {
+	type testCase struct {
+		name             string
+		programName      string
+		ttVersion        string
+		expectedVersions []string
+		expectedError    error
+	}
+
+	cases := []testCase{
+		{
+			name:             "without prefix",
+			programName:      search.ProgramTt,
+			ttVersion:        "1.2.3",
+			expectedVersions: []string{"1.2.3", "v1.2.3"},
+			expectedError:    nil,
+		},
+		{
+			name:             "with prefix",
+			programName:      search.ProgramTt,
+			ttVersion:        "v1.2.3",
+			expectedVersions: []string{"v1.2.3"},
+			expectedError:    nil,
+		},
+		{
+			name:             "not tt program",
+			programName:      search.ProgramCe,
+			ttVersion:        "1.2.3",
+			expectedVersions: []string{"1.2.3"},
+			expectedError:    nil,
+		},
+		{
+			name:             "not <major.minor.patch> format",
+			programName:      search.ProgramCe,
+			ttVersion:        "e902206",
+			expectedVersions: []string{"e902206"},
+			expectedError:    nil,
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			ttVersions, err := getAllTtVersionFormats(tc.programName, tc.ttVersion)
+			assert.NoError(t, err)
+			assert.Equal(t, ttVersions, tc.expectedVersions)
 		})
 	}
 }
--- a/test/integration/uninstall/test_uninstall.py
+++ b/test/integration/uninstall/test_uninstall.py
@@ -6,7 +6,8 @@ import subprocess
 
 import pytest
 
-from utils import config_name, is_valid_tarantool_installed
+from utils import (config_name, is_valid_tarantool_installed,
+                   run_command_and_get_output)
 
 
 def test_uninstall_tt(tt_cmd, tmp_path):
@@ -104,7 +105,7 @@ def test_uninstall_missing(tt_cmd, tmp_path):
     uninstall_output = uninstall_process.stdout.readline()
     assert re.search(r"Removing binary...", uninstall_output)
     uninstall_output = uninstall_process.stdout.readline()
-    assert re.search(r"there is no", uninstall_output)
+    assert "program is not installed" in uninstall_output
 
 
 def test_uninstall_foreign_program(tt_cmd, tmpdir_with_cfg):
@@ -306,3 +307,44 @@ def test_uninstall_tarantool_no_switch(tt_cmd, tmp_path):
         os.path.join(testdata_path, "bin", "tarantool_2.10.4"),
         os.path.join(testdata_path, "inc", "include", "tarantool_2.10.4")
     )
+
+
+@pytest.mark.parametrize("installed_versions, version_to_uninstall", [
+    (["v1.2.3"], "1.2.3"),
+    (["v1.2.3", "1.2.3"], "1.2.3"),
+    (["v1.2.3", "1.2.3"], "v1.2.3")
+])
+def test_uninstall_tt_missing_version_character(
+        installed_versions,
+        version_to_uninstall,
+        tt_cmd,
+        tmp_path):
+    configPath = os.path.join(tmp_path, config_name)
+    # Create test config.
+    with open(configPath, 'w') as f:
+        f.write(f'env:\n bin_dir: {tmp_path}\n inc_dir:\n')
+
+    # Copying built executable to bin directory,
+    # renaming it to version and change symlink
+    # to emulate that we have installed it like
+    # tt install tt v1.2.3.
+    # We don't really install it because bug we test
+    # is fixed only for current patch. Due to
+    # binary file is changing while running this
+    # bug won't appear only for current and future versions.
+    for version in installed_versions:
+        shutil.copy(tt_cmd, os.path.join(tmp_path, f"tt_{version}"))
+    os.symlink(os.path.join(tmp_path, f"tt_{installed_versions[0]}"),
+               os.path.join(tmp_path, "tt"))
+
+    # Remove not installed program.
+    uninstall_cmd = [tt_cmd, "--cfg", configPath, "uninstall", "tt", version_to_uninstall]
+    uninstall_rc, uninstall_output = run_command_and_get_output(uninstall_cmd, cwd=tmp_path)
+
+    assert uninstall_rc == 0
+    assert f"tt={version_to_uninstall} is uninstalled." in uninstall_output
+
+    # Check that the passed version has been removed.
+    assert os.path.isfile(os.path.join(tmp_path, "tt_" +
+                          "v" if "v" not in version_to_uninstall else "" +
+                                       version_to_uninstall)) is False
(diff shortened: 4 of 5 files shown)


cli/uninstall/uninstall.go
```diff
@@ -111,6 +111,11 @@ func UninstallProgram(program string, programVersion string, binDst string, head
 		if programVersion, err = getDefault(program, binDst); err != nil {
 			return err
 		}
+	} else if program == search.ProgramTt &&
+		strings.Contains(programVersion, ".") && programVersion[0] != 'v' {
```
oleg-jukovec, line 115: https://github.com/tarantool/tt/blob/e90220698092b936858d8ed34b101b0b911fe6a5/cli/install/install.go#L498

For fully symmetry, we need to try to delete two versions (with 'v' and without). Since the `install` could install `1.2.3-xxx-foo-bar` if the tag exist.

test/integration/uninstall/test_uninstall.py
```diff
@@ -306,3 +306,36 @@ def test_uninstall_tarantool_no_switch(tt_cmd, tmp_path):
...
+    # We don't really install it because bug we test
+    # is fixed only for current patch. Due to
+    # binary file is changing while running this
+    # bug won't appear only for current and future versions.
+    shutil.copy(tt_cmd, os.path.join(tmp_path, "tt_v1.2.3"))
+    os.symlink(os.path.join(tmp_path, "tt_v1.2.3"),
+               os.path.join(tmp_path, "tt"))
+
+    # Remove not installed program.
+    uninstall_cmd = [tt_cmd, "--cfg", configPath, "uninstall", "tt", "1.2.3"]
```
oleg-jukovec, line 330: Please, add a test case with v1.2.3.
```suggestion
    uninstall_cmd = [tt_cmd, "--cfg", configPath, "uninstall", "tt", "1.2.3"]
```

cli/uninstall/uninstall.go
themilchenko: Fixed.

test/integration/uninstall/test_uninstall.py
themilchenko: Added.

cli/uninstall/uninstall.go
```diff
@@ -112,12 +112,30 @@ func UninstallProgram(program string, programVersion string, binDst string, head
 			return err
 		}
 	}
+	// We need to try to delete all versions with 'v' prefix before x.y.z version
+	// and without it.
+	versionsToDelete := make([]string, 0, 2)
+	versionsToDelete = append(versionsToDelete, programVersion)
```
oleg-jukovec, line 118: I don't think we need such micro-optimizations here. The code is just harder to read.
```suggestion
	versionsToDelete := []string{programVersion}
```

cli/uninstall/uninstall.go
```diff
@@ -112,12 +112,30 @@ func UninstallProgram(program string, programVersion string, binDst string, head
...
+
+	if program == search.ProgramTt &&
+		strings.Contains(programVersion, ".") && programVersion[0] != 'v' {
+		versionsToDelete = append(versionsToDelete, "v"+programVersion)
+	} else if program == search.ProgramTt &&
+		strings.Contains(programVersion, ".") && programVersion[0] == 'v' {
+		versionsToDelete = append(versionsToDelete, programVersion[1:])
+	}
```
oleg-jukovec, line 126: Nit, tt install does not have such logic.
```suggestion
	}
```

cli/uninstall/uninstall.go
```diff
@@ -112,12 +112,30 @@ func UninstallProgram(program string, programVersion string, binDst string, head
...
 		}
 	}
+	// We need to try to delete all versions with 'v' prefix before x.y.z version
+	// and without it.
+	versionsToDelete := make([]string, 0, 2)
+	versionsToDelete = append(versionsToDelete, programVersion)
+
+	if program == search.ProgramTt &&
+		strings.Contains(programVersion, ".") && programVersion[0] != 'v' {
+		versionsToDelete = append(versionsToDelete, "v"+programVersion)
```
oleg-jukovec, line 122: Please, add a comment why we need the logic here.

oleg-jukovec:
We need a changelog entry too.

cli/uninstall/uninstall.go
```diff
@@ -112,12 +112,30 @@ func UninstallProgram(program string, programVersion string, binDst string, head
 			return err
 		}
 	}
+	// We need to try to delete all versions with 'v' prefix before x.y.z version
+	// and without it.
+	versionsToDelete := make([]string, 0, 2)
+	versionsToDelete = append(versionsToDelete, programVersion)
+
+	if program == search.ProgramTt &&
+		strings.Contains(programVersion, ".") && programVersion[0] != 'v' {
```
oleg-jukovec, line 121:
```suggestion
		programVersion != 'master' && programVersion[0] != 'v' {
```
The `tt install` logic does not contain such checks.

cli/uninstall/uninstall.go
```diff
@@ -112,12 +112,30 @@ func UninstallProgram(program string, programVersion string, binDst string, head
...
+	} else if program == search.ProgramTt &&
+		strings.Contains(programVersion, ".") && programVersion[0] == 'v' {
+		versionsToDelete = append(versionsToDelete, programVersion[1:])
+	}
 
 	var isSymlinkRemoved bool
-	isSymlinkRemoved, err = remove(program, programVersion, binDst, cmdCtx)
+	for _, verToDel := range versionsToDelete {
+		isSymlinkRemoved, err = remove(program, verToDel, binDst, cmdCtx)
+		if err == nil {
```
psergee, line 131: If the first `remove` try failed, the loop continues for the second version string. So, we lost the info about the error occurred.

test/integration/uninstall/test_uninstall.py
```diff
@@ -306,3 +306,44 @@ def test_uninstall_tarantool_no_switch(tt_cmd, tmp_path):
...
+
+    # Copying built executable to bin directory,
+    # renaming it to version and change symlink
+    # to emulate that we have installed it like
+    # tt install tt v1.2.3.
+    # We don't really install it because bug we test
+    # is fixed only for current patch. Due to
+    # binary file is changing while running this
+    # bug won't appear only for current and future versions.
+    shutil.copy(tt_cmd, os.path.join(tmp_path, f"tt_{installed_version}"))
```
psergee, line 333: Add a test that if there are two versions are installed: v123 and 123
The passed version only will be removed.

cli/uninstall/uninstall.go
```diff
@@ -112,12 +112,30 @@ func UninstallProgram(program string, programVersion string, binDst string, head
 			return err
 		}
 	}
+	// We need to try to delete all versions with 'v' prefix before x.y.z version
```
psergee, line 115: This function is already big enough. Let's extract the code, which decides which versions to be remove to a separate function.

psergee (commented):
Thank you for the patch, see my comments below.

cli/uninstall/uninstall.go
themilchenko: What if we have `tt uninstall tt <hash>`? The `programVersion` won't be master and also won't contains 'v' prefix, so it will be `v<hash>` and program won't be uninstalled as it was expected. That is why I want to determine if `programVersion` satisfies the `x.y.z`.

oleg-jukovec: Than we need the same for `tt install` too. The idea is keep the logic in sync for install and uninstall.

test/integration/uninstall/test_uninstall.py
themilchenko: How could it be? I thought we can have only one tag version installed using `tt install tt <tag>`: with v tag or without but not both.

cli/uninstall/uninstall.go
themilchenko: Then it won't work for `tt uninstall tt v0.1.0` when file `tt_0.1.0` will exists.

cli/uninstall/uninstall.go
themilchenko: Added.

cli/uninstall/uninstall.go
themilchenko: Added handling an error excluding case when file was not found.

cli/uninstall/uninstall.go
themilchenko: Extracted to it's own function.

cli/util/errors.go
```diff
@@ -5,4 +5,8 @@ import "errors"
 var (
 	// ErrCmdAbort is reported when user aborts the program.
 	ErrCmdAbort = errors.New("aborted by user")
+
+	// ErrNoTtInstalled is reported when program cannot find tt binary
+	// file with name provided.
+	ErrNoTtInstalled = errors.New("there is no tt installed")
```
oleg-jukovec, line 11: This doesn't look like some kind of general-public error so it's better define it on the `uninstall` module level like a private error.

cli/uninstall/uninstall.go
```diff
@@ -113,11 +113,25 @@ func UninstallProgram(program string, programVersion string, binDst string, head
...
+	versionsToDelete := []string{programVersion}
+	if program == search.ProgramTt {
+		versionsToDelete = getAllTtVersionFormats(programVersion)
+	}
+
 	var isSymlinkRemoved bool
-	isSymlinkRemoved, err = remove(program, programVersion, binDst, cmdCtx)
+	for _, verToDel := range versionsToDelete {
+		isSymlinkRemoved, err = remove(program, verToDel, binDst, cmdCtx)
+		if err != nil && err.Error() != util.ErrNoTtInstalled.Error() {
```
oleg-jukovec, line 124:
```suggestion
		if err != nil && !errors.Is(err, ErrNoTtInstalled) {
```

cli/uninstall/uninstall.go
```diff
@@ -133,6 +147,21 @@ func UninstallProgram(program string, programVersion string, binDst string, head
...
 }
 
+// getAllTtVersionFormats returns all version formats with 'v' prefix and
+// without it before x.y.z version.
+func getAllTtVersionFormats(ttVersion string) []string {
+	var versionsToDelete []string
+
+	// Need to determine if we have x.y.z format in tt uninstall argument
+	// to make sure we add version prefix.
+	if strings.Contains(ttVersion, ".") && !strings.HasPrefix(ttVersion, "v") {
```
oleg-jukovec, line 157: The logic still incompatible with the `install`.

`tt install 123` installs 123 if exist or v123 if exist.
`tt uninstall 123` will not uninstall v123.

You need to add the major.minor.patch check with `regexp.Match("^[0-9]+\.[0-9]+\.[0-9]+", ttVersion)` check to the `tt install` in sync with `tt uninstall` or don't add the confusing (`strings.Contains(ttVersion, ".")`) condition here.

cli/util/errors.go
themilchenko: Have deleted it and put in uninstall.go as private error.

cli/uninstall/uninstall.go
themilchenko: `errors.Is` is not working for this case. Fixed to second suggestion.

cli/uninstall/uninstall.go
themilchenko: Added regular expression version check symmetrically for `tt install` and `tt uninstall`.

cli/install/install.go
```diff
@@ -495,7 +499,12 @@ func installTt(binDir string, installCtx InstallCtx, distfiles string) error {
 				return err
 			}
 			for _, ver := range versions {
-				if ttVersion == ver.Str || (ttVersion[:1] != "v" && "v"+ttVersion == ver.Str) {
+				versionMatches, err := regexp.Match(majorMinorPatchRegexp, []byte(ttVersion))
+				if err != nil {
+					return err
+				}
+				if ttVersion == ver.Str || (versionMatches && ttVersion[:1] != "v" &&
```
oleg-jukovec, line 506: The regular expression `^[0-9]+\.[0-9]+\.[0-9]+`
) excludes the case because it only matches `begin_of_a_string+digits.digits.digits`:
```suggestion
				if ttVersion == ver.Str || (versionMatches &&
```

cli/uninstall/uninstall.go
```diff
@@ -133,6 +155,25 @@ func UninstallProgram(program string, programVersion string, binDst string, head
...
+func getAllTtVersionFormats(ttVersion string) ([]string, error) {
+	var versionsToDelete []string
+
+	// Need to determine if we have x.y.z format in tt uninstall argument
+	// to make sure we add version prefix.
+	versionMatches, err := regexp.Match(majorMinorPatchRegexp, []byte(ttVersion))
+	if err != nil {
+		return versionsToDelete, err
+	}
+	if versionMatches && !strings.HasPrefix(ttVersion, "v") {
```
oleg-jukovec, line 169: The same:
```suggestion
	if versionMatches {
```

cli/uninstall/uninstall_test.go
```diff
@@ -143,3 +143,35 @@ func TestSearchLatestVersion(t *testing.T) {
...
+		ttVersion        string
+		expectedVersions []string
+		expectedError    error
+	}
+
+	cases := []testCase{
+		{
+			name:             "without prefix",
+			ttVersion:        "1.2.3",
+			expectedVersions: []string{"v1.2.3", "1.2.3"},
```
oleg-jukovec, line 159: There should be reverse priority here because we want to prefer to delete a specified version:
```suggestion
			expectedVersions: []string{"1.2.3", "v1.2.3"},
```

cli/uninstall/uninstall.go
oleg-jukovec: > `errors.Is` is not working for this case

It seems like a problem becase `remove(` does not return the error variable as is. So it could be changed independently. It is better to return ErrNoTtInstalled from the `remove(` call (or from a nested call).

oleg-jukovec: It seems like a problem because the `remove(` call does not return the error variable as is. This is not obvious and makes the code difficult to maintain. It is better to return `ErrNoTtInstalled` from the `remove(` call (or from a nested call).

themilchenko: Now `remove(` is returning defined `errNoTtInstalled` when program name is "tt". It is now possible to use `errors.Is`.

cli/install/install.go
themilchenko: Deleted redundant expression.

cli/uninstall/uninstall_test.go
themilchenko: Changed to this behaviour.

oleg-jukovec (approved):
Thank you for the fix!

cli/install/install.go
```diff
@@ -495,7 +499,11 @@ func installTt(binDir string, installCtx InstallCtx, distfiles string) error {
 				return err
 			}
 			for _, ver := range versions {
-				if ttVersion == ver.Str || (ttVersion[:1] != "v" && "v"+ttVersion == ver.Str) {
+				versionMatches, err := regexp.Match(majorMinorPatchRegexp, []byte(ttVersion))
+				if err != nil {
+					return err
+				}
+				if ttVersion == ver.Str || (versionMatches && "v"+ttVersion == ver.Str) {
```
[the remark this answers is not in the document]
themilchenko: Then we won't have comparison in this expresion for case when the version will be without 'v' prefix.

cli/uninstall/uninstall.go
```diff
@@ -23,7 +23,12 @@ const (
 		search.ProgramTt + "|" +
 		search.ProgramCe + "|" +
 		search.ProgramEe + ")"
-	verRegexp = "(?P<ver>.*)"
+	verRegexp             = "(?P<ver>.*)"
+	majorMinorPatchRegexp = `^[0-9]+\.[0-9]+\.[0-9]+`
```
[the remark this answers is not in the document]
themilchenko: Deleted duplicate.

cli/install/install.go
```diff
@@ -495,7 +499,11 @@ func installTt(binDir string, installCtx InstallCtx, distfiles string) error {
 				return err
 			}
 			for _, ver := range versions {
-				if ttVersion == ver.Str || (ttVersion[:1] != "v" && "v"+ttVersion == ver.Str) {
+				versionMatches, err := regexp.Match(majorMinorPatchRegexp, []byte(ttVersion))
```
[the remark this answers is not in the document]
themilchenko: Took it out of the cycle.

psergee (approved):
lgtm
just minor comments left.

cli/install/install.go
```diff
@@ -68,11 +68,15 @@ var unwindPatch []byte
...
-// read,write for user and only read for others.
-const defaultDirPermissions = 0755
+const (
+	// defaultDirPermissions is rights used to create folders.
+	// 0755 - drwxr-xr-x
+	// We need to give permission for all to execute
+	// read,write for user and only read for others.
+	defaultDirPermissions = 0755
+
+	MajorMinorPatchRegexp = `^[0-9]+\.[0-9]+\.[0-9]+`
```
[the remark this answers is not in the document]
oleg-jukovec: `-alpha`, `-beta`, a commit hash and what we would like there in the future.

cli/uninstall/uninstall.go
```diff
@@ -113,11 +120,28 @@ func UninstallProgram(program string, programVersion string, binDst string, head
 		}
 	}
 
+	versionsToDelete := []string{programVersion}
+	if program == search.ProgramTt {
```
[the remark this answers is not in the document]
themilchenko: I thought that it is simpler to read: we go to the `getAllTtVersionFormats` block only if current program is `tt`.

cli/uninstall/uninstall.go
```diff
@@ -53,6 +57,9 @@ func remove(program string, programVersion string, directory string,
 	path := filepath.Join(directory, fileName)
 
 	if _, err := os.Stat(path); os.IsNotExist(err) {
+		if program == search.ProgramTt {
+			return false, errTtNotInstalled
```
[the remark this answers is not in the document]
themilchenko: Added separate error for each program.

cli/uninstall/uninstall.go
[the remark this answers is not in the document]
themilchenko: Added this logic into function.

cli/uninstall/uninstall_test.go
```diff
@@ -143,3 +143,35 @@ func TestSearchLatestVersion(t *testing.T) {
...
+
+func TestGetAllVersionFormats(t *testing.T) {
+	type testCase struct {
+		name             string
+		ttVersion        string
+		expectedVersions []string
+		expectedError    error
+	}
+
+	cases := []testCase{
```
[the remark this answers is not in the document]
themilchenko: Added.

test/integration/uninstall/test_uninstall.py
```diff
@@ -306,3 +306,52 @@ def test_uninstall_tarantool_no_switch(tt_cmd, tmp_path):
         os.path.join(testdata_path, "bin", "tarantool_2.10.4"),
         os.path.join(testdata_path, "inc", "include", "tarantool_2.10.4")
     )
+
+
+@pytest.mark.parametrize("installed_versions, version_to_uninstall", [
+    (["v1.2.3"], "1.2.3"),
+    (["v1.2.3", "1.2.3"], "1.2.3")
```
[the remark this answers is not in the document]
themilchenko: Added.

cli/install/install.go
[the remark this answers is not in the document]
psergee: > `-alpha`, `-beta`, a commit hash and what we would like there in the future.

If so, why don't we just rely of version.Parse. If it is passed, user provided version is a version, not a commit or PR num,

cli/uninstall/uninstall.go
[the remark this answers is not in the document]
psergee: omg, please no... Just use one error for all programs.

cli/install/install.go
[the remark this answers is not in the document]
oleg-jukovec: I don't see any problem using the function if we have it in the `cli/version`. To be honest, I didn’t know about it.

PR #901, opened 2024-07-12
`tt start` interactive mode

psergee:
This patch adds support for `tt start` interactive mode with
    `-i/--interactive` flag. In this mode `tt` will wait for
    child Tarantool process completions. Standard output will be
    used instead of the default log files. Use Ctrl-C combination
    to stop `tt` and its child Tarantool instances. No watchdog
    processes created.

### DIFF
--- a/cli/ttlog/logger.go
+++ b/cli/ttlog/logger.go
@@ -1,10 +1,17 @@
 package ttlog
 
 import (
+	"errors"
+	"fmt"
 	"io"
 	"log"
+	"os"
+	"path/filepath"
+)
 
-	"gopkg.in/natefinch/lumberjack.v2"
+const (
+	logOpenFlags   = os.O_CREATE | os.O_WRONLY | os.O_APPEND
+	logCreatePerms = 0640
 )
 
 // LoggerOpts describes the logger options.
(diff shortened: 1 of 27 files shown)


better0fdead (commented):
Hi, thx for patch. See comments below

cli/ttlog/logger.go
```diff
@@ -15,68 +22,113 @@ type LoggerOpts struct {
...
+	Close() error
+}
+
+// writerLogger is a thin log.Logger wrapper providing the Logger interface.
+type writerLogger struct {
+	*log.Logger
+}
+
+// fileLogger is a logger that writes log messages to the file.
+type fileLogger struct {
```
DerekBum, line 47: Maybe it will be better to restructure this file. So we would first we have `writerLogger` type and all its methods. And only then `fileLogger` and all its methods. Right now everything is mixed up.

I think this is a nit, up to you.

DerekBum (approved):
Lgtm, just one nit.

cli/ttlog/logger.go
psergee: Done.

PR #904, opened 2024-07-23
cli: add integrity-check-period flag in global context

themilchenko:
It was impossible to provide integrity check period in `tt restart`. After the patch flag "--integrity-check-period" is locating in the global cli context.

Part of https://github.com/tarantool/tt-ee/issues/203

### DIFF
--- a/cli/cmd/start.go
+++ b/cli/cmd/start.go
@@ -8,7 +8,6 @@ import (
 	"strconv"
 	"sync"
 	"syscall"
-	"time"
 
 	"github.com/spf13/cobra"
 	"github.com/tarantool/tt/cli/cmd/internal"
@@ -25,7 +24,7 @@ var (
 	// In go, we can't just fork the process (reason - goroutines).
 	// So, for daemonize, we restarts the process with "watchdog" flag.
 	watchdog bool
-	// integrityCheckPeriod is a flag enables periodic integrity checks.
+	// integrityCheckPeriod is a flag that enables periodic integrity checks.
 	// The default period is 1 day.
 	integrityCheckPeriod = 24 * 60 * 60
 	// startInteractive is startInteractive mode flag. If set, the main process does not exit after
@@ -60,7 +59,7 @@ func NewStartCmd() *cobra.Command {
 	startCmd.Flags().MarkHidden("watchdog")
 	startCmd.Flags().BoolVarP(&startInteractive, "interactive", "i", false, "")
 
-	integrity.RegisterIntegrityCheckPeriodFlag(startCmd.Flags(), &integrityCheckPeriod)
+	integrity.RegisterIntegrityCheckPeriodFlag(startCmd.Flags(), &cmdCtx.Cli.IntegrityCheckPeriod)
 
 	return startCmd
 }
@@ -75,7 +74,7 @@ func startInstancesUnderWatchdog(cmdCtx *cmdcontext.CmdCtx, instances []running.
 	startArgs := []string{}
 	if cmdCtx.Cli.IntegrityCheck != "" {
 		startArgs = append(startArgs, "--integrity-check-period",
-			strconv.Itoa(integrityCheckPeriod))
+			strconv.FormatUint(uint64(cmdCtx.Cli.IntegrityCheckPeriod), 10))
 	}
 
 	for _, instance := range instances {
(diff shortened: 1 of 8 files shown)


cli/cmd/start.go
```diff
@@ -105,8 +99,8 @@ func internalStartModule(cmdCtx *cmdcontext.CmdCtx, args []string) error {
 
 	checkPeriod := time.Duration(0)
 
-	if cmdCtx.Cli.IntegrityCheck != "" && integrityCheckPeriod > 0 {
-		checkPeriod = time.Duration(integrityCheckPeriod * int(time.Second))
+	if cmdCtx.Cli.IntegrityCheck != "" && cmdCtx.Cli.IntegrityCheckPeriod > 0 {
+		checkPeriod = time.Duration(cmdCtx.Cli.IntegrityCheckPeriod * int(time.Second))
 	}
 
 	if err := running.Start(cmdCtx, &runningCtx.Instances[0], checkPeriod); err != nil {
```
oleg-jukovec, line 106: At now for each commend we will need to implement the global flag pass into `Start()`. It's better to use the global flag here:

https://github.com/tarantool/tt/blob/7a920f35d986536ddde6c45f62391d97b95579f6/cli/running/running.go#L721-L723

directly. It will allow us not to change a code of `restart` as example.

oleg-jukovec (changes requested):
1. We need to save the `--integrity-check-period` flag for the `tt start` command for the backward compatibility.
2. A global-level flag usage should be moved somewhere to the `running` package to avoid to update each command that will use `running.Start()`.

cli/cmd/start.go
themilchenko: Do you mean to use global flag directly passing context into `running.Start()` and use `cmdCtx.Cli.IntegrityCheckPeriod` directly in this function instead of `integrityCheckPeriod` variable?

themilchenko: Yes, I understood, fixed it.

lib/integrity/integrity.go
```diff
@@ -10,6 +10,9 @@ var (
 	// ErrNotConfigured is reported when integrity check is not configured
 	// in the command context.
 	ErrNotConfigured = errors.New("integrity check is not configured")
+	// integrityCheckPeriod is a flag enables periodic integrity checks.
+	// The default period is 1 day.
+	integrityCheckPeriod = 24 * 60 * 60
```
oleg-jukovec, line 15:
```suggestion
	// integrityCheckPeriod is a flag enables periodic integrity checks.
	// The default period is 1 day.
	integrityCheckPeriod = 24 * 60 * 60
```
The default value is not used anywhere.

cli/cmd/start.go
```diff
@@ -50,8 +46,6 @@ func NewStartCmd() *cobra.Command {
 	startCmd.Flags().BoolVar(&watchdog, "watchdog", false, "")
 	startCmd.Flags().MarkHidden("watchdog")
 
-	integrity.RegisterIntegrityCheckPeriodFlag(startCmd.Flags(), &integrityCheckPeriod)
-
```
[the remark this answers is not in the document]
themilchenko: Created [issue](https://github.com/tarantool/tt-ee/issues/214) for this problem.

lib/integrity/integrity.go
themilchenko: Deleted it.

oleg-jukovec: But we need the default value :(

cli/cmd/start.go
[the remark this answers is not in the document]
oleg-jukovec: Thank you!

lib/integrity/integrity.go
themilchenko: The default value will be set in `tt-ee` in implementation function of creating this flag. In this edition this flag cannot be used, it is initing by nil value which is 0.

oleg-jukovec: Okay, I'll wait then for a patch for `tt-ee`.

oleg-jukovec: To be honest, I don't see the test here )

oleg-jukovec (approved):
Thank you for the patch!

oleg-jukovec:
Please, rebase the pull request.

PR #908, opened 2024-07-29
ci: add step with killing tarantool after tests

themilchenko:
There were still tarantool processes running after all tests were completed. After the patch all not killed instances of tarantool will be killed in CI pipeline.


psergee (approved):
lgtm

oleg-jukovec:
What is the reason that a tarantool still running?
Please, create an issue to find a test that do not stop a tarantool instance.

PR #909, opened 2024-07-30
install: fix tests failure in master branch

themilchenko:
It was impossible to test detection of latest version in master branch due to the same commit hash.

After the patch tests create executable file with emulation of work 'tt verison --commit' to avoid this problem.


psergee (approved):
Thank you for the patch, LGTM.

PR #927, opened 2024-08-20
tt status: print detailed alerts for instances

### LINKED ISSUE #836: `tt status`: give more status information
`tt status` prints an instance in the green color in the following situations:

* `config:info()` reports `startup_in_progress` status, because a replicaset bootstrap process waits for a quorum of instances.
* `config:info()` reports errors after a reload.
* `box.info` reports the `orphan` status.
* All the instances are read-only due to a mistake in a login or a password in the replication configuration (`box.info.replication` reports errors) -- see also #835.

I propose to print such instances in a yellow color and give all the collected warnings/errors.

I think that a preferred way to collect the status information is a control socket (console), but it is OK to use iproto as a backup way (for example, if the instance is on a remote machine).

mandesero:
This patch improves the `tt status` command by displaying all collected errors, warnings, and the current instance status. Instances with warnings or errors will now be highlighted in yellow.

Closes #836

### DIFF
--- a/cli/status/lua/instance_state.lua
+++ b/cli/status/lua/instance_state.lua
@@ -1,0 +1,76 @@
+local ok, config = pcall(require, 'config')
+
+-- Default configuration in case the Tarantool 3.x config module is unavailable.
+local config_info_full = {
+    status = "uninitialized",
+    alerts = {},
+}
+
+-- The type check for 'config.info' ensures that we are working with the correct
+-- Tarantool 3.x config module. This is necessary because another module,
+-- 'moonlibs/config', shares the same name.
+if ok and config ~= nil and type(config) == "table" and type(config.info) == "function" then
+    config_info_full = config:info()
+end
+
+local config_info = {
+    status = config_info_full.status,
+    alerts = {}
+}
+
+-- This loop extracts only the 'type' and 'message' fields from each alert,
+-- discarding any additional fields that might be present in the config.alerts.
+-- We only need the type of the error and its message for the status report,
+-- as other fields are considered extra and unnecessary for our purposes.
+for _, alert in ipairs(config_info_full.alerts) do
+    table.insert(config_info.alerts, {
+        type = alert.type,
+        message = alert.message
+    })
+end
+
+local ok, box_replication = pcall(function() return box.info.replication end)
+if not ok or box_replication == nil then
+    box_replication = {}
+end
+
+-- We collect information only for instances that have an upstream,
+-- as this field is necessary for subsequent error reporting
+local replication_info = {}
+for _, instance in pairs(box_replication) do
+    if instance.upstream ~= nil then
+        table.insert(replication_info, {
+            uuid = instance.uuid,
+            name = instance.name,
+            upstream = instance.upstream and {
+                status = instance.upstream.status or "--",
+                message = instance.upstream.message
+            }
+        })
+    end
+end
+
+local ok, read_only = pcall(function() return box.info.ro end)
+if ok then
+    read_only = read_only and "RO" or "RW"
+else
+    read_only = "RO"
+end
+
+local ok, box_status = pcall(function() return box.info.status end)
+if not ok then
+    box_status = "N/A"
+end
+
+local ok, uuid = pcall(function() return box.info.uuid end)
+if not ok then
+    uuid = "00000000-0000-0000-0000-000000000000"
+end
+
+return {
+    replication_info = replication_info,
+    config_info = config_info,
+    read_only = read_only,
+    box_status = box_status,
+    uuid = uuid,
+}
--- a/cli/status/status.go
+++ b/cli/status/status.go
@@ -1,58 +1,248 @@
 package status
 
 import (
+	_ "embed"
+	"fmt"
 	"os"
+	"strings"
 
+	"github.com/fatih/color"
 	"github.com/jedib0t/go-pretty/v6/table"
 	"github.com/jedib0t/go-pretty/v6/text"
+	"github.com/mitchellh/mapstructure"
 	"github.com/tarantool/tt/cli/connector"
 	"github.com/tarantool/tt/cli/process_utils"
 	"github.com/tarantool/tt/cli/running"
 )
 
+//go:embed lua/instance_state.lua
+var instanceInfoLuaScript string
+
+var defaultModuleStatus = "--"
+
+func filterComments(script string) string {
+	var filteredLines []string
+	lines := strings.Split(script, "\n")
+	for _, line := range lines {
+		trimmedLine := strings.TrimSpace(line)
+		if !strings.HasPrefix(trimmedLine, "--") {
+			filteredLines = append(filteredLines, line)
+		}
+	}
+	return strings.Join(filteredLines, "\n")
+}
+
 // StatusOpts contains options for tt status.
 type StatusOpts struct {
 	// Option for pretty-formatted table output.
 	Pretty bool
+	// Option for detailed alerts output for each instance, such as warnings and errors.
+	Details bool
+}
+
+type alert struct {
+	Type    string `mapstructure:"type"`
+	Message string `mapstructure:"message"`
+}
+
+type configInfo struct {
+	Status string  `mapstructure:"status"`
+	Alerts []alert `mapstructure:"alerts"`
+}
+
+type upstream struct {
+	Status  string `mapstructure:"status"`
+	Message string `mapstructure:"message"`
+}
+
+type replicationInfo struct {
+	UUID     string   `mapstructure:"uuid"`
+	Name     *string  `mapstructure:"name"`
+	Upstream upstream `mapstructure:"upstream"`
+}
+
+type instanceState struct {
+	ReplicationInfo []replicationInfo `mapstructure:"replication_info"`
+	ConfigInfo      configInfo        `mapstructure:"config_info"`
+	ReadOnly        string            `mapstructure:"read_only"`
+	BoxStatus       string            `mapstructure:"box_status"`
+	UUID            string            `mapstructure:"uuid"`
+}
+
+func newInstanceStatusMap() map[string]interface{} {
+	return map[string]interface{}{
+		"STATUS":   "",
+		"PID":      nil,
+		"MODE":     "",
+		"CONFIG":   defaultModuleStatus,
+		"BOX":      defaultModuleStatus,
+		"UPSTREAM": defaultModuleStatus,
+	}
+}
+
+var instances = map[string]map[string]interface{}{}
+var instancesAlerts = map[string][]string{}
+var uuid2name = map[string]string{}
+
+var printYellow = color.New(color.FgYellow).SprintFunc()
+var printRed = color.New(color.FgRed).SprintFunc()
+
+func processReplicationInfo(fullInstanceName string, instanceState instanceState) {
+	for _, repl := range instanceState.ReplicationInfo {
+		fullInstanceUpstreamName, ok := uuid2name[repl.UUID]
+		// Use repl.Name if available, otherwise fallback to repl.UUID
+		if !ok {
+			if repl.Name != nil {
+				fullInstanceUpstreamName = *repl.Name
+			} else {
+				fullInstanceUpstreamName = repl.UUID
+			}
+		}
+		if repl.Upstream.Status == "follow" || len(repl.Upstream.Message) == 0 {
+			continue
+		}
+		instances[fullInstanceName]["UPSTREAM"] = repl.Upstream.Status
+
+		var upstreamInstanceDesc string
+		if ok || repl.Name != nil {
+			upstreamInstanceDesc = fmt.Sprintf("instance with name %q",
+				fullInstanceUpstreamName)
+		} else {
+			upstreamInstanceDesc = fmt.Sprintf("instance with UUID %s",
+				fullInstanceUpstreamName)
+		}
+		instancesAlerts[fullInstanceName] = append(instancesAlerts[fullInstanceName],
+			printYellow(fmt.Sprintf(
+				"[upstream][warning]: replication from %s is in %q status: %q",
+				upstreamInstanceDesc, repl.Upstream.Status,
+				repl.Upstream.Message)))
+	}
+}
+
+func processConfigInfo(fullInstanceName string, instanceState instanceState) {
+	if len(instanceState.ConfigInfo.Alerts) == 0 {
+		return
+	}
+	for _, alert := range instanceState.ConfigInfo.Alerts {
+		msg := ""
+		if alert.Type == "error" {
+			msg = printRed(fmt.Sprintf("[config][error]: %s", alert.Message))
+		} else {
+			msg = printYellow(fmt.Sprintf("[config][warning]: %s", alert.Message))
+		}
+		instancesAlerts[fullInstanceName] = append(instancesAlerts[fullInstanceName], msg)
+	}
 }
 
 // Status writes the status as a table.
 func Status(runningCtx running.RunningCtx, opts StatusOpts) error {
 	ts := table.NewWriter()
 	ts.SetOutputMirror(os.Stdout)
-	ts.AppendHeader(table.Row{"INSTANCE", "STATUS", "PID", "MODE"})
+	ts.AppendHeader(
+		table.Row{"INSTANCE", "STATUS", "PID", "MODE", "CONFIG", "BOX", "UPSTREAM"})
+
+	instanceRawState := map[string]instanceState{}
 
 	for _, run := range runningCtx.Instances {
-		row := []interface{}{}
 		fullInstanceName := running.GetAppInstanceName(run)
 		procStatus := running.Status(&run)
+		instances[fullInstanceName] = newInstanceStatusMap()
+		instances[fullInstanceName]["STATUS"] = procStatus.ColorSprint(procStatus.Status)
 
-		row = append(row, fullInstanceName)
-		row = append(row, procStatus.ColorSprint(procStatus.Status))
 		if procStatus.Code == process_utils.ProcessRunningCode {
-			row = append(row, procStatus.PID)
+			instances[fullInstanceName]["PID"] = procStatus.PID
 		}
 
 		conn, err := connector.Connect(connector.ConnectOpts{
 			Network: "unix",
 			Address: run.ConsoleSocket,
 		})
-		if err == nil {
-			res, err := conn.Eval("return (type(box.cfg) == 'function') or box.info.ro",
-				[]any{}, connector.RequestOpts{})
-			if err == nil && len(res) != 0 {
-				mode := res[0].(bool)
-				mode_str := "RO"
-				if !mode {
-					mode_str = "RW"
-				}
-				row = append(row, mode_str)
-			}
+
+		if err != nil && procStatus.Code == process_utils.ProcessRunningCode {
+			instancesAlerts[fullInstanceName] = append(
+				instancesAlerts[fullInstanceName],
+				printRed(fmt.Sprintf(
+					"Error while connecting to instance %s via socket %s: %v",
+					fullInstanceName, run.ConsoleSocket, err)))
+			continue
+		} else if err != nil {
+			continue
 		}
+
+		var instanceState instanceState
+		res, err := conn.Eval(filterComments(instanceInfoLuaScript), []any{},
+			connector.RequestOpts{})
+
+		if err != nil {
+			instancesAlerts[fullInstanceName] = append(
+				instancesAlerts[fullInstanceName],
+				printRed(fmt.Sprintf(
+					"Error while executing Lua script on instance %s: %v",
+					fullInstanceName, err)))
+			continue
+		}
+
+		if len(res) == 0 {
+			instancesAlerts[fullInstanceName] = append(
+				instancesAlerts[fullInstanceName],
+				printRed(fmt.Sprintf(
+					"No data returned from Lua script on instance %s",
+					fullInstanceName)))
+			continue
+		}
+
+		err = mapstructure.Decode(res[0], &instanceState)
+		if err != nil {
+			instancesAlerts[fullInstanceName] = append(
+				instancesAlerts[fullInstanceName],
+				printRed(fmt.Sprintf("Error while decoding data from "+
+					"instance %s: %v", fullInstanceName, err)))
+			continue
+		}
+
+		// Since Tarantool 2.x doesn't support instance names, only UUIDs are available.
+		// To make the alerts more readable, we map the UUIDs to instance names.
+		uuid2name[instanceState.UUID] = fullInstanceName
+
+		instanceRawState[fullInstanceName] = instanceState
+		instances[fullInstanceName]["MODE"] = instanceState.ReadOnly
+		instances[fullInstanceName]["CONFIG"] = instanceState.ConfigInfo.Status
+		instances[fullInstanceName]["BOX"] = instanceState.BoxStatus
+	}
+
+	// Alert handling placed later because we need to know the mapping of instance UUIDs
+	// to their names for a more user-friendly output.
+	for fullInstanceName, instanceState := range instanceRawState {
+		processReplicationInfo(fullInstanceName, instanceState)
+		processConfigInfo(fullInstanceName, instanceState)
+	}
+
+	for instName, instData := range instances {
+		row := []interface{}{}
+		row = append(row, instName)
+		row = append(row, instData["STATUS"])
+		if instData["PID"] == nil {
+			ts.AppendRow(row)
+			continue
+		}
+		row = append(row, instData["PID"])
+		row = append(row, instData["MODE"])
+		row = append(row, instData["CONFIG"])
+		row = append(row, instData["BOX"])
+		row = append(row, instData["UPSTREAM"])
 		ts.AppendRow(row)
 	}
 	ts.SortBy([]table.SortBy{{Name: "INSTANCE", Mode: table.Asc}})
 
+	if opts.Details {
+		for instance, alerts := range instancesAlerts {
+			fmt.Printf("Alerts for %s:\n", instance)
+			for _, alert := range alerts {
+				fmt.Printf("  • %s\n", alert)
+			}
+			fmt.Println()
+		}
+	}
 	if opts.Pretty {
 		ts.SetStyle(table.StyleRounded)
 	} else {
@@ -67,5 +257,18 @@ func Status(runningCtx running.RunningCtx, opts StatusOpts) error {
 		{Number: 4, Align: text.AlignLeft, AlignHeader: text.AlignLeft},
 	})
 	ts.Render()
+
+	if !opts.Details {
+		msg := "\nThe status of some instances requires attention.\n" +
+			"Please rerun the command with the --details flag to see " +
+			"more information"
+		for _, alerts := range instancesAlerts {
+			if len(alerts) > 0 {
+				fmt.Println(msg)
+				break
+			}
+		}
+	}
+
 	return nil
 }
--- a/test/integration/status/test_status.py
+++ b/test/integration/status/test_status.py
@@ -1,0 +1,325 @@
+import os
+import re
+import shutil
+import subprocess
+import tempfile
+import time
+
+import pytest
+import tarantool
+
+from utils import (control_socket, extract_status, get_tarantool_version,
+                   pid_file, run_command_and_get_output, run_path, wait_file)
+
+tarantool_major_version, tarantool_minor_version = get_tarantool_version()
+
+
+def start_application(cmd, workdir, app_name, instances):
+    instance_process = subprocess.Popen(
+        cmd,
+        cwd=workdir,
+        stderr=subprocess.STDOUT,
+        stdout=subprocess.PIPE,
+        text=True
+    )
+    start_output = instance_process.stdout.read()
+    for inst in instances:
+        assert f"Starting an instance [{app_name}:{inst}]" in start_output
+
+
+def stop_application(tt_cmd, app_name, workdir):
+    stop_cmd = [tt_cmd, "stop", app_name, "-y"]
+    stop_rc, stop_out = run_command_and_get_output(stop_cmd, cwd=workdir)
+    assert stop_rc == 0
+
+
+def break_config(path):
+    with open(path, "a") as config:
+        config.write("invalid_field: invalid_value\n")
+
+
+def run_command_on_instance(tt_cmd, tmpdir, full_inst_name, cmd):
+    con_cmd = [tt_cmd, "connect", full_inst_name, "-f", "-"]
+    instance_process = subprocess.Popen(
+        con_cmd,
+        cwd=tmpdir,
+        stderr=subprocess.STDOUT,
+        stdout=subprocess.PIPE,
+        stdin=subprocess.PIPE,
+        text=True
+    )
+    instance_process.stdin.writelines([cmd])
+    instance_process.stdin.close()
+    output = instance_process.stdout.read()
+    return output
+
+
+def wait_instance_status(tt_cmd, tmpdir, full_inst_name, status, port=None, timeout=10):
+    if status == "config":
+        cmd = "return require('config'):info().status"
+        expected_statuses = ["ready", "check_warnings"]
+    elif status == "box":
+        cmd = "return box.info.status"
+        expected_statuses = ["running"]
+    else:
+        raise RuntimeError(f"Not supported status to check: {status}")
+
+    conn = None
+    end_time = time.time() + timeout
+    while True:
+        try:
+            if port:
+                # if socket file doesn't exist use connection by ip:port
+                if not conn:
+                    conn = tarantool.Connection(host="localhost", port=port)
+                res = conn.eval(cmd)[0]
+            else:
+                res = run_command_on_instance(
+                    tt_cmd,
+                    tmpdir,
+                    full_inst_name,
+                    cmd
+                )
+
+            if any(expected_status in res for expected_status in expected_statuses):
+                if conn:
+                    conn.close()
+                return True
+        except tarantool.error.Error:
+            pass
+        if time.time() > end_time:
+            print(f"[{full_inst_name}]: {status} wait timed out after {timeout} seconds.")
+            return False
+        time.sleep(1)
+
+
+@pytest.mark.skipif(tarantool_major_version < 3,
+                    reason="skip cluster instances test for Tarantool < 3")
+def test_t3_instance_names_with_config(tt_cmd, tmpdir_with_cfg):
+    tmpdir = tmpdir_with_cfg
+    app_name = "small_cluster_app"
+    app_path = os.path.join(tmpdir, app_name)
+    shutil.copytree(os.path.join(os.path.dirname(__file__), f"../running/{app_name}"), app_path)
+
+    run_dir = os.path.join(tmpdir, app_name, run_path)
+    instances = ['storage-master', 'storage-replica']
+    try:
+        # Start an instance.
+        start_cmd = [tt_cmd, "start", app_name]
+        start_application(start_cmd, tmpdir, app_name, instances)
+
+        # Check status.
+        for inst in instances:
+            file = wait_file(os.path.join(run_dir, inst), 'tarantool.pid', [])
+            assert file != ""
+            file = wait_file(os.path.join(run_dir, inst), control_socket, [])
+            assert file != ""
+
+        status_cmd = [tt_cmd, "status", app_name]
+        status_rc, status_out = run_command_and_get_output(status_cmd, cwd=tmpdir)
+        assert status_rc == 0
+        status_info = extract_status(status_out)
+        for inst in instances:
+            assert status_info[app_name+":"+inst]["STATUS"] == "RUNNING"
+            assert os.path.exists(os.path.join(tmpdir, app_name, "var", "lib", inst))
+            assert os.path.exists(os.path.join(tmpdir, app_name, "var", "log", inst, "tt.log"))
+            assert not os.path.exists(os.path.join(tmpdir, app_name, "var", "log", inst,
+                                                   "tarantool.log"))
+
+        full_master_inst_name = f"{app_name}:{instances[0]}"
+
+        # Wait for the configuration setup to complete
+        assert wait_instance_status(tt_cmd, tmpdir, full_master_inst_name, "config")
+
+        # Break the configuration by modifying the config.yaml file
+        config_path = os.path.join(app_path, "config.yaml")
+        break_config(config_path)
+
+        # Reload the configuration on the instance
+        reload_cmd = "require('config'):reload()"
+        res = run_command_on_instance(tt_cmd, tmpdir, full_master_inst_name, reload_cmd)
+
+        # Check if the expected error message is present in the response
+        error_message = "[cluster_config] Unexpected field \"invalid_field\""
+        assert error_message in res
+
+        status_cmd = [tt_cmd, "status", full_master_inst_name, "--details"]
+        status_rc, status_out = run_command_and_get_output(status_cmd, cwd=tmpdir)
+        assert status_rc == 0
+
+        status_table = status_out[status_out.find("INSTANCE"):]
+        status_info = extract_status(status_table)
+
+        assert status_info[full_master_inst_name]["STATUS"] == "RUNNING"
+        assert status_info[full_master_inst_name]["MODE"] == "RW"
+        assert status_info[full_master_inst_name]["CONFIG"] == "check_errors"
+        assert status_info[full_master_inst_name]["BOX"] == "running"
+
+        # We cannot be certain that the instance bootstrap has completed.
+        assert status_info[full_master_inst_name]["UPSTREAM"] in ["--", "loading"]
+        assert f"[config][error]: {error_message}" in status_out
+    finally:
+        stop_application(tt_cmd, app_name, tmpdir)
+
+
+@pytest.mark.skipif(tarantool_major_version < 3,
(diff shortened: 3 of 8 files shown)


oleg-jukovec:
Please provide examples of how it looks like.

cli/status/status.go
```diff
@@ -20,35 +68,110 @@ type StatusOpts struct {
...
+				[]any{}, connector.RequestOpts{})
+
+			if err == nil && len(res) != 0 {
+				configInfo := res[0].(map[string]interface{})
+
+				status := configInfo["status"]
+				if status == "startup_in_progress" {
+					warnings = append(warnings,
+						fmt.Sprintf("[Status]: %s", status))
+				}
```
Totktonada, line 96: I would notify a user about any status except `ready`.

cli/status/status.go
```diff
@@ -20,35 +68,110 @@ type StatusOpts struct {
...
+		warnings := []string{}
+		mode_str := ""
 
 		conn, err := connector.Connect(connector.ConnectOpts{
 			Network: "unix",
 			Address: run.ConsoleSocket,
 		})
 		if err == nil {
-			res, err := conn.Eval("return (type(box.cfg) == 'function') or box.info.ro",
+			res, err := conn.Eval("return require('config'):info()",
```
Totktonada, line 86: We're doing 4 network requests plus one for each upstream. It means at least 3 round-trip-time waiting for each of the instances.

We can coalesce several requests into one:

```lua
eval("return <..expr X..>")
eval("return <..expr X..>")
eval("return <..expr X..>")

=> 

eval("do
    local x = <..expr X..>
    local y = <..expr Y..>
    local z = <..expr Z..>
    return x, y, z
end")
```

In fact, all the expressions are taking data from the `box.info` result. I guess we can call it once and parse on the client then.

I see that the RO/RW mode request has `type(box.cfg) == 'function'` check. I guess it is an attempt to handle a situation, when `box.cfg()` is not called yet. If so, we should keep this check. It should be verified on tarantool that has no `box.info` before first `box.cfg`. See https://github.com/tarantool/tarantool/issues/7255.

cli/status/status.go
```diff
@@ -20,35 +68,110 @@ type StatusOpts struct {
...
 		fullInstanceName := running.GetAppInstanceName(run)
 		procStatus := running.Status(&run)
-
 		row = append(row, fullInstanceName)
-		row = append(row, procStatus.ColorSprint(procStatus.Status))
-		if procStatus.Code == process_utils.ProcessRunningCode {
-			row = append(row, procStatus.PID)
-		}
+		warnings := []string{}
+		mode_str := ""
```
Totktonada, line 79: I understand it as follows. We need these variables to save results to add them later as columns in some specified order. But I don't see why not to produce the columns in the order how they should appear in the row.

I see that we use network requests only if the network connection succeeds, i.e. under the `if err == nil` branch. However, we can save some `isConnected` property and use it where appropriate. Maybe we can even use just `conn != nil` for it.

This way we can produce the needed values and add them into corresponding columns right after this. It should make the code easier to follow.

oleg-jukovec:
Perhaps it's better to add an additional flag for this feature (-v/--verbose)? It looks like the output is now becoming less readable, and at the same time much more difficult to parse in scripts.

oleg-jukovec:
image
status/status for the same instance looks confusing. I think the second one needs to be clarified.

oleg-jukovec:
> Perhaps it's better to add an additional flag for this feature (-v/--verbose)? It looks like the output is now becoming less readable, and at the same time much more difficult to parse it in scripts.

An another idea: print alerts with the verbose flag, but the instance status by default. Something like this:
```
INSTANCE STATE        PID STATUS
foo              RUNNING  123 orphan
```

Totktonada:
I would discuss the output format in context of a human reader, because a machine-readable output is easier to implement separately (with some `--json` or `--output json` CLI option).

I think that if something is suspectful on an instance, we should let a user know about it by default. OTOH, we're not obligated to print anything if everything is OK. We can hide the column, for example. Or we can print these warnings on separate lines under the instance name if any.

The key idea is the following:

* Don't flood with extra output if everything is OK.
* Let a user know if something potentially goes wrong with the instances (by default, wthout extra flags).

oleg-jukovec:
> I would discuss the output format in context of a human reader, because a machine-readable output is easier to implement separately (with some `--json` or `--output json` CLI option).

Now the output can be easily parsed using regular shell utilities. We can make things complicated here, or we can just keep it simple.

> I think that if something is suspectful on an instance, we should let a user know about it by default. OTOH, we're not obligated to print anything if everything is OK.

An instance state The instance status looks interesting in most cases from my point of view. Even if it is `running`. The column appearing and disappearing time to time looks strange here.

Totktonada:
> > I would discuss the output format in context of a human reader, because a machine-readable output is easier to implement separately (with some `--json` or `--output json` CLI option).
>
> Now the output can be easily parsed using regular shell utilities. We can make things complicated here, or we can just keep it simple.

I would separate goals of the `tt status` command and its properties.

In my understanding the goal is to show a human if something is wrong with the instances. Without the `config:info().{status,alerts}`, `box.info.status`, `box.info.replication` information the goal is not reached in typical scenarios. All the cases from #836 are from my experience with users who reached me during investigation of a problem and in all the cases `tt status` says nothing suspectful -- the instances are running.

In contrary, ability to parse the output using simple `grep`/`awk` expressions is a property, not the goal. And I guess that the expressions are more or less simple anyway (just add `| grep -v '^ '` to exclude extra rows if we're about the currently implemented output format). `awk`'s `{ print $2; }` can be kept as before.

> > I think that if something is suspectful on an instance, we should let a user know about it by default. OTOH, we're not obligated to print anything if everything is OK.
>
> An instance status looks interesting in most cases from my point of view. Even if it is `running`. The column appearing and disappearing time to time looks strange here.

Hm. I didn't mean to touch the existing status anyhow (which is `RUNNING` if the process exists AFAIU).

However, if we're discussing it, it actually brings not much information. It would be better to give some cumulative 'overview' status that takes into account the `config:info()` and the `box.info` information. For example, `UNCALLABLE`, `ORPHAN`, `RO`, `RW`, `RO/HAS WARNINGS` (`RW/HAS WARNINGS`), `RO/HAS ERRORS` (`RW/HAS ERRORS`). OTOH, it is not where I insist anyhow.

In the quoted sentence I meant to hide the extra information we're adding in this pull request if there is nothing important in it. I'm OK with showing the new column always or to hide it if it is empty. I'm OK to place all this information to its own rows (start it with some indent) and don't add any column.

IOW, I'm interested in this information to be present and it is not really important for me how exactly it is formatted.

I disagree with your proposal to hide the health information under a CLI option, because it is counter-intuitive to have status command, which just says whether a process is running. It is natural expectation that it checks the well known `box.info` indicators at least.

cli/status/status.go
```diff
@@ -10,6 +12,52 @@ import (
...
+		}
+	}
+
+	if len(formattedMessages) > 0 {
+		return strings.Join(formattedMessages, "; ")
+	}
+	return ""
+}
+
+func FormatUpstreamWarning(upstream map[string]interface{}) string {
```
psergee, line 36: Why is this external?

psergee (commented):
Thank you for the patch.

oleg-jukovec:
> In my understanding the goal is to show a human if something is wrong with the instances. All the cases from #836 are from my experience with users who reached me during investigation of a problem and in all the cases `tt status` says nothing suspectful -- the instances are running.

I agree, but show a brief status of instances. To be able to see at a glance whether there are problems.

> Without the `config:info().{status,alerts}`, `box.info.status`, `box.info.replication` information the goal is not reached in typical scenarios.

Wouldn't it be enough to create a new column just with `box.info.status`? This way, at a quick glance it will be immediately clear that something is going wrong.

The rest can be hidden under a `verbose` option to allow a user to make a brief diagnostic if he wants to.

`tt` is actively used in development. The fact that some instances did not in the `running` status - may be normal at some stages. And each time going through many messages can be difficult and unnecessary.

psergee:
> Perhaps it's better to add an additional flag for this feature (-v/--verbose)? It looks like the output is now becoming less readable, and at the same time much more difficult to parse it in scripts.
> 
> An another idea: print alerts with the verbose flag, but the instance status by default. Something like this:
> 
> ```
> INSTANCE STATE   MODE PID STATUS
> foo      RUNNING RW   123 orphan
> bar      RUNNING RO   123 ready
> ```
> 
> I think we need @psergee opinion here.

tt verbose mode is used to enable debug-level logging that is already present in the tt code, making it easier to troubleshoot issues. This is not about controlling the command output.
Yes, the output can become less readable with long lines. We need some way to control the `tt status` output. If the long line is not the last field, we should truncate it, so we need a maximum width for each field.

Totktonada:
> > Without the `config:info().{status,alerts}`, `box.info.status`, `box.info.replication` information the goal is not reached in typical scenarios.
>
> Wouldn't it be enough to create a new column just with `box.info.status`? This way, at a quick glance it will be immediately clear that something is going wrong.

I agree that it would be nice to show some overall health status. Sadly, tarantool has no such field. `box.info.status` may report `running`, while there are problems with some of the upstreams, last configuration is not applied with an error and so on. If we want a short status, we should deduce it from a few different fields.

cli/status/status.go
mandesero: There can be several statuses here (from `box.info.status` and `config.info.status`), I decided to combine them and write them together in case `status != ready`. For example `check_errors/orphan`. If everything is ok - then `status = ready`. I left some screenshots below.

cli/status/status.go
mandesero: Replaced with one `Eval` call.

cli/status/status.go
mandesero: function removed

oleg-jukovec:
image
The current output will look bad on small width consoles. I would move it under instances and under a special flag.

Totktonada:
@oleg-jukovec @psergee @mandesero I propose to discuss it voicely on the next week to agree on the output format and don't ask Maksim to implement it several times.

DifferentialOrange:
"Replication follow" is already "OK". Moreover, as far as I remember, there are more than two [replication statuses](https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_info/replication/#box-info-replication):
image

DifferentialOrange:
```
HEALTH
ER/ORP/OK
```
is useless without some header telling what is first and what is third.

```
HEALTH
check_errors/orphan
OK
```

is non-consistent, as well as the third one. I think health (if you decided that it should be a single column) should start with a clear message: `OK`, `ERR`, `WARN`, which can be followed by failure reasons: `ERR: check_errors`, `WARN: orphan`, not vice versa.

Totktonada:
> ## Option 2: Ignore box.replication status
>
> It seems that box.replication.status is set to `loading` during `orphan`.

There is no `box.replication` or `box.replication.status` API. We discussed upstream statuses in `box.info.replication`, but it is unclear how they're mapped to your terminology. Please, clarify it and try to be close to tarantool's API/terminology. This way the proposals should be more clear.

Also, if `box.info.status` is `running` (not `orphan`) some of the upstreams may be in a bad state. It seems, we can't ignore their states.

Totktonada:
## Option 4: abbreviation in its own columns

Just raw idea.

| box.info.config.status | abbreviation |
| ---------------------- | ------------ |
| **column name**        | **CFG**      |
| uninitialized          | N/A          |
| startup_in_progress    | ...          |
| reload_in_progress     | ...          |
| check_warnings         | WW           |
| check_errors           | EE           |
| ready                  | OK           |

| box.info.status | abbreviation |
| --------------- | ------------ |
| **column name** | **BOX**      |
| running         | OK           |
| loading         | LO           |
| orphan          | OR           |
| hot_standby     | HS           |

| box.info.replication.*.upstream.status | abbreviation |
| ------------------------------------- | ------------ |
| **column name**                       | **UPS**      |
| connect                               | ...          |
| auth                                  | ...          |
| wait_snapshot                         | ...          |
| fetch_snapshot                        | ...          |
| final_join                            | ...          |
| sync                                  | ...          |
| follow                                | OK           |
| stopped                               | EE           |
| disconnected                          | EE           |

* All OK => OK
* OK + ... => ...
* OK + EE => EE
* ... + EE => EE

```
INSTANCE STATE   MODE  PID  CFG  BOX  UPS
aaa      RUNNING RW    123  N/A  OK   OK
bbb      RUNNING RO    123  ...  ...  ...
ccc      RUNNING RO    123  ...  OK   OK
ddd      RUNNING RO    123  OK   OK   OK
eee      RUNNING RO    123  WW   OK   EE
fff      RUNNING RO    123  EE   OR   EE
```

mandesero:
> ## Option 4: abbreviation in its own columns
> Just raw idea.
> ...

What about `loading` status in `box.info.replication.*.upstream.status`?
```
app2:storage-001-b> box.info.replication[1].upstream
---
- peer: alice@localhost:3301
  lag: 0
  status: loading
  idle: 0.33725906201289
  message: Read access to universe '' is denied for user 'alice'
...
```

mandesero:
> > ## Option 4: abbreviation in its own columns
> > Just raw idea.
> > ...
> 
> What about `loading` status in `box.info.replication.*.upstream.status` (for some reasons it is not in [replication upstream statuses](https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_info/replication/))?
> 
> ```
> app2:storage-001-b> box.info.replication[1].upstream
> ---
> - peer: alice@localhost:3301
>   lag: 0
>   status: loading
>   idle: 0.33725906201289
>   message: Read access to universe '' is denied for user 'alice'
> ...
> ```

oleg-jukovec:
> This is a demonstration of printing alerts. It was decided that they are displayed on top using the option (called `full` for now). Until there is an approved decision, I print all statuses.
> 
> ```
> $ tt status
>  INSTANCE            STATUS   PID     MODE  CFG           BOX      UPS
>  app2:storage-001-a  RUNNING  776821  RW    check_errors  running  follow
>  app2:storage-001-c  RUNNING  776827  RO    ready         running  follow
>  app2:storage-001-b  RUNNING  776824  RO    ready         running  follow
>  app2:router-001-a   RUNNING  776828  RW    ready         running  --
> 
> $ tt status --full=all
> Instance app2:storage-001-a:
>     type = error; message = "[cluster_config] Unexpected field \"ssharding\""
> 
>  INSTANCE            STATUS   PID     MODE  CFG           BOX      UPS
>  app2:storage-001-a  RUNNING  776821  RW    check_errors  running  follow
>  app2:storage-001-c  RUNNING  776827  RO    ready         running  follow
>  app2:storage-001-b  RUNNING  776824  RO    ready         running  follow
>  app2:router-001-a   RUNNING  776828  RW    ready         running  --
> ```

I generally agree, but but I see two potential problems here:
1. What are we show in case of Tarantool 1/2 with/without cartridge?
2. We need describe all columns in the command help.

Please, don't change anything, let's wait for other reviews from @Totktonada  and @psergee .

Totktonada:
@mandesero Personally I'm OK with the proposed output format. It is similar to my proposal 4, but without abbreviations.

My opinion about the abbreviations is the following. Personally I prefer non-abbreviated status information. If other reviewers are OK with it, that's nice. If somebody insist on a short output format, I'm OK with the abbreviated variant as well.

I think that `--full=all` is a kind of repeating (`full` and `all` feels the same). I think that `--details`, `--more`, `--diagnostics` or something like this is better.

oleg-jukovec:
> Personally I prefer non-abbreviated status information.

+

mandesero:
> Personally I prefer non-abbreviated status information. If other reviewers are OK with it, that's nice. If somebody insist on a short output format, I'm OK with the abbreviated variant as well.
> 
> I think that `--full=all` is a kind of repeating (`full` and `all` feels the same). I think that `--details`, `--more`, `--diagnostics` or something like this is better.

@psergee what do you think about this?

psergee:
> > Personally I prefer non-abbreviated status information. If other reviewers are OK with it, that's nice. If somebody insist on a short output format, I'm OK with the abbreviated variant as well.
> > I think that `--full=all` is a kind of repeating (`full` and `all` feels the same). I think that `--details`, `--more`, `--diagnostics` or something like this is better.
> 
> @psergee what do you think about this?

I like to see small but non-abbreviated statuses also. And column names. Probably 
UPS -> UPSTEAM
CFG -> CONFIG
we already have "status" in command name, so it is clear that "upstream" it is about "upstream status".
`--details`, or `--explain`, or `--more` look fine for me.
Shouldn't we skip printing `CONFIG` field if all instances are not tarantool 3?

mandesero:
Taking the above into account:
- Statuses are displayed in full.
- The option is named `details`.

I also think that we can highlight errors with colors based on their level.

Снимок экрана 2024-09-02 в 11 24 41

Now I think we need to address the following questions:
> - Shouldn't we skip printing CONFIG field if all instances are not tarantool 3?
> - What are we show in case of Tarantool 1/2 with/without cartridge?
> - We need to describe all columns in the command help.

@Totktonada  @psergee  @oleg-jukovec

Totktonada:
@mandesero I'm OK the the proposed format and the option name.

> * Shouldn't we skip printing CONFIG field if all instances are not tarantool 3?

I wouldn't skip it, just print `N/A` in the column or something of this kind. However, if other reviewers want to skip the column, I don't mind.

> * What are we show in case of Tarantool 1/2 with/without cartridge?

It seems that it is exactly 'not tarantool 3' / 'not started from config' case.

> * We need to describe all columns in the command help.

Sure.

mandesero:
> We need to describe all columns in the command help.

I suggest the following format `tt help status`.
```
$ tt help status
The 'status' command provides information about the status of Tarantool instances

Columns:
- INSTANCE: The name of the Tarantool instance
- STATUS: The current status of the instance:
	- RUNNING: The instance is up and running
	- NOT RUNNING: The instance is not running
	- ERROR: The process has terminated unexpectedly
- PID: The process ID (PID) of the instance
- MODE: The mode of the instance, indicating its read/write status:
	- RO: The instance is in read-only mode
	- RW: The instance is in read-write mode
- CONFIG: The status returned by require('config'):info().status (for Tarantool 3+)
- BOX: The status returned by box.info.status
- UPSTREAM: The status returned by box.info.replication[*].upstream.status

USAGE
  tt status [<APP_NAME> | <APP_NAME:INSTANCE_NAME>] [flags]


FLAGS
  -d, --details   print detailed alerts
  -h, --help      help for status
  -p, --pretty    pretty-print table
```

Totktonada:
> - PID: The process ID (PID) of the instance

AFAIR, it is PID of a supervisor process. @psergee, please, correct me if I'm wrong.

psergee:
> > * PID: The process ID (PID) of the instance
> 
> AFAIR, it is PID of a supervisor process. @psergee, please, correct me if I'm wrong.

Yes, this is a `tt` watchdog process PID, not tarantool instance.
N/A in config field for <3 tnt ver is ok for me.

psergee:
> The status returned by require('config'):info().status

> The status returned by box.info.replication[*].upstream.status

Do we need to provide code examples in the help section? Is the "replication upstream status", for example, not clear enough?

Totktonada:
How about printing a note like the following if there is at least one non-ok status?

> Add --details for detailed alert and status information.

It is similar to how `valgrind` suggests how to obtain more information:

```
==25516== Rerun with --leak-check=full to see details of leaked memory
```

```
==25314== Reachable blocks (those to which a pointer was found) are not shown.
==25314== To see them, rerun with: --leak-check=full --show-leak-kinds=all
```

And also reminds me `eix`'s size limited output, which shows a suggestion at the end:

```sh
$ eix f
<...>
Found 2942 matches
Only 50 matches displayed on terminal
Set EIX_LIMIT=0 to show all matches
```

mandesero:
> Do we need to provide code examples in the help section? Is the "replication upstream status", for example, not clear enough?

Replaced with a simple description without using code

> How about printing a note like the following if there is at least one non-ok status?

If any errors occur, I print a following message.
```
$ tt status
 INSTANCE            STATUS   PID     MODE  CONFIG        BOX      UPSTREAM
 app2:storage-001-b  RUNNING  168040  RO    check_errors  running  follow

Errors occurred, rerun with --details to see details
```

psergee:
> Errors occurred, rerun with --details to see details

"Errors occurred" sounds like there were errors during the `tt status` command execution. I'm not sure, maybe:

> The status of some instances requires attention. Please rerun the command with the --details flag to see more information.

?

mandesero:
> I'm not sure, maybe:
> 
> > The status of some instances requires attention. Please rerun the command with the --details flag to see more information.
> 
> ?

Sounds great, let's do it that way.

cli/status/lua/instance_state.lua
```diff
@@ -0,0 +1,57 @@
...
+    })
+end
+
+local ok, box_replication = pcall(function() return box.info.replication end)
+if not ok or box_replication == nil then
+    box_replication = {}
+end
+
+local replication_info = {}
+for _, instance in ipairs(box_replication) do
```
Totktonada, line 28: The keys in this table are instance IDs ([code](https://github.com/tarantool/tarantool/blob/8fae8004f79ecd555537960c60c6e646b037c4cc/src/box/lua/info.c#L228-L236) and I guess there is no guarantee that there are no holes in this table if we consder it as an array. It is better to use `pairs` here.

cli/status/lua/instance_state.lua
```diff
@@ -0,0 +1,57 @@
...
+local config_info = {
+    status = config.status,
+    alerts = {}
+}
+for _, alert in pairs(config.alerts) do
+    table.insert(config_info.alerts, {
+        type = alert.type,
+        message = alert.message
+    })
+end
```
Totktonada, line 20: AFAIS, this code just cuts off extra fields. It would be nice to a have some remark, why we need to do it.

cli/status/lua/instance_state.lua
```diff
@@ -0,0 +1,57 @@
...
+        status = "uninitialized",
+        alerts = {},
+    }
+end
+
+local config_info = {
+    status = config.status,
+    alerts = {}
+}
+for _, alert in pairs(config.alerts) do
```
Totktonada, line 15: Nit: `alerts` is an array-like table, we can use `ipairs` here.

cli/status/lua/instance_state.lua
```diff
@@ -0,0 +1,57 @@
+local ok, config = pcall(require, 'config')
+if ok and config ~= nil and type(config) == "table" and type(config.info) == "function" then
```
Totktonada, line 2: This `config.info` type check has non-obvious reason and I would leave a remark. We can explicitly mention that `moonlibs/config` has the same module name and that it shouldn't be considered as tarantool 3.x config module.

cli/status/lua/instance_state.lua
```diff
@@ -0,0 +1,57 @@
...
+    config = config:info()
```
Totktonada, line 3: Nit: The `config` variable stores the module first, then the info result. Then `config_info` stores a shrinked info result. I would name it as `config`, `config_info_full` and `config_info` or somehow like this. I don't like that we name the info result as `config`.

cli/status/status.go
```diff
@@ -1,57 +1,185 @@
...
 	"github.com/jedib0t/go-pretty/v6/table"
 	"github.com/jedib0t/go-pretty/v6/text"
+	"github.com/mitchellh/mapstructure"
 	"github.com/tarantool/tt/cli/connector"
 	"github.com/tarantool/tt/cli/process_utils"
 	"github.com/tarantool/tt/cli/running"
 )
 
+//go:embed lua/instance_state.lua
+var replicationInfoLuaScript string
```
Totktonada, line 19: Nit: Shouldn't the variable have some more general name? It is not only about the replication info.

cli/status/lua/instance_state.lua
```diff
@@ -0,0 +1,57 @@
...
+
+local ok, box_replication = pcall(function() return box.info.replication end)
+if not ok or box_replication == nil then
+    box_replication = {}
+end
+
+local replication_info = {}
+for _, instance in ipairs(box_replication) do
+    table.insert(replication_info, {
+        name = instance.name,
```
Totktonada, line 30: Just in case: does it work good if the name is not set?

cli/status/lua/instance_state.lua
oleg-jukovec: Maybe check the version Tarantool's major version here too?

Totktonada: > Maybe check the version Tarantool's major version here too?

The existing check seems enough, but I don't mind.

cli/status/status.go
```diff
@@ -1,57 +1,185 @@
...
 			if err == nil && len(res) != 0 {
-				mode := res[0].(bool)
-				mode_str := "RO"
-				if !mode {
-					mode_str = "RW"
+				isParsed = mapstructure.Decode(res[0], &instanceState) == nil
+			}
+			if !isParsed {
+				continue
+			}
```
Totktonada, line 103: Just thought: if we can't parse tarantool's state information, doesn't it mean that something likely is wrong with the instance and we should add something into the `--details` output?

cli/status/status.go
```diff
@@ -1,57 +1,185 @@
...
-		row = append(row, procStatus.ColorSprint(procStatus.Status))
+		if _, exists := instances[fullInstanceName]; !exists {
+			instances[fullInstanceName] = map[string]interface{}{
+				"STATUS": "",
+				"PID":    nil,
+				"MODE":   "",
+				"CFG":    "--",
+				"BOX":    "--",
+				"UPS":    "--",
+			}
```
Totktonada, line 80: It feels a bit inconsistent that we have `CONFIG` column in the output and `CFG` field here, the same for `UPSTREAM` and `UPS`.

Also, shouldn't we use a structure here instead of a map?

cli/status/lua/instance_state.lua
```diff
@@ -0,0 +1,57 @@
...
+        upstream = instance.upstream and {
+            status = instance.upstream.status or "--",
+            message = instance.upstream.message
+        } or nil
+    })
+end
```
Totktonada, line 36: We skip items without the `upstream` field on the Go side, so, maybe, don't add them into the result at all?

```lua
for _, instance in ipairs(box_replication) do
    if instance.upstream ~= nil then
        table.insert(<...>, <...>)
    end
end

File: cli/status/status.go [hunk 15]
```diff
@@ -1,57 +1,185 @@
...
+						map[string]interface{}{
+							"STATUS": "",
+							"PID":    nil,
+							"MODE":   "",
+							"CFG":    "--",
+							"BOX":    "--",
+							"UPS":    "--",
+						}
+				}
+				instances[fullInstanceUpstreamName]["UPS"] = repl.Upstream.Status
```
Totktonada, line 129: Should we really report a connetivity problem in the A->B replication data flow that is detected on B in the A's row?

For me it looks like a problem that definitely affects B, but it is not necessary that something is wrong with A. Maybe just a network problem between them.

cli/status/status.go
```diff
@@ -1,57 +1,185 @@
...
+				isParsed = mapstructure.Decode(res[0], &instanceState) == nil
+			}
+			if !isParsed {
+				continue
+			}
+			instances[fullInstanceName]["MODE"] = instanceState.ReadOnly
+			instances[fullInstanceName]["CFG"] = instanceState.ConfigInfo.Status
+			instances[fullInstanceName]["BOX"] = instanceState.BoxStatus
+
+			for _, repl := range instanceState.ReplicationInfo {
```
Totktonada, line 108: Nit: The replication problems analysing block can be moved to its own function, I suppose.

cli/status/status.go
```diff
@@ -1,57 +1,185 @@
...
+				"STATUS": "",
+				"PID":    nil,
+				"MODE":   "",
+				"CFG":    "--",
+				"BOX":    "--",
+				"UPS":    "--",
+			}
+		}
+
+		instances[fullInstanceName]["STATUS"] = procStatus.ColorSprint(procStatus.Status)
```
Totktonada, line 83: If status is `ERROR`, should we add something into the alert list?

cli/status/status.go
```diff
@@ -1,57 +1,185 @@
...
 			Network: "unix",
 			Address: run.ConsoleSocket,
 		})
+
 		if err == nil {
-			res, err := conn.Eval("return (type(box.cfg) == 'function') or box.info.ro",
-				[]any{}, connector.RequestOpts{})
+			var instanceState InstanceState
+			isParsed := false
+			res, err := conn.Eval(replicationInfoLuaScript, []any{},
```
Totktonada, line 96: If the call fails, should we report it via the alert list?

cli/status/status.go
```diff
@@ -1,57 +1,185 @@
...
+			}
+		}
+
+		instances[fullInstanceName]["STATUS"] = procStatus.ColorSprint(procStatus.Status)
 		if procStatus.Code == process_utils.ProcessRunningCode {
-			row = append(row, procStatus.PID)
+			instances[fullInstanceName]["PID"] = procStatus.PID
 		}
 
 		conn, err := connector.Connect(connector.ConnectOpts{
```
Totktonada, line 88: The same regarding a failure in connection: should we report it via the alert list?

Totktonada (approved):
I'm more a user here rather a developer and I would be happy to have this functionality in `tt`.

I left a few comments to places that doubt me, but in general I'm OK with the patch. Thank you for working on it!

Please, proceed with @oleg-jukovec and @psergee.

cli/status/lua/instance_state.lua
mandesero: Added.

cli/status/lua/instance_state.lua
mandesero: Fixed.

cli/status/lua/instance_state.lua
mandesero: remark added.

cli/status/status.go
mandesero: replaced `replication` with `instance`.

cli/status/lua/instance_state.lua
mandesero: added default name `N/A` if `instance.name` is not set.

cli/status/status.go
mandesero: made errors more informative and covered more cases.

cli/status/status.go
mandesero: Renamed.
I don't prefer to use structure here because we can't change the structure in the map directly. We have to do something like this:
```go
s := map[name]
s.field_1 = ...
s.field_2 = ...
map[name] = s
```

cli/status/lua/instance_state.lua
Totktonada: My initial remark is resolved, so I'm closing this discussion. If extra checks are wishful, please, open a separate discussion thread.

cli/status/status.go
mandesero: Done.

cli/status/status.go
mandesero: I think no. The `ERROR` status means that the process died and we can't get a connection to get more information.

cli/status/status.go
mandesero: report added.

cli/status/status.go
mandesero: same as previous

cli/status/status.go
Totktonada: OK, that's just thoughts around. I'll leave all the details about the implementation to @oleg-jukovec and @psergee. Closing this discussion.

cli/status/status.go
mandesero: > Should we really report a connetivity problem in the A->B replication data flow that is detected on B in the A's row?

I actually don't know.

>For me it looks like a problem that definitely affects B, but it is not necessary that something is wrong with A. Maybe just a network problem between them.

Maybe you are right, but it seems to me that if we have any alerts in `box.info.replication.[*].upstream`, then there are none in the instance component itself

cli/status/status.go
Totktonada: I've a similar status-and-details composition in `config:info()` and here if `status` is `check_errors`, then `alerts` have at least one diagnostic (with the `error` type).

So, it seems logical for me to have at least some 'process is died' in the detailed alerts in the case. The opposite (bad status, no alerts) seems less logical.

OTOH, all the information is on the screen anyway, so I don't insist here at all. Feel free to close this discussion with any your decision.

cli/status/status.go
Totktonada: Am I understand you right: taking my example with broken A->B replication you're worrying that nothing said in A's row in the `tt status` output?

Let's assume that we report an upstream's problem according to my proposal: in B's row.

If the problem is purely about a network between A and B, then we see something like `disconnected` in `UPSTREAM` column in B's row. B is disconnected from some of its upstreams. OK, seems intuitive.

If A is in a bad state (the process is terminated, for example), we see this information in A's row and see `disconnected` in B's row. It also looks intuitive.

----

TBH, I have ideas like reporting `1/3 disconnected` or something of this kind in `UPSTREAM`, but I think that I shouldn't raise a discussion around such details in the scope of this pull request.

cli/status/lua/instance_state.lua
```diff
@@ -0,0 +1,71 @@
...
+-- We only need the type of the error and its message for the status report,
+-- as other fields are considered extra and unnecessary for our purposes.
+for _, alert in ipairs(config_info_full.alerts) do
+    table.insert(config_info.alerts, {
+        type = alert.type,
+        message = alert.message
+    })
+end
+
+local ok, box_replication = pcall(function() return box.info.replication end)
```
oleg-jukovec, line 32: It looks like the code hasn't been tested with Tarantool 2.
```suggestion
local ok, box_replication = pcall(function()
    if box.info.replication ~= nil then
         return box.info.replication
    end
    return box.info.cluster
end)
```

oleg-jukovec (changes requested):
Thank you for the patch. It looks great from a user point of view.

Please, add integration [tests ](https://github.com/tarantool/tt/tree/master/test). Pay attention to especially for Tarantool 2 & Tarantool 2 + Cartridge.
Please, add a changelog entry into the [CHANGELOG.md](https://github.com/tarantool/tt/blob/master/CHANGELOG.md).

cli/status/lua/instance_state.lua
Totktonada: Oh, I missed it. Thanks for catching!

I would also highlight that tarantool 2.x has no string instance name, only UUID (and tarantool 3.x may work without the name). It means that we should collect UUIDs for the `--details` report (for instances without a name).

AFAIU, it also means that matching of `tt`'s instance names to upstream names is not so straighforward and needs more complicated logic[^1]. I propose to don't attempt to match upstream/instance name on the Go side (see [here](https://github.com/tarantool/tt/pull/927#discussion_r1758674620)) and use upstream's name/UUID only for the `--details` alerts.

[^1]: Fetch `box.info.id` to the client from all the instances and create an `id2name` mapping. Then use upstream's ID and the `id2name` mapping to determine the name. I think we don't need this complication for now.

oleg-jukovec: In general, I would suggest making full support for Tarantool 3.

But for Tarantool 2 the status should be meaningful if it is not possible to support a column ​​or it is too difficult (as example, we could support alerts for Tarantool 2 + Cartridge, but that will only complicate things and maybe no one needs it, so it would be enough to create an issue).

Totktonada: I agree. In future we can even make the report extensible by a role/an application. However, it seems meaningful to deliver a first version that accumulates some useful information and hear to users.

Totktonada: `box.info.replication` is the same key in 2.x and 3.x. `box.info.cluster` was renamed to `box.info.replicaset`: https://github.com/tarantool/doc/issues/3465, but it is not relevant to the given code.

However, points regarding name/UUID handling are actual.

oleg-jukovec: Sorry, I was wrong. Anyway, the idea is to test and support Tarantool 2 as far as it could be done easily the same.

mandesero: > I would also highlight that tarantool 2.x has no string instance name, only UUID (and tarantool 3.x may work without the name). It means that we should collect UUIDs for the --details report (for instances without a name).

`tt` retrieves instance or application names from the configuration. The bottleneck arose from relying on `box.info.replication.[*].name`. To address this, I added a preparation step before collecting status information. This step extracts the UUID from each available instance and fills a `uuid2name` map.

For alerts printing I suggest the following:
- If `replica.UUID` exists in the `uuid2name` map => report alerts using the replica name.
- Otherwise report alerts using the replica UUID.

This ensures that we prioritize readable names when available, but fallback to UUIDs when necessary.

mandesero:
> Please, add integration [tests ](https://github.com/tarantool/tt/tree/master/test?rgh-link-date=2024-09-17T20%3A05%3A39Z).

Here is the list of test scenarios suggested by @Totktonada:

1. - [ ] Tarantool 3.x with a YAML configuration.
2. - [ ] Tarantool 3.x without a YAML configuration, but with instance names.
3. - [ ] Tarantool 3.x without a YAML configuration and without instance names.
4. - [ ] Tarantool 2.x without Cartridge.
5. - [ ] Tarantool 2.x with Cartridge.

oleg-jukovec:
> > Please, add integration [tests ](https://github.com/tarantool/tt/tree/master/test?rgh-link-date=2024-09-17T20%3A05%3A39Z).
> 
> Here is the list of test scenarios suggested by @Totktonada:
> 
>     1. * [x]  Tarantool 3.x with a YAML configuration.
> 
>     2. * [x]  Tarantool 3.x without a YAML configuration, but with instance names.
> 
>     3. * [x]  Tarantool 3.x without a YAML configuration and without instance names.
> 
>     4. * [ ]  Tarantool 2.x without Cartridge.
> 
>     5. * [ ]  Tarantool 2.x with Cartridge.

In any case we need to test `tt status` in all scenarios. Everything should to work somehow (maybe with some placeholders, but should).

cli/status/status.go
Totktonada: We discussed it with Maksim. Results:

1. Agreed to report B's upstream problem in B's row.
2. Agreed to report simple `disconnected` or similar status if different ones are present. To be precise: report first non-follow status (or just follow if all the upstreams are in the follow status).
3. Agreed on the following `--details` message (or something similar):

   > [upstream][warning] replication from "app3:storage-001-b" is in "disconnected" status: <..error message..>

cli/status/lua/instance_state.lua
Totktonada: Discussed in a chat with Maksim. I agree with this proposal with a few remarks:

1. It is possible that we have no tt's instance name in the mapping, but there is tarantool's instance name in `box.info.replication[<id>].upstream.name`. In this case it worth to report the tarantool's name.
2. Collecting the mapping as a separate step doubles the network requests. It seems acceptable for the initial implementation, but it would be nice to collect this information in the same eval request that grabs statuses. I would keep it as is in the scope of this PR (if other reviewers don't mind) and rework later.

cli/status/status.go
```diff
@@ -1,58 +1,266 @@
...
+	}
+	return strings.Join(filteredLines, "\n")
+}
+
 // StatusOpts contains options for tt status.
 type StatusOpts struct {
 	// Option for pretty-formatted table output.
-	Pretty bool
+	Pretty  bool
+	Details bool
```
psergee, line 39: please add a comment for this external member

cli/status/status.go
mandesero: Done.

cli/status/lua/instance_state.lua
mandesero: Done.

test/integration/status/test_status.py
```diff
@@ -0,0 +1,341 @@
...
+            pass
+        if time.time() > end_time:
+            print(f"[{full_inst_name}]: {status} wait timed out after {timeout} seconds.")
+            return False
+        time.sleep(1)
+
+
+@pytest.mark.skipif(tarantool_major_version < 3,
+                    reason="skip cluster instances test for Tarantool < 3")
+@pytest.mark.flaky(reruns=3, reruns_delay=3)
```
oleg-jukovec, line 98: This is a bad pattern, let's write a working test.

cli/status/status.go
```diff
@@ -1,58 +1,267 @@
...
 	ts.SetOutputMirror(os.Stdout)
-	ts.AppendHeader(table.Row{"INSTANCE", "STATUS", "PID", "MODE"})
+	ts.AppendHeader(
+		table.Row{"INSTANCE", "STATUS", "PID", "MODE", "CONFIG", "BOX", "UPSTREAM"})
+
+	// XXX: This is a temporary solution that currently requires twice the number of network
+	// calls to instances when requesting detailed output. To optimize, minimize network
+	// calls by refactoring to work in the following way: gather the necessary information
+	// from each available instance, create a map between instance UUIDs and instance names,
+	// and then process statuses and alerts.
```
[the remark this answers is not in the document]
mandesero: This solution was initially implemented to add support for mapping `uuid` to `instance_name` for a more user-friendly output ([discussion outcome](https://github.com/tarantool/tt/pull/927#discussion_r1767495439)) without changing the existing code. However, I decided to implement the optimization right away — please check the changes.

test/integration/status/test_status.py
mandesero: Fixed.

oleg-jukovec:
Please, rebase to the master branch.

mandesero:
> Please, rebase to the master branch.

Rebased.

oleg-jukovec (approved):
Thank you for the patch!

psergee (approved):
lgtm

Issue #917, opened 2024-07-31
tt cluster replicasets roles remove


oleg-jukovec:
We need to add a command to remove a role from a Tarantool 3 centralized configuration:
```shell
tt cluster replicaset roles remove <URI> role_name [--replicaset replicaset_name] [--group group_name] [--instance instance_name]
```

URI is a target configuration source (etcd/tcs), see other commands in `tt cluster` as an example.

The command should support following flags (+ a target configuration source flags, as for other commands in the `tt cluster`):

*  `--instance` for an application name target to specify an instance to remove a role.
*  `--replicaset` for an application name target to specify a replicaset to remove a role.
*  `--group` for an application name target to specify a group (vshard-group in the Cartridge case) to specify a group to remove a role.

PR #928, opened 2024-08-21
cluster: add command 'roles remove' 

themilchenko:
@TarantoolBot document
Title: `tt cluster rs roles remove` removes role in config scope provided by flags.

This patch introduces new command for the cluster replicaset module:

```
tt cluster replicaset roles remove <URI> <ROLE_NAME> [flags]
```

URI is a target configuration source (etcd/tcs). ROLE_NAME is a role removing from a Tarantool config storage. It is necessary to provide flag with destination in config for role removing. It can be one or more destinations.

Command supports following flags:
- Flags for the destination of role removing:
  - `--global (-G)` for a global scope to remove a role;
  - `--instance (-i) string` for an application name target to specify
    an instance to remove a role;
  - `--replicaset (-r) string` for an application name target to specify
    a replicaset to remove a role;
  - `--group (-g) string` for an application name target to specify a group.
- Common flag to interact with config storages (etcd / tarantool config
  storage):
  - `--force (-f)` to skip selecting a key for patching;
  - `--username (-u)` is a username (used as etcd/tarantool config
    storage credentials);
  - `--password (-p)` is a password (used as etcd/tarantool config
    storage credentials).

Closes https://github.com/tarantool/tt/issues/917

### DIFF
--- a/cli/cluster/cmd/replicaset.go
+++ b/cli/cluster/cmd/replicaset.go
@@ -73,6 +73,9 @@ type PromoteCtx struct {
 }
 
 // pickPatchKey prompts to select a key to patch the config.
+// If force is true, picks the first passed key.
+// If there is only one key prompt not appears. Otherwise
+// appears prompt with message of path in config to patch.
 func pickPatchKey(keys []string, force bool, pathMsg string) (int, error) {
 	if len(keys) == 0 {
 		return 0, fmt.Errorf("no keys for the config patching")
@@ -87,7 +90,7 @@ func pickPatchKey(keys []string, force bool, pathMsg string) (int, error) {
 			label = fmt.Sprintf("%s for destination path %q", label, pathMsg)
 		}
 		programSelect := promptui.Select{
-			Label:        "Select a key for the config patching",
+			Label:        label,
 			Items:        keys,
 			HideSelected: true,
 		}
@@ -272,17 +275,17 @@ func Expel(uri *url.URL, ctx ExpelCtx) error {
 	return err
 }
 
-// RolesAddCtx describes the context to add role of instance.
-type RolesAddCtx struct {
-	// InstName is an instance name in which add or remove role.
+// RolesChangeCtx describes the context to add/remove role.
+type RolesChangeCtx struct {
+	// InstName is an instance name in which add/remove role.
 	InstName string
-	// GroupName is an replicaset name in which add or remove role.
+	// GroupName is an replicaset name in which add/remove role.
 	GroupName string
-	// ReplicasetName is an replicaset name in which add or remove role.
+	// ReplicasetName is an replicaset name in which add/remove role.
 	ReplicasetName string
-	// IsGlobal is an boolean value if role needs to add in global scope.
+	// IsGlobal is an boolean value if role needs to add/remove in global scope.
 	IsGlobal bool
-	// RoleName is a name of role to add.
+	// RoleName is a name of role to add/remove.
 	RoleName string
 	// Publishers is data publisher factory.
 	Publishers libcluster.DataPublisherFactory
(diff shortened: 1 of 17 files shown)


cli/cluster/cmd/replicaset.go
```diff
@@ -73,30 +73,27 @@ type PromoteCtx struct {
 }
 
 // pickPatchKey prompts to select a key to patch the config.
-func pickPatchKey(keys []string, force bool, pathMsg string) (int, error) {
-	if len(keys) == 0 {
+// If force is true, picks the first passed key.
+func pickPatchKey(opts replicaset.PickPatchKeyOpts) (int, error) {
```
oleg-jukovec, line 77: Or I missed something?
```suggestion
func pickPatchKey(keys []string, force bool) (int, error) {
```

cli/cluster/cmd/replicaset.go
```diff
@@ -317,7 +314,66 @@ func AddRole(uri *url.URL, ctx RolesAddCtx) error {
...
+		Force:          ctx.Force,
+	})
+	if err == nil {
+		log.Info("Done.")
+	}
+	return err
+}
+
+// RolesRemoveCtx describes the context to remove role.
+type RolesRemoveCtx struct {
```
oleg-jukovec, line 332: If this structure is the same as `RolesAddCtx`, than is will be better to rename `RolesAddCtx` to a common/single `RolesChangeCtx` instead of using both.

cli/cluster/cmd/replicaset.go
```diff
@@ -317,7 +314,66 @@ func AddRole(uri *url.URL, ctx RolesAddCtx) error {
...
+	Username string
+	// Password defines a password for connection.
+	Password string
+	// Force true if the key selection for patching the config
+	// should be skipped.
+	Force bool
+}
+
+// RemoveRole removes a role by patching the cluster config.
+func RemoveRole(uri *url.URL, ctx RolesRemoveCtx) error {
```
oleg-jukovec, line 357: The same, it could be refactored to something common:
```suggestion
// RemoveRole removes a role by patching the cluster config.
func ChangeRole(uri *url.URL, ctx RolesChangeCtx, action RolesChanger/RolesChangeAction) error {
```
The `RolesChanger/RolesChangeAction` could be an interface/func type, or enum + switch.

cli/replicaset/configsource.go
```diff
@@ -10,15 +10,25 @@ import (
...
+	pTargets []patchTarget
+	// Keys is a destination for patching a key in storage.
+	Keys []string
+	// Force is an option for picking the first passed key
+	// if it is true.
+	Force bool
+	// PathMsg is a path destination in config to patch. If it is not
+	// empty path displays for user.
+	PathMsg string
 }
```
oleg-jukovec, line 27: To be honest, this is turning into something scary. It would be better to think about another design here.

cli/replicaset/roles.go
```diff
@@ -2,17 +2,17 @@ package replicaset
 
 import "fmt"
 
-// RolesAddCtx describes a context for adding roles.
-type RolesAddCtx struct {
-	// InstName is an instance name in which add role.
+// RolesUpdateCtx describes a context for adding/removing roles.
+type RolesUpdateCtx struct {
```
oleg-jukovec, line 6: Also a good name as an alternative for `RolesChangeCtx`.

oleg-jukovec (commented):
Please, add a changelog entry to the `CHANGELOG.md`.

Thank you for the patch! A couple of non-critical comments:

Totktonada:
I see that the fix of the linter checks is also present in the other pull request (#927). I think it would be convenient to have it in `master` if there are no objection. Can we push it separately if it allows to do it earlier?

cli/cluster/cmd/replicaset.go
themilchenko: Yes, It's a typo. Corrected.

cli/cluster/cmd/replicaset.go
themilchenko: Refactored to the `RolesChanger` with func type.

cli/replicaset/configsource.go
themilchenko: Deleted opts struct and passed args with enumeration how it was.

cli/replicaset/roles.go
themilchenko: Edited.

oleg-jukovec (approved):

Issue #483, opened 2023-06-07
it is not clear where tarantool is installed


ligurio:
```
[0] ~/sources/MRG/tarantool$ tt install tarantool 2.11.0
   • Checking existing...     
   • Installing tarantool=2.11.0
   • Checking dependencies... 
   • Downloading tarantool... 
   • Patching tarantool...    
   • Building tarantool...    
   • Copying executable...    
   • Copying headers...       
   • Changing symlinks...     
   • Done.                    
[0] ~/sources/MRG/tarantool$ man tt
No manual entry for tt
[0] ~/sources/MRG/tarantool$ 
```

Please add a message with a text like this:

"tarantool executable is installed to /opt/tarantool/bin/tarantool_2.11.0 and made default by symlink to /opt/tarantool/bin/tarantool. It is worth adding /opt/tarantool/bin to a PATH: export PATH=$PATH: /opt/tarantool/bin/".

Japsty:
I would like to work on this issue. Could you please assign it to me?

PR #935, opened 2024-09-03
install: added message with installation path

Japsty:
Since it was not entirely clear where tarantool/tt was installed, messages were added indicating the installation path of binary and symbolic link.

Closes https://github.com/tarantool/tt/issues/483

### DIFF
--- a/cli/install/install.go
+++ b/cli/install/install.go
@@ -418,8 +418,15 @@ func copyBuildedTT(binDir, path, version string, installCtx InstallCtx,
 			return err
 		}
 	}
-	err = util.CopyFilePreserve(filepath.Join(path, "tt"), filepath.Join(binDir, version))
-	return err
+
+	binPathTt := filepath.Join(binDir, version)
+	err = util.CopyFilePreserve(filepath.Join(path, "tt"), binPathTt)
+	if err != nil {
+		return err
+	}
+	log.Infof("Tt executable is installed to: %q", binPathTt)
+
+	return nil
 }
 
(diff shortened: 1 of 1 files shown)


psergee (commented):
Thank you for the patch!

cli/install/install.go
```diff
@@ -1204,6 +1206,8 @@ func installTarantool(binDir string, incDir string, installCtx InstallCtx,
 	}
 
 	log.Infof("Done.")
+	log.Infof("Tarantool executable is installed to: %s/%s", binDir, versionStr)
```
psergee, line 1209: It is possible to forget to update this log message if the logic of the copyBuilded* function is changed, as we generate these paths twice (in the function and in the log message). It would be more appropriate to provide a message regarding the installed executable, even if the symbolic link creation fails.

Japsty:
This is how the message output for tt and tarantool looks like now: 
image
image

cli/install/install.go
```diff
@@ -855,7 +864,12 @@ func copyBuildedTarantool(binPath, incPath, binDir, includeDir, version string,
 	} else if err != nil {
 		return fmt.Errorf("unable to create %s\n Error: %s", includeDir, err)
 	}
+
 	err = copy.Copy(incPath, filepath.Join(includeDir, version)+"/")
+	if err == nil {
+		log.Infof("Tarantool executable is installed to: %q", filepath.Join(binDir, version))
```
psergee, line 870: Let's use the var for the target file name and for printing it.

psergee (approved):

PR #936, opened 2024-09-03
tt: add command `tt upgrade`

mandesero:
### `tt upgrade` command steps:
- For each replicaset:
  - On the master instance:
    1. Execute the following commands sequentially:
       ```lua
       box.schema.upgrade()
       box.snapshot()
       ```
  - On each replica:
    1. Wait until the replica applies all transactions produced by `box.schema.upgrade()` on the master by comparing the vector clocks (vclock).
    2. Once synchronization is confirmed, execute the following command on the replica:
       ```lua
       box.snapshot()
       ```

> If any errors occur during the upgrade process, the process will stop and an error report will be generated.

-----

The replica is waiting for synchronization for `Timeout` seconds. The default value for `Timeout` is 5 seconds, but you can specify it manually using the `--timeout` option.
```bash
$tt upgrade [<APP_NAME>] --timeout 10
```

You can also specify which replicaset(s) to upgrade by using the `--replicaset` option.
```bash
$tt upgrade [<APP_NAME>] --replicaset <RS_NAME_1> -r <RS_NAME_2> ...
```

### DIFF
--- a/cli/replicaset/cmd/lua/upgrade.lua
+++ b/cli/replicaset/cmd/lua/upgrade.lua
@@ -1,0 +1,10 @@
+local ok, err = pcall(box.schema.upgrade)
+if ok then
+   ok, err = pcall(box.snapshot)
+end
+
+return {
+        lsn = box.info.lsn,
+        iid = box.info.id,
+        err = (not ok) and tostring(err) or nil,
+}
--- a/cli/replicaset/cmd/status.go
+++ b/cli/replicaset/cmd/status.go
@@ -10,41 +10,43 @@ import (
 	"github.com/tarantool/tt/cli/running"
 )
 
-// StatusCtx contains information about replicaset status command execution
-// context.
-type StatusCtx struct {
+// DiscoveryCtx contains information about replicaset discovery.
+type DiscoveryCtx struct {
 	// IsApplication true if an application passed.
 	IsApplication bool
 	// RunningCtx is an application running context.
 	RunningCtx running.RunningCtx
 	// Conn is an active connection to a passed instance.
 	Conn connector.Connector
-	// Orchestrator is a forced orchestator choice.
+	// Orchestrator is a forced orchestrator choice.
 	Orchestrator replicaset.Orchestrator
 }
 
-// Status shows a replicaset status.
-func Status(statusCtx StatusCtx) error {
-	orchestratorType, err := getOrchestratorType(statusCtx.Orchestrator,
-		statusCtx.Conn, statusCtx.RunningCtx)
+// getReplicasets discovers and returns the list of replicasets.
+func getReplicasets(ctx DiscoveryCtx) (replicaset.Replicasets, error) {
+	orchestratorType, err := getOrchestratorType(ctx.Orchestrator, ctx.Conn, ctx.RunningCtx)
 	if err != nil {
-		return err
+		return replicaset.Replicasets{}, err
 	}
 
 	var orchestrator replicasetOrchestrator
-	if statusCtx.IsApplication {
-		if orchestrator, err = makeApplicationOrchestrator(
-			orchestratorType, statusCtx.RunningCtx, nil, nil); err != nil {
-			return err
-		}
+	if ctx.IsApplication {
+		orchestrator, err = makeApplicationOrchestrator(orchestratorType,
+			ctx.RunningCtx, nil, nil)
 	} else {
-		if orchestrator, err = makeInstanceOrchestrator(
-			orchestratorType, statusCtx.Conn); err != nil {
-			return err
-		}
+		orchestrator, err = makeInstanceOrchestrator(orchestratorType, ctx.Conn)
 	}
 
-	replicasets, err := orchestrator.Discovery(replicaset.SkipCache)
+	if err != nil {
+		return replicaset.Replicasets{}, err
+	}
+
+	return orchestrator.Discovery(replicaset.SkipCache)
+}
+
+// Status shows a replicaset status.
+func Status(discoveryCtx DiscoveryCtx) error {
+	replicasets, err := getReplicasets(discoveryCtx)
 	if err != nil {
 		return err
 	}
--- a/cli/replicaset/cmd/upgrade.go
+++ b/cli/replicaset/cmd/upgrade.go
@@ -1,0 +1,268 @@
+package replicasetcmd
+
+import (
+	_ "embed"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/mitchellh/mapstructure"
+	"github.com/tarantool/tt/cli/connector"
+	"github.com/tarantool/tt/cli/replicaset"
+	"github.com/tarantool/tt/cli/running"
+)
+
+// UpgradeOpts contains options used for the upgrade process.
+type UpgradeOpts struct {
+	// List of replicaset names specified by the user for the upgrade.
+	ChosenReplicasetAliases []string
+	// Timeout period (in seconds) for waiting on LSN synchronization.
+	LsnTimeout int
+}
+
+type instanceMeta struct {
+	run  running.InstanceCtx
+	conn connector.Connector
+}
+
+//go:embed lua/upgrade.lua
+var upgradeMasterLua string
+
+type syncInfo struct {
+	LSN uint64  `mapstructure:"lsn"`
+	IID uint32  `mapstructure:"iid"`
+	Err *string `mapstructure:"err"`
+}
+
+// filterReplicasetsByAliases filters the given replicaset list by chosen aliases and
+// returns chosen replicasets. If a non-existent alias is found, it returns an error.
+func filterReplicasetsByAliases(replicasets replicaset.Replicasets,
+	chosenReplicasetAliases []string) ([]replicaset.Replicaset, error) {
+	// If no aliases are provided, return all replicasets.
+	if len(chosenReplicasetAliases) == 0 {
+		return replicasets.Replicasets, nil
+	}
+
+	// Create a map for fast lookup of replicasets by alias.
+	replicasetMap := make(map[string]replicaset.Replicaset)
+	for _, rs := range replicasets.Replicasets {
+		replicasetMap[rs.Alias] = rs
+	}
+
+	var chosenReplicasets []replicaset.Replicaset
+	for _, alias := range chosenReplicasetAliases {
+		rs, exists := replicasetMap[alias]
+		if !exists {
+			return nil, fmt.Errorf("replicaset with alias %q doesn't exist", alias)
+		}
+		chosenReplicasets = append(chosenReplicasets, rs)
+	}
+
+	return chosenReplicasets, nil
+}
+
+// Upgrade upgrades tarantool schema.
+func Upgrade(discoveryCtx DiscoveryCtx, opts UpgradeOpts) error {
+	replicasets, err := getReplicasets(discoveryCtx)
+	if err != nil {
+		return err
+	}
+
+	replicasets = fillAliases(replicasets)
+	replicasetsToUpgrade, err := filterReplicasetsByAliases(replicasets,
+		opts.ChosenReplicasetAliases)
+	if err != nil {
+		return err
+	}
+
+	return internalUpgrade(replicasetsToUpgrade, opts.LsnTimeout)
+}
+
+func internalUpgrade(replicasets []replicaset.Replicaset, lsnTimeout int) error {
+	for _, replicaset := range replicasets {
+		err := upgradeReplicaset(replicaset, lsnTimeout)
+		if err != nil {
+			fmt.Printf("• %s: error\n", replicaset.Alias)
+			return fmt.Errorf("replicaset %s: %w", replicaset.Alias, err)
+		}
+		fmt.Printf("• %s: ok\n", replicaset.Alias)
+	}
+	return nil
+}
+
+func closeConnectors(master *instanceMeta, replicas []instanceMeta) {
+	if master != nil {
+		master.conn.Close()
+	}
+	for _, replica := range replicas {
+		replica.conn.Close()
+	}
+}
+
+func getInstanceConnector(instance replicaset.Instance) (connector.Connector, error) {
+	run := instance.InstanceCtx
+	fullInstanceName := running.GetAppInstanceName(run)
+	if fullInstanceName == "" {
+		fullInstanceName = instance.Alias
+	}
+	if fullInstanceName == "" {
+		fullInstanceName = "unknown"
+	}
+
+	// Try to connect via unix socket.
+	conn, err := connector.Connect(connector.ConnectOpts{
+		Network: "unix",
+		Address: run.ConsoleSocket,
+	})
+
+	if err != nil {
+		return nil, fmt.Errorf("instance %s failed to connect via UNIX socket "+
+			": %w", fullInstanceName, err)
+	}
+	return conn, nil
+}
+
+func collectRWROInfo(replset replicaset.Replicaset) (*instanceMeta, []instanceMeta,
+	error) {
+	var master *instanceMeta = nil
+	var replicas []instanceMeta
+	for _, instance := range replset.Instances {
+		run := instance.InstanceCtx
+		fullInstanceName := running.GetAppInstanceName(run)
+		conn, err := getInstanceConnector(instance)
+
+		if err != nil {
+			return nil, nil, err
+		}
+
+		if instance.Mode == replicaset.ModeUnknown {
+			closeConnectors(master, replicas)
+			return nil, nil, fmt.Errorf(
+				"can't determine RO/RW mode on instance: %s", fullInstanceName)
+		}
+
+		isRW := instance.Mode.String() == "rw"
+
+		if isRW && master != nil {
+			closeConnectors(master, replicas)
+			return nil, nil, fmt.Errorf("%s and %s are both masters",
+				running.GetAppInstanceName((*master).run), fullInstanceName)
+		} else if isRW {
+			master = &instanceMeta{run, conn}
+		} else {
+			replicas = append(replicas, instanceMeta{run, conn})
+		}
+	}
+	return master, replicas, nil
+}
+
+func waitLSN(conn connector.Connector, masterIID uint32, masterLSN uint64, lsnTimeout int) error {
+	var lastError error
+	query := fmt.Sprintf("return box.info.vclock[%d]", masterIID)
+
+	deadline := time.Now().Add(time.Duration(lsnTimeout) * time.Second)
+	for {
+		res, err := conn.Eval(query, []any{}, connector.RequestOpts{})
+		if err != nil {
+			lastError = fmt.Errorf("failed to evaluate LSN query: %w", err)
+		} else if len(res) == 0 {
+			lastError = errors.New("empty result from LSN query")
+		} else {
+			var lsn uint64
+			if err := mapstructure.Decode(res[0], &lsn); err != nil {
+				lastError = fmt.Errorf("failed to decode LSN: %w", err)
+			} else if lsn >= masterLSN {
+				return nil
+			} else {
+				lastError = fmt.Errorf("current LSN %d is behind required "+
+					"master LSN %d", lsn, masterLSN)
+			}
+		}
+
+		if time.Now().After(deadline) {
+			break
+		}
+
+		time.Sleep(1 * time.Second)
+	}
+
+	return lastError
+}
+
+func upgradeMaster(master *instanceMeta) (syncInfo, error) {
+	var upgradeInfo syncInfo
+	fullMasterName := running.GetAppInstanceName(master.run)
+	res, err := master.conn.Eval(upgradeMasterLua, []any{}, connector.RequestOpts{})
+	if err != nil {
+		return upgradeInfo, fmt.Errorf(
+			"failed to execute upgrade script on master instance - %s: %w",
+			fullMasterName, err)
+	}
+
+	if err := mapstructure.Decode(res[0], &upgradeInfo); err != nil {
+		return upgradeInfo, fmt.Errorf(
+			"failed to decode response from master instance - %s: %w",
+			fullMasterName, err)
+	}
+
+	if upgradeInfo.Err != nil {
+		return upgradeInfo, fmt.Errorf(
+			"master instance upgrade failed - %s: %w",
+			fullMasterName, err)
+	}
+	return upgradeInfo, nil
+}
+
+func snapshot(instance *instanceMeta) error {
+	res, err := instance.conn.Eval("return box.snapshot()", []any{},
+		connector.RequestOpts{})
+	if err != nil {
+		return fmt.Errorf("failed to execute snapshot on replica: %w", err)
+	}
+	if len(res) == 0 {
+		return fmt.Errorf("snapshot command on %s returned an empty result, "+
+			"'ok' expected", running.GetAppInstanceName(instance.run))
+	}
+
+	if result, ok := res[0].(string); !ok || result != "ok" {
+		return fmt.Errorf("snapshot command on %s returned unexpected result: '%v', "+
+			"'ok' expected", running.GetAppInstanceName(instance.run), res[0])
+	}
+	return nil
+}
+
+func upgradeReplicaset(replicaset replicaset.Replicaset, lsnTimeout int) error {
+	master, replicas, err := collectRWROInfo(replicaset)
+	if err != nil {
+		return err
+	}
+
+	defer closeConnectors(master, replicas)
+
+	// Upgrade master instance, collect LSN and IID from master instance.
+	upgradeInfo, err := upgradeMaster(master)
+	if err != nil {
+		return err
+	}
+
+	// Upgrade replica instances.
+	masterLSN := upgradeInfo.LSN
+	masterIID := upgradeInfo.IID
+
+	for _, replica := range replicas {
+		fullReplicaName := running.GetAppInstanceName(replica.run)
+		err := waitLSN(replica.conn, masterIID, masterLSN, lsnTimeout)
+		if err != nil {
+			return fmt.Errorf("can't ensure that upgrade operations performed on %s "+
+				"are replicated to %s to perform snapshotting on it: error "+
+				"waiting LSN %d in vclock component %d: %w",
+				running.GetAppInstanceName(master.run), fullReplicaName,
+				masterLSN, masterIID, err)
+		}
+		err = snapshot(&replica)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
--- a/test/integration/replicaset/single-t2-app/init.lua
+++ b/test/integration/replicaset/single-t2-app/init.lua
@@ -1,0 +1,11 @@
+local fiber = require('fiber')
+local fio = require('fio')
+
+box.cfg({})
+
+fh = fio.open('ready', {'O_WRONLY', 'O_CREAT'}, tonumber('644',8))
+fh:close()
+
+while true do
+    fiber.sleep(5)
+end
--- a/test/integration/replicaset/test_replicaset_upgrade.py
+++ b/test/integration/replicaset/test_replicaset_upgrade.py
@@ -1,0 +1,194 @@
+import os
+import shutil
+import subprocess
+import tempfile
+
+import pytest
+from replicaset_helpers import stop_application
+from vshard_cluster import VshardCluster
+
+from utils import get_tarantool_version, run_command_and_get_output, wait_file
+
+tarantool_major_version, _ = get_tarantool_version()
+
+
+def run_command_on_instance(tt_cmd, tmpdir, full_inst_name, cmd):
+    con_cmd = [tt_cmd, "connect", full_inst_name, "-f", "-"]
+    instance_process = subprocess.Popen(
+        con_cmd,
+        cwd=tmpdir,
+        stderr=subprocess.STDOUT,
+        stdout=subprocess.PIPE,
+        stdin=subprocess.PIPE,
+        text=True,
+    )
+    instance_process.stdin.writelines([cmd])
+    instance_process.stdin.close()
+    output = instance_process.stdout.read()
+    return output
+
+
+@pytest.mark.skipif(
+    tarantool_major_version < 3, reason="skip centralized config test for Tarantool < 3"
+)
+def test_upgrade_multi_master(tt_cmd, tmpdir_with_cfg):
+    tmpdir = tmpdir_with_cfg
+    app_name = "test_ccluster_app"
+    app_path = os.path.join(tmpdir, app_name)
+    shutil.copytree(os.path.join(os.path.dirname(__file__), app_name), app_path)
+    try:
+        # Start a cluster.
+        start_cmd = [tt_cmd, "start", app_name]
+        rc, out = run_command_and_get_output(start_cmd, cwd=tmpdir)
+        assert rc == 0
+
+        for i in range(1, 6):
+            file = wait_file(
+                os.path.join(tmpdir, app_name), f"ready-instance-00{i}", []
+            )
+            assert file != ""
+
+        status_cmd = [tt_cmd, "replicaset", "upgrade", app_name]
+
+        rc, out = run_command_and_get_output(status_cmd, cwd=tmpdir)
+        assert rc == 1
+        assert "replicaset-002: error" in out and "are both masters" in out
+
+    finally:
+        stop_application(tt_cmd, app_name, tmpdir, [])
+
+
+def test_upgrade_t2_app_dummy_replicaset(tt_cmd):
+    app_name = "single-t2-app"
+    test_app_path_src = os.path.join(os.path.dirname(__file__), app_name)
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+        test_app_path = os.path.join(tmpdir, app_name)
+        shutil.copytree(test_app_path_src, test_app_path)
+        memtx_dir = os.path.join(test_app_path, "var", "lib", app_name)
+        os.makedirs(memtx_dir, exist_ok=True)
+
+        try:
+            start_cmd = [tt_cmd, "start", app_name]
+            rc, out = run_command_and_get_output(start_cmd, cwd=test_app_path)
+            assert rc == 0
+
+            file = wait_file(test_app_path, "ready", [])
+            assert file != ""
+
+            # Downgrade schema.
+            out = run_command_on_instance(
+                tt_cmd,
+                test_app_path,
+                app_name,
+                "box.schema.downgrade('2.8.2') box.snapshot()",
+            )
+
+            upgrade_cmd = [tt_cmd, "replicaset", "upgrade", app_name, "--custom"]
+            rc, out = run_command_and_get_output(upgrade_cmd, cwd=test_app_path)
+            assert rc == 0
+            # Out is `• <uuid>: ok` because the instance has no name.
+            assert "ok" in out
+        finally:
+            stop_application(tt_cmd, app_name, test_app_path, [])
+
+
+@pytest.mark.skipif(tarantool_major_version < 3,
+                    reason="skip test with cluster config for Tarantool < 3")
+def test_upgrade_downgraded_cluster_replicasets(tt_cmd, tmp_path):
+    app_name = "vshard_app"
+    replicasets = {
+        "router-001": ["router-001-a"],
+        "storage-001": ["storage-001-a", "storage-001-b"],
+        "storage-002": ["storage-002-a", "storage-002-b"],
+    }
+    app = VshardCluster(tt_cmd, tmp_path, app_name)
+    try:
+        app.build()
+        app.start()
+        cmd_master = '''box.space._schema:run_triggers(false)
+box.space._schema:delete('replicaset_name')
+box.space._schema:run_triggers(true)
+
+box.space._cluster:run_triggers(false)
+box.atomic(function()
+    for _, tuple in box.space._cluster:pairs() do
+        pcall(box.space._cluster.update, box.space._cluster, {tuple.id}, {{'#', 'name', 1}})
+    end
+end)
+box.space._cluster:run_triggers(true)
+box.schema.downgrade('2.11.1')
+box.snapshot()
+        '''
+
+        # Downgrade cluster.
+        for _, replicaset in replicasets.items():
+            for replica in replicaset:
+                out = run_command_on_instance(
+                    tt_cmd,
+                    tmp_path,
+                    f"{app_name}:{replica}",
+                    "box.cfg{force_recovery=true} return box.cfg.force_recovery"
+                )
+                assert "true" in out
+
+        for _, replicaset in replicasets.items():
+            _ = run_command_on_instance(
+                tt_cmd,
+                tmp_path,
+                f"{app_name}:{replicaset[0]}",
+                cmd_master
+            )
+
+        for _, replicaset in replicasets.items():
+            if len(replicaset) == 2:
+                _ = run_command_on_instance(
+                    tt_cmd,
+                    tmp_path,
+                    f"{app_name}:{replicaset[1]}",
+                    "box.snapshot()"
+                )
+
+        for _, replicaset in replicasets.items():
+            for replica in replicaset:
+                out = run_command_on_instance(
+                    tt_cmd,
+                    tmp_path,
+                    f"{app_name}:{replica}",
+                    "box.cfg{force_recovery=false} return box.cfg.force_recovery"
+                )
+                assert "false" in out
+
+        # Can't create data (old schema).
+        out = run_command_on_instance(
+            tt_cmd,
+            tmp_path,
+            f"{app_name}:storage-001-a",
+            "box.schema.space.create('example_space')"
+        )
+        assert "error: Your schema version is 2.11.1" in out
+
+        # For some reason, the storage-002 replica set is having problems with
+        # replication after downgrade. For now check only replicaset storage-001.
+        upgrade_cmd = [tt_cmd, "replicaset", "upgrade", app_name, "-t=15"]
+        rc, out = run_command_and_get_output(upgrade_cmd, cwd=tmp_path)
+
+        assert rc == 0
+
+        upgrade_out = out.strip().split("\n")
+        assert len(upgrade_out) == len(replicasets)
+
+        for i in range(len(replicasets)):
+            assert "ok" in upgrade_out[i]
+
+        # Create data (new schema).
+        out = run_command_on_instance(
+            tt_cmd,
+            tmp_path,
+            f"{app_name}:storage-001-a",
+            "box.schema.space.create('example_space')"
+        )
+        assert "name: example_space" in out
+
+    finally:
+        app.stop()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 
+- `tt replicaset upgrade`: command to upgrade the schema on a Tarantool cluster.
+  * `-r (--replicaset)`: specify the replicaset name(s) to upgrade.
+  * `-t (--timeout)`: timeout for waiting the LSN synchronization (in seconds) (default 5).
+
 ### Changed
 
 ### Fixed
--- a/cli/cmd/replicaset.go
+++ b/cli/cmd/replicaset.go
@@ -46,11 +46,41 @@ var (
 	replicasetIsGlobal                 bool
 	rebootstrapConfirmed               bool
 
+	chosenReplicasetAliases []string
+	lsnTimeout              int
+
 	replicasetUriHelp = "  The URI can be specified in the following formats:\n" +
 		"  * [tcp://][username:password@][host:port]\n" +
 		"  * [unix://][username:password@]socketpath\n" +
 		"  To specify relative path without `unix://` use `./`."
 )
+
+// newUpgradeCmd creates a "replicaset upgrade" command.
+func newUpgradeCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:                   "upgrade (<APP_NAME>) [flags]",
+		DisableFlagsInUseLine: true,
+		Short:                 "Upgrade tarantool cluster",
+		Long: "Upgrade tarantool cluster.\n\n" +
+			libconnect.EnvCredentialsHelp + "\n\n",
+		Run: func(cmd *cobra.Command, args []string) {
+			cmdCtx.CommandName = cmd.Name()
+			err := modules.RunCmd(&cmdCtx, cmd.CommandPath(), &modulesInfo,
+				internalReplicasetUpgradeModule, args)
+			util.HandleCmdErr(cmd, err)
+		},
+		Args: cobra.ExactArgs(1),
+	}
+
+	cmd.Flags().StringArrayVarP(&chosenReplicasetAliases, "replicaset", "r",
+		[]string{}, "specify the replicaset name(s) to upgrade")
+
+	cmd.Flags().IntVarP(&lsnTimeout, "timeout", "t", 5,
+		"timeout for waiting the LSN synchronization (in seconds)")
+
+	addOrchestratorFlags(cmd)
+	return cmd
+}
 
 // newStatusCmd creates a "replicaset status" command.
 func newStatusCmd() *cobra.Command {
@@ -341,6 +371,7 @@ func NewReplicasetCmd() *cobra.Command {
 		Aliases: []string{"rs"},
 	}
 
+	cmd.AddCommand(newUpgradeCmd())
 	cmd.AddCommand(newStatusCmd())
 	cmd.AddCommand(newPromoteCmd())
 	cmd.AddCommand(newDemoteCmd())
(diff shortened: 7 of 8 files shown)


mandesero:
Examples:
## Case 1: OK
```
INSTANCE            STATUS   PID    MODE
 app2:storage-002-b  RUNNING  12556  RO
 app2:router-001-a   RUNNING  12560  RW
 app2:storage-001-a  RUNNING  12567  RO
 app2:storage-001-b  RUNNING  12554  RO
 app2:storage-002-a  RUNNING  12555  RW
 ```
 ```
 $ tt upgrade app2
 ╭──────────────┬────────────────────┬─────────┬────────────────╮
│ REPLICASET   │ INSTANCE           │ ROLE    │ UPGRADE-STATUS │
├──────────────┼────────────────────┼─────────┼────────────────┤
│ Replicaset 1 │ app2:storage-002-a │ master  │ ok             │
│              │ app2:storage-002-b │ replica │ ok             │
├──────────────┼────────────────────┼─────────┼────────────────┤
│ Replicaset 2 │ app2:router-001-a  │ master  │ ok             │
├──────────────┼────────────────────┼─────────┼────────────────┤
│ Replicaset 3 │ app2:storage-001-a │ master  │ ok             │
│              │ app2:storage-001-b │ replica │ ok             │
╰──────────────┴────────────────────┴─────────┴────────────────╯
 ```
 ------
 ## Case 2: More than one master in the same replicaset
 ```
INSTANCE            STATUS   PID    MODE
 app2:storage-002-b  RUNNING  12556  RO
 app2:router-001-a   RUNNING  12560  RW
 app2:storage-001-a  RUNNING  12567  RO
 app2:storage-001-b  RUNNING  12554  RW
 app2:storage-002-a  RUNNING  12555  RW
 ```
 ```
 $ tt upgrade app2
   ⨯ instances app2:storage-001-a and app2:storage-001-b are both masters in the same replicaset
 ```
 ------
  ## Case 3: LSN didn't update
```
INSTANCE            STATUS   PID    MODE
 app2:storage-002-b  RUNNING  12556  RO
 app2:router-001-a   RUNNING  12560  RW
 app2:storage-001-a  RUNNING  12567  RO
 app2:storage-001-b  RUNNING  12554  RW
 app2:storage-002-a  RUNNING  12555  RO
```
```
$ tt upgrade app2
╭──────────────┬────────────────────┬─────────┬────────────────╮
│ REPLICASET   │ INSTANCE           │ ROLE    │ UPGRADE-STATUS │
├──────────────┼────────────────────┼─────────┼────────────────┤
│ Replicaset 1 │ app2:storage-002-a │ master  │ ok             │
│              │ app2:storage-002-b │ replica │ ok             │
├──────────────┼────────────────────┼─────────┼────────────────┤
│ Replicaset 2 │ app2:router-001-a  │ master  │ ok             │
├──────────────┼────────────────────┼─────────┼────────────────┤
│ Replicaset 3 │ app2:storage-001-b │ master  │ ok             │
╰──────────────┴────────────────────┴─────────┴────────────────╯
   ⨯ [app2:storage-001-a]: LSN wait timeout
```
 ------
 ## Case 4: Not all instances of one replicaset choosen
```
$ tt upgrade app2:storage-001-b
╭──────────────┬────────────────────┬────────╮
│ REPLICASET   │ INSTANCE           │ ROLE   │
├──────────────┼────────────────────┼────────┤
│ Replicaset 1 │ app2:storage-001-b │ master │
╰──────────────┴────────────────────┴────────╯
   ⨯ some instances were not assigned to any replicaset, or not all replicasets include all of their associated instances
   ```
   -----
    ## Case 5: There is a replicaset that does not have a master
   ```
   INSTANCE            STATUS   PID    MODE
 app2:storage-002-b  RUNNING  12556  RO
 app2:router-001-a   RUNNING  12560  RW
 app2:storage-001-a  RUNNING  12567  RO
 app2:storage-001-b  RUNNING  12554  RO
 app2:storage-002-a  RUNNING  12555  RO
   ```
   ```
   $ tt upgrade app2
╭──────────────┬────────────────────┬─────────╮
│ REPLICASET   │ INSTANCE           │ ROLE    │
├──────────────┼────────────────────┼─────────┤
│ Replicaset 1 │ app2:storage-002-a │ master  │
│              │ app2:storage-002-b │ replica │
├──────────────┼────────────────────┼─────────┤
│ Replicaset 2 │ app2:router-001-a  │ master  │
├──────────────┼────────────────────┼─────────┤
│ N/A          │ app2:storage-001-a │ replica │
│ N/A          │ app2:storage-001-b │ replica │
╰──────────────┴────────────────────┴─────────╯
   ⨯ some instances were not assigned to any replicaset, or not all replicasets include all of their associated instances
   ```

cli/upgrade/upgrade.go
```diff
@@ -0,0 +1,182 @@
...
+			}
+			err = WaitLSN(conn, masterUpgradeInfo, lsnTimeout)
+			if err != nil {
+				printReplicasetStatus(rs.Alias, "error")
+				return fmt.Errorf("[%s]: LSN wait timeout: error waiting LSN %d "+
+					"in vclock component %d on %s: time quota %d seconds "+
+					"exceeded", rs.Alias, masterUpgradeInfo.LSN,
+					masterUpgradeInfo.IID, fullInstanceName, lsnTimeout)
+			}
+			_, err = internalUpgrade(conn)
```
Totktonada, line 173: We shouldn't call `box.schema.upgrade()` on replicas (they're anyway in the read-only mode), just `box.snapshot()`.

mandesero: Fixed.

cli/upgrade/upgrade.go
```diff
@@ -0,0 +1,183 @@
...
+
+func WaitLSN(conn connector.Connector, masterUpgradeInfo SyncInfo, timeout int) error {
+	for i := 0; i < timeout; i++ {
+		var res []interface{}
+		res, err := conn.Eval(
+			fmt.Sprintf("return box.info.vclock[%d]", masterUpgradeInfo.IID),
+			[]any{}, connector.RequestOpts{})
+		if err != nil || len(res) == 0 {
+			return err
+		}
```
Totktonada, line 47: It is unclear, whether the `err == nil && len(res) == 0` case is supposed to be a successful result or an error. It looks like this `if` returns an error (it has `return err`), but `err` is `nil` in the case.

Personally, I often use the following pattern (pseudocode):

```go
var res, err = foo()
// First failure type.
if <predicate 1> {
    return err
}
// Second failure type.
if <predicate 2> {
    return err
}
// Success, exit earlier.
if <predicate 3> {
    return res
}
// Continue retrying...
```

The idea is to visually split handling of different `foo`'s exit situations (even if the actions to handle them are the same). It takes more lines, but easier to read.

cli/upgrade/upgrade.go
```diff
@@ -0,0 +1,183 @@
...
+		return upgradeInfo, err
+	}
+	if upgradeInfo.Err != nil {
+		return upgradeInfo, errors.New(*upgradeInfo.Err)
+	}
+	return upgradeInfo, nil
+}
+
+func WaitLSN(conn connector.Connector, masterUpgradeInfo SyncInfo, timeout int) error {
+	for i := 0; i < timeout; i++ {
```
Totktonada, line 40: Wow, this is a non-typical retrying pattern, at least it is surprising for me:

```go
for i := 0; i < timeout; i++ {
    // <..actions..>
    time.Sleep(time.Second)
}
```

I usually use a `double` timeout and something like the following (pseudocode):

```go
var deadline = time.Now() + timeout
for {
    // <..actions..>
    if time.Now() > deadline {
        break // or return error
    }
    time.Sleep(DELAY)
}
```

It performs at least one action disregarding the timeout (even if it is zero) and it accounts a time needs to perform the actions (`Eval` in our case).

cli/upgrade/upgrade.go
```diff
@@ -0,0 +1,183 @@
...
+			lsn = uint32(v)
+		case uint32:
+			lsn = v
+		}
+		if lsn >= masterUpgradeInfo.LSN {
+			return nil
+		}
+		time.Sleep(time.Second)
+	}
+	return errors.New("")
```
Totktonada, line 60: I'm a bit worrying about such an error. Can we write at least 'timeout exceeded' into it? Even if it is not used anywhere, it is more readable.

cli/upgrade/upgrade.go
```diff
@@ -0,0 +1,183 @@
...
+func WaitLSN(conn connector.Connector, masterUpgradeInfo SyncInfo, timeout int) error {
+	for i := 0; i < timeout; i++ {
+		var res []interface{}
+		res, err := conn.Eval(
+			fmt.Sprintf("return box.info.vclock[%d]", masterUpgradeInfo.IID),
+			[]any{}, connector.RequestOpts{})
+		if err != nil || len(res) == 0 {
+			return err
+		}
+		var lsn uint32
```
Totktonada, line 48: I've looked over tarantool's source code and `struct vclock` has an array of `int64_t` LSNs (TBH, I don't know why not `uint64_t`). Using a 64 bit number is logical: a 32 bit counter would overflow after two months of an 1k RPS load.

cli/upgrade/upgrade.go
```diff
@@ -0,0 +1,183 @@
...
+		if lsn >= masterUpgradeInfo.LSN {
+			return nil
+		}
+		time.Sleep(time.Second)
+	}
+	return errors.New("")
+}
+
+func GetAllowedReplicasets(allReplicasets replicaset.Replicasets,
+	pendingReplicasetAliases []string) ([]replicaset.Replicaset, error) {
```
Totktonada, line 64: It seems you're using two terms for the same idea: 'pending' and 'allowed'. It seems, the idea is about replicasets selected by a user. I think we should use one term here. And 'selected' or 'chosen' seems more appropriate here.

psergee:
> ⨯ [storage-001]: LSN wait timeout: error waiting LSN 2003085 in vclock component 1

This message about LSN seems a bit confusing to me. Maybe add something like "Schema upgrade replication timeout exceeded:" before "error waiting LSN..." to make it clearer what has just happened?

cli/upgrade/upgrade.go
```diff
@@ -0,0 +1,183 @@
...
+				return fmt.Errorf("[%s][%s]: %s", rs.Alias, fullInstanceName, err)
+			}
+			err = WaitLSN(conn, masterUpgradeInfo, lsnTimeout)
+			if err != nil {
+				printReplicasetStatus(rs.Alias, "error")
+				return fmt.Errorf("[%s]: LSN wait timeout: error waiting LSN %d "+
+					"in vclock component %d on %s: time quota %d seconds "+
+					"exceeded", rs.Alias, masterUpgradeInfo.LSN,
+					masterUpgradeInfo.IID, fullInstanceName, lsnTimeout)
+			}
```
Totktonada, line 172: The error returned from the LSN waiting function may be useful for a user. There may be different situations: we can't connect to the instance or the LSN just stucks.

I propose the following retrying behavior: each iteration saves its error and goes to the next one. When the timeout is reached, the error from the last iteration is returned.

This error is reported to the user together with the timeout error report. We can just add it at the end of the message.

cli/upgrade/upgrade.go
```diff
@@ -0,0 +1,183 @@
...
+				masterRun = &run
+			} else {
+				replicRun = append(replicRun, run)
+			}
+		}
+		if masterRun == nil {
+			printReplicasetStatus(rs.Alias, "error")
+			return fmt.Errorf("[%s]: has not master instance", rs.Alias)
+		}
+		var conn connector.Connector
```
Totktonada, line 138: Nit: In this context it seems like `masterConn` would be a more intuitive name.

cli/upgrade/upgrade.go
```diff
@@ -0,0 +1,183 @@
...
+		conn, err = connector.Connect(connector.ConnectOpts{
+			Network: "unix",
+			Address: masterRun.ConsoleSocket,
+		})
```
Totktonada, line 142: Can we re-use a connection from the loop above? It seems like we don't have to connect twice.

The same for the replica connections below.

cli/upgrade/upgrade.go
```diff
@@ -0,0 +1,183 @@
...
+	allowedReplicasets, err := GetAllowedReplicasets(replicasets, pendingReplicasetAliases)
+	if err != nil {
+		return err
+	}
+
+	printReplicasetStatus := func(alias, status string) {
+		fmt.Printf("• %s: %s\n", alias, status)
+	}
+
+	for _, rs := range allowedReplicasets {
```
Totktonada, line 98: I'm a bit surpsired that `tt` has no utility function that collects RO/RW information for a replicaset/a cluster. @psergee is it really so?

Anyway, I would comment this loop with something like 'collect RO/RW information', this way it is easier to spot an idea of this block on a glance. Or, better, exclude it isnot a function.

cli/upgrade/upgrade.go
Totktonada, line 172: @psergee Let's discuss it in a thread (to ease navigation over the discussion later).

> > ⨯ [storage-001]: LSN wait timeout: error waiting LSN 2003085 in vclock component 1
>
> This message about LSN seems a bit confusing to me. Maybe add something like "Schema upgrade replication timeout exceeded:" before "error waiting LSN..." to make it clearer what has just happened?

I agree that it may be convenient to have some reason, why the LSN waiting is needed in the first place. However, 'schema upgrade replication timeout' seems a bit vague for me. How about the following variant?

> ⨯ [storage-001-replica]: can't ensure that upgrade operations performed on "storage-001-master" are replicated to "storage-001-replica" to perform snapshotting on it: error waiting LSN 2003085 in vclock component 1: <..last error from eval or actual LSN..>

test/integration/upgrade/test_upgrade.py
```diff
@@ -0,0 +1,106 @@
...
+from vshard_cluster import VshardCluster
+
+from utils import get_tarantool_version, run_command_and_get_output
+
+tarantool_major_version, _ = get_tarantool_version()
+
+
+@pytest.mark.skipif(tarantool_major_version < 3,
+                    reason="skip test with cluster config for Tarantool < 3")
+def test_upgrade_all_replicasets(tt_cmd, tmp_path):
```
Totktonada, line 13: All the test cases perform a no-op upgrade: the instances are already on the last database schema version. As result an implementation that doesn't call `box.schema.upgrade()` at all would pass this test. No-op `WaitLSN` would also pass it.

@psergee Should we add scenarios that perform a real upgrade from, say, a 2.11.1 schema? Or it is considered as too complicated in the given testing infrastructure?

I see two variants how to implement it:

* Save the pre-generated 2.11 snapshot into the repository, copy them into appropriate instance directories before starting the cluster.
* Downgrade the database schema to 2.11.1 on running instances before calling the `tt upgrade` command.

Dowgrading may need extra tricks before https://github.com/tarantool/tarantool/issues/10150 is solved if a cluster configuration is in use.

cli/upgrade/upgrade.go
Totktonada: Taking into account my suggestion below to report either last eval's error or the last seen LSN to a user, I think we should include the information about the last LSN into this message to concatenate it into the report printed to the user.

oleg-jukovec (changes requested):
We have a special subcomand `tt replicaset` for such staff. I suggest to add `tt replicaset upgrade` instead of `tt upgrade`.

oleg-jukovec:
> > We have a special subcomand `tt replicaset` for such staff. I propose to add a `tt replicaset upgrade` command instead of the `tt upgrade`.
> 
> @oleg-jukovec Is it a decision or a proposal to discuss?
> 
> If the former, OK.
> 
> If the latter I would highlight my opinion on it.
> 
>     * We also have `tt cluster` and the command relates to the cluster in some sense -- it may work with several replicasets. From a user point of view `tt cluster` and `tt replicaset` looks very close and where a particular subcommand resides seems arbitrary. And there is `tt cluster replicaset`...
> 
>     * As a user who wish to perform an upgrade I would look for it on the top level first and, if it is not there, would look in `tt cluster`, `tt replicaset`, `tt cluster replicaset`. The verb (the action) looks more important here. I'm thinking on 'upgrading a cluster', not on 'do something with a cluster and this action is upgrade' (it reminds me `find` command syntax :)).
> 
>     * I see no point in these three namespaces, TBH. `tt status` is just `tt status` and nobody complains that it is not `tt <..something..> status`. I'm not a big fan of doing things less obvious for a user, because things were done in a non-obvious way before. Maybe it is a good time to start to change this approach piece-by-piece?
> 
>     * As a meaningful alternative I can propose `tt admin upgrade`, because the upgrade looks as a typical administration task. OTOH, `tt` is at all about administration and development tasks. So `tt upgrade` is still my favorite.

I agree that the current naming `tt replicaset`, `tt cluster`, and `tt cluster replicaset` are confusing. This is a separate issue that we will try to fix in the next major release.

Currently, `tt replicaset` is used for managing the local cluster, including actions like bootstrap, vshard bootstrap, and rebootstrap. The upgrade command, in its current context, logically complements this set of commands.

I don’t see any reason to introduce further confusion with individual commands or to add new sets of subcommands.

Totktonada:
Hm. My [expectation](https://github.com/tarantool/tt/pull/936#discussion_r1766810470) is that the upgrade command is suitable for an arbitrary cluster, not only local cluster. (AFAIU, `tt` can switch to `iproto` for remote instances, at least in some commands.) Does it change its category from `tt replicaset` to something different?

> This should be a centralized design decision that is outside the scope of this pull request.

OK. I'm a bit worrying about a need for compatibility aliases even for such a new functionality. However, I understand the wish to separate the problems.

It sounds like a decision, so I'll not insist on my feelings here anymore (unless the discussion will gain new points from others).

psergee:
Regarding the `tt upgrade` command compared to the `tt replicaset upgrade` command, I agree with Oleg that it should be a subcommand of the `replicaset` command. I still find it confusing to use the `cluster`, `cluster replicaset`, and `replicaset` commands, and I hope that they will be reorganized in a more user-friendly manner in the future. Yes, it could even be called the `tt admin` command or something similar.

mandesero:
The command is now called `tt replicaset upgrade [<APP_NAME> | <URI>] [flags]` (*No tests yet*).
The command works similarly to `tt replicaset status [..args..]`, collecting information about replicasets using the Discovery mechanism.

*Changes:*

- The structure has been refactored to maximize reuse of the new functionality in `tt replicaset downgrade`.
- An initial draft has been created for working with remote instances (`getInstanceConnector`). However, further discussion is needed to determine the best approach for implementing this functionality. The following issues are debatable:
	- 1) To connect to a remote instance using `iproto`, the user needs to grant the `super` or `lua_eval` role.
	- 2) Configuration of the remote cluster needs to be specified (likely requiring a configuration for Tarantool 3.x).
	- 3) There may be other factors to consider.

In the current version, there is already a (workaround-based) ability to perform an upgrade on a remote cluster (each replicaset separately). This uses a mechanism similar to `tt replicaset status <URI>`, where replicaset information is obtained from `box.info.replication` of the instance connected via <URI>.

-----

For example, the application is deployed on the server `10.0.10.123` with the following configuration:

<details>
<summary>Config.yaml</summary>

```yaml
credentials:
  users:
    client:
      password: 'secret'
      roles: [super]
    replicator:
      password: 'secret'
      roles: [replication]
    storage:
      password: 'secret'
      roles: [sharding]

iproto:
  advertise:
    peer:
      login: replicator
    sharding:
      login: storage

sharding:
  bucket_count: 3000

groups:
  storages:
    app:
      module: storage
    sharding:
      roles: [storage]
    replication:
      failover: manual
    replicasets:
      storage-001:
        leader: storage-001-a
        instances:
          storage-001-a:
            iproto:
              listen:
                - uri: 10.0.10.123:3301
          storage-001-b:
            iproto:
              listen:
                - uri: 10.0.10.123:3302
      storage-002:
        leader: storage-002-a
        instances:
          storage-002-a:
            iproto:
              listen:
                - uri: 10.0.10.123:3303
          storage-002-b:
            iproto:
              listen:
                - uri: 10.0.10.123:3304
  routers:
    app:
      module: router
    sharding:
      roles: [router]
    replicasets:
      router-001:
        instances:
          router-001-a:
            iproto:
              listen:
                - uri: 10.0.10.123:3305
```
</details>

Upgrade commands:
```bash
$ tt replicaset upgrade tcp://client:secret@10.0.10.123:3301
• storage-001: ok

$ tt replicaset upgrade tcp://client:secret@10.0.10.123:3303
• storage-002: ok

tt replicaset upgrade tcp://client:secret@10.0.10.123:3305
• router-001: ok
```

However, this is likely not the ideal solution that we aim for, since this method requires a lot of manual work when upgrading a large cluster.

We can make the TCP connection a mock in this patch, support only upgrades on local instances. However, since the mechanism for connecting to remote instances already exists, we might want to implement the full upgrade functionality right away. I would like to hear your thoughts on this @oleg-jukovec @psergee.

cli/upgrade/upgrade.go
mandesero: Fixed.

cli/upgrade/upgrade.go
mandesero: Replaced with the suggested format.

cli/upgrade/upgrade.go
mandesero: Fixed.

cli/upgrade/upgrade.go
mandesero: Fixed.

cli/upgrade/upgrade.go
mandesero: prefix `chosen` is now used.

cli/upgrade/upgrade.go
mandesero: > We can just add it at the end of the message.

Done.

cli/upgrade/upgrade.go
mandesero: Subject to discussion. See my [comment](https://github.com/tarantool/tt/pull/936#issuecomment-2393858831) below.

cli/upgrade/upgrade.go
```diff
@@ -0,0 +1,183 @@
...
+		var masterRun *running.InstanceCtx = nil
+		var replicRun []running.InstanceCtx
+
+		for _, inst := range rs.Instances {
+			run := inst.InstanceCtx
+			fullInstanceName := running.GetAppInstanceName(run)
+			conn, err := connector.Connect(connector.ConnectOpts{
+				Network: "unix",
+				Address: run.ConsoleSocket,
+			})
```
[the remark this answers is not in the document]
mandesero: Subject to discussion. See my [comment](https://github.com/tarantool/tt/pull/936#issuecomment-2393858831) below.

cli/upgrade/upgrade.go
mandesero: Irrelevant.

cli/upgrade/upgrade.go
mandesero: I’m now reusing the connection, but I believe we should implement the `conn.isAlive` method to verify if the connection is active, especially for non-unix socket connections.

cli/upgrade/upgrade.go
mandesero: > utility function that collects RO/RW

I found a functionality to check RO/RW.

 It now operates in two stages: 
 - if `instance.Mode == 'unknown'`, a network call is made to determine the mode;
 - otherwise, the value is taken directly from `instance.Mode`.

cli/upgrade/upgrade.go
mandesero: I chose the last from the ones suggested.

test/integration/upgrade/test_upgrade.py
```diff
@@ -0,0 +1,106 @@
...
+import pytest
+from vshard_cluster import VshardCluster
+
+from utils import get_tarantool_version, run_command_and_get_output
+
+tarantool_major_version, _ = get_tarantool_version()
+
+
+@pytest.mark.skipif(tarantool_major_version < 3,
+                    reason="skip test with cluster config for Tarantool < 3")
```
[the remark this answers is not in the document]
mandesero: Still in progress.

test/integration/upgrade/test_upgrade.py
mandesero: Still in progress.

cli/upgrade/upgrade.go
Totktonada: Not exactly: the last seen LSN is not reported to the user (if eval succeeds, but the `lsn >= masterLSN` check fails). I would set `lastError` in this case too.

cli/upgrade/upgrade.go
Totktonada: There are still `ChosenReplicasetAliases` and `allowedReplicasets`.

cli/upgrade/upgrade.go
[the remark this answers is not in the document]
Totktonada: NB: I filed #968 in case we'll implement the main logic here and remote instance support separately. We're free to solve them both at once if we want/agreed.

cli/upgrade/upgrade.go
mandesero: > I would set `lastError` in this case too.

Added.

cli/upgrade/upgrade.go
mandesero: Fixed.

cli/upgrade/upgrade.go
Totktonada: AFAIU, this `conn.isAlive` check wouldn't affect our scenario much.

The pattern of accesses to the instances is a kind of RPC (remote procedure call), not PubSub (publish-subscribe). So, if we're going to check this `conn.isAlive` property before an eval, it is the same as just attempt to perform the eval itself.

Moreover, an instance may change its RO/RW status or go offline while our eval request goes over the wire or even while it is executed on the instance. A client side `conn.isAlive` property doesn't guarantee that the next request is executed successfully. It means that we anyway should handle eval's failure, disregarding whether we check `conn.isAlive` or not.

*(My question is resolved. Feel free to close this discussion.)*

cli/upgrade/upgrade.go
Totktonada: Thanks, my initial point is solved. I'll look into the precise algorithm how RO/RW is determined and will open separate comments if there is something to discuss.

cli/upgrade/upgrade.go
Totktonada: Thanks!

@psergee Is it OK for you?

cli/replicaset/cmd/upgrade.go
```diff
@@ -0,0 +1,242 @@
...
+
+	if err := mapstructure.Decode(res[0], &upgradeInfo); err != nil {
+		return fmt.Errorf("failed to decode response from master instance - %s: %w",
+			fullMasterName, err)
+	}
+
+	if upgradeInfo.Err != nil {
+		return fmt.Errorf("master instance upgrade failed - %s: %w",
+			fullMasterName, err)
+	}
```
Totktonada, line 215: Nit: I like how the `WaitLSN` function wraps the eval request: all decoding and interpretation of the result is inside. We can do the similar `upgradeMaster`, `snapshot` functions that accept a connection and incapsulates the logic of result processing.

This way this block would be like the following:

```go
res, err := upgradeMaster(master.conn)
if err != nil {
	return err
}
```

Maybe we should use some naming pattern for functions that perform actions on an instance? Say, `execute<...>`? `executeUpgrade` and `executeSnapshot` look good, but `waitLSN` doesn't fit there. And maybe it is OK to leave it just `waitLSN`.

cli/replicaset/cmd/upgrade.go
```diff
@@ -0,0 +1,279 @@
...
+			instance.InstanceCtx = discoveryCtx.RunningCtx.Instances[0]
+			instance.Alias = running.GetAppInstanceName(instance.InstanceCtx)
+			instance.InstanceCtxFound = true
+
+			dummyReplicaset.Alias = instance.Alias
+			dummyReplicaset.Instances = append(dummyReplicaset.Instances, instance)
+			replicasetList = append(replicasetList, dummyReplicaset)
+
+			return internalUpgrade(replicasetList)
+		}
```
mandesero, line 81: I'm not sure if this functionality is really needed, but I thought about it when writing tests (see the last one in `test_replication_upgrade.py`).

cli/replicaset/cmd/upgrade.go
mandesero: I chose to name the wrappers `upgradeMaster` and `Snapshot`.
Done.

test/integration/replicaset/test_replicaset_upgrade.py
```diff
@@ -0,0 +1,245 @@
...
+                os.path.join(tmpdir, app_name), f"ready-instance-00{i}", []
+            )
+            assert file != ""
+
+        _ = run_command_on_instance(
+            tt_cmd, tmpdir, "test_ccluster_app:instance-004", "box.cfg{read_only=true}"
+        )
+
+        # Instance bootstrap may not have finished yet.
+        sleep(5)
```
oleg-jukovec, line 64: Please, replace the sleep with some condition. As example, you could try to execute some command on the instance and wait for a bootstrap.

test/integration/replicaset/test_replicaset_upgrade.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+@pytest.mark.skipif(
+    tarantool_major_version < 3, reason="skip centralized config test for Tarantool < 3"
+)
+def test_upgrade_t2_app_dummy_replicaset(tt_cmd):
+    app_name = "single-t2-app"
+    test_app_path_src = os.path.join(os.path.dirname(__file__), app_name)
+
+    # snapshot from tarantool 2.11.4 app
```
oleg-jukovec, line 119: Please, start all comments with a capital letter and end with a dot.
```suggestion
    # Snapshot from tarantool 2.11.4 app.
```

oleg-jukovec, line 119:
```suggestion
    # Snapshot from tarantool 2.11.4 app.
```

test/integration/replicaset/test_replicaset_upgrade.py
```diff
@@ -0,0 +1,245 @@
...
+    )
+    instance_process.stdin.writelines([cmd])
+    instance_process.stdin.close()
+    output = instance_process.stdout.read()
+    return output
+
+
+@pytest.mark.skipif(
+    tarantool_major_version < 3, reason="skip centralized config test for Tarantool < 3"
+)
```
oleg-jukovec, line 35: The tests in the file is not about centralized config, why do we need to skip it?

cli/replicaset/cmd/status.go
```diff
@@ -73,9 +75,9 @@ func statusReplicasets(replicasets replicaset.Replicasets) error {
 	return nil
 }
 
-// fillAliases fills missed aliases with UUID. The case: Tarantool 1.10 without
+// FillAliases fills missed aliases with UUID. The case: Tarantool 1.10 without
 // an orchestrator.
-func fillAliases(replicasets replicaset.Replicasets) replicaset.Replicasets {
+func FillAliases(replicasets replicaset.Replicasets) replicaset.Replicasets {
```
oleg-jukovec, line 80: We need a unit-test for the method since it is public at now.

cli/replicaset/cmd/upgrade.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+//go:embed lua/upgrade.lua
+var upgradeMasterLua string
+
+type SyncInfo struct {
+	LSN uint64  `mapstructure:"lsn"`
+	IID uint32  `mapstructure:"iid"`
+	Err *string `mapstructure:"err"`
+}
```
oleg-jukovec, line 32: Please, add comments to the public API structs/fields/methods.

cli/replicaset/cmd/upgrade.go
```diff
@@ -0,0 +1,279 @@
...
+
+// "FilterReplicasetsByAliases" filters the given replicaset list by chosen aliases
```
oleg-jukovec, line 34:
```suggestion
// FilterReplicasetsByAliases filters the given replicaset list by chosen aliases
```

cli/replicaset/cmd/upgrade.go
```diff
@@ -0,0 +1,279 @@
...
+// and returns the chosen replicasets. If a non-existent alias is found, it returns an error.
+func FilterReplicasetsByAliases(replicasets replicaset.Replicasets) ([]replicaset.Replicaset,
```
oleg-jukovec, line 36: We could add a unit-test for the call.

cli/replicaset/cmd/upgrade.go
```diff
@@ -0,0 +1,279 @@
...
+	error) {
+	// If no aliases are provided, return all replicasets.
+	if len(ChosenReplicasetAliases) == 0 {
+		return replicasets.Replicasets, nil
+	}
+
+	// Create a map for fast lookup of replicasets by alias
```
oleg-jukovec, line 43:
```suggestion
	// Create a map for fast lookup of replicasets by alias.
```

cli/replicaset/cmd/status.go
```diff
@@ -10,41 +10,43 @@ import (
...
+	// Orchestrator is a forced orchestrator choice.
 	Orchestrator replicaset.Orchestrator
 }
 
-// Status shows a replicaset status.
-func Status(statusCtx StatusCtx) error {
-	orchestratorType, err := getOrchestratorType(statusCtx.Orchestrator,
-		statusCtx.Conn, statusCtx.RunningCtx)
+// GetReplicasets discovers and returns the list of replicasets.
+func GetReplicasets(ctx DiscoveryCtx) (replicaset.Replicasets, error) {
```
oleg-jukovec, line 26: Why do we need to make the call public?

cli/replicaset/cmd/status.go
oleg-jukovec, line 80: Why do we need to make the call public? The same for all new methods/structures.

cli/replicaset/cmd/upgrade.go
```diff
@@ -0,0 +1,279 @@
...
+	})
+
+	if err != nil {
+		// try to connect via TCP [experimental]
+		conn, err = connector.Connect(connector.ConnectOpts{
+			Network:  "tcp",
+			Address:  instance.URI,
+			Username: "client", // should be opt
+			Password: "secret", // should be opt
+		})
```
oleg-jukovec, line 129: It is a strange code. Probably it should be removed.

cli/replicaset/cmd/upgrade.go
```diff
@@ -0,0 +1,279 @@
...
+
+func Snapshot(instance *InstanceMeta) error {
+	res, err := instance.conn.Eval("return box.snapshot()", []any{},
+		connector.RequestOpts{})
+	if err != nil {
+		return fmt.Errorf("failed to execute snapshot on replica: %w", err)
+	}
+	if len(res) == 0 {
+		return fmt.Errorf("snapshot command on %s returned an empty result, "+
+			"'ok' - expected", running.GetAppInstanceName(instance.run))
```
oleg-jukovec, line 239: If 'ok' expected than we need to check the value.

cli/replicaset/cmd/upgrade.go
```diff
@@ -0,0 +1,279 @@
...
+
+	return chosenReplicasets, nil
+}
+
+func Upgrade(discoveryCtx DiscoveryCtx) error {
+	replicasets, err := GetReplicasets(discoveryCtx)
+	if err != nil {
+		// This may be a single-instance application without Tarantool-3 config
+		// or instances.yml file.
+		if len(discoveryCtx.RunningCtx.Instances) == 1 {
```
oleg-jukovec, line 66: GetReplicasets should work in this case.

cli/replicaset/cmd/upgrade.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+	return conn, nil
+}
+
+func СollectRWROInfo(replicaset replicaset.Replicaset, master **InstanceMeta,
+	replicas *[]InstanceMeta) error {
+	for _, instance := range replicaset.Instances {
+		run := instance.InstanceCtx
+		fullInstanceName := running.GetAppInstanceName(run)
+		conn, err := getInstanceConnector(instance)
```
oleg-jukovec, line 143: The connection is never close.

cli/replicaset/cmd/upgrade.go
```diff
@@ -0,0 +1,279 @@
...
+
+		if err != nil {
+			return err
+		}
+
+		var isRW bool
+		if instance.Mode.String() != "unknown" {
```
oleg-jukovec, line 150: We have constants for that.
```suggestion
		if instance.Mode != ModeUnknown {
```

cli/replicaset/cmd/upgrade.go
```diff
@@ -0,0 +1,279 @@
...
+			isRW = instance.Mode.String() == "rw"
+		} else {
+			res, err := conn.Eval(
+				"return (type(box.cfg) == 'function') or box.info.ro",
+				[]any{}, connector.RequestOpts{})
+			if err != nil {
+				return fmt.Errorf("[%s]: %w", fullInstanceName, err)
+			}
+			isRW = !res[0].(bool)
```
oleg-jukovec, line 159: The code may panic by two reasons:
1. len(res) < 1
2. res[0] is not a boolean value.
```suggestion
			isRW = !res[0].(bool)
```

cli/replicaset/cmd/upgrade.go
```diff
@@ -0,0 +1,279 @@
...
+		})
+		if err != nil {
+			return nil, fmt.Errorf("instance %s failed to connect via both TCP "+
+				"and UNIX socket [%s]: %w", fullInstanceName, instance.URI, err)
+		}
+	}
+	return conn, nil
+}
+
+func СollectRWROInfo(replicaset replicaset.Replicaset, master **InstanceMeta,
```
oleg-jukovec, line 138: This is cyrillic 'С'.
```suggestion
func CollectRWROInfo(replicaset replicaset.Replicaset, master **InstanceMeta,
```

cli/replicaset/cmd/upgrade.go
```diff
@@ -0,0 +1,279 @@
...
+	replicas *[]InstanceMeta) error {
```
oleg-jukovec, line 139: Please, avoid pointers to pointers if it is not really required in Go.
```suggestion
func СollectRWROInfo(replicaset replicaset.Replicaset) (InstanceMeta, []InstanceMeta, error) {
```

cli/replicaset/cmd/upgrade.go
```diff
@@ -0,0 +1,279 @@
...
+			return err
+		}
+
+		var isRW bool
+		if instance.Mode.String() != "unknown" {
+			isRW = instance.Mode.String() == "rw"
+		} else {
+			res, err := conn.Eval(
+				"return (type(box.cfg) == 'function') or box.info.ro",
+				[]any{}, connector.RequestOpts{})
```
oleg-jukovec, line 155: Why are we trying to get mode it again? It should be done by a discovery.

oleg-jukovec (changes requested):

oleg-jukovec:
Please, rebase on the latest commit in the master branch.

cli/replicaset/cmd/upgrade.go
mandesero: Using the `GetReplicasets` command doesn't work with applications that lack a configuration. For example, when I start a single application:

```
/start-from-snap $ tree
.
├── init.lua
├── ready
├── tt.yaml
└── var
    ├── lib
    │   └── start-from-snap
    │       └── 00000000000000000000.snap
    ├── log
    │   └── start-from-snap
    │       └── tt.log
    └── run
        └── start-from-snap
            ├── tarantool.control
            ├── tarantool.sock
            └── tt.pid
```

Checking the application status with `tt status` gives the following output:

```
/start-from-snap $ tt status
 INSTANCE         STATUS   PID   MODE  CONFIG         BOX      UPSTREAM
 start-from-snap  RUNNING  3127  RW    uninitialized  running  --
```

However, when attempting to upgrade using `GetReplicasets`, I encounter an error:

```
/start-from-snap $ tt replicaset upgrade start-from-snap
   ⨯ config:get(): no instance config available yet
```

test/integration/replicaset/test_replicaset_upgrade.py
mandesero: I removed this test because it is flaky and duplicates the functionality tested by the last test (`test_upgrade_downgraded_cluster_replicasets`).

test/integration/replicaset/test_replicaset_upgrade.py
mandesero: Removed based on feedback in https://github.com/tarantool/tt/pull/936#discussion_r1818533222

test/integration/replicaset/test_replicaset_upgrade.py
mandesero: Rewrited skip condition according to https://github.com/tarantool/tt/pull/936#discussion_r1818115933

test/integration/replicaset/test_replicaset_upgrade.py
mandesero: The same as the previous.

cli/replicaset/cmd/status.go
mandesero: It's private again.

cli/replicaset/cmd/upgrade.go
mandesero: Done.

cli/replicaset/cmd/upgrade.go
mandesero: Done.

cli/replicaset/cmd/upgrade.go
mandesero: This is now a private method.

cli/replicaset/cmd/upgrade.go
mandesero: Done.

cli/replicaset/cmd/status.go
mandesero: This is now a private method.

cli/replicaset/cmd/status.go
mandesero: This is now a private method.

cli/replicaset/cmd/upgrade.go
mandesero: Done.

cli/replicaset/cmd/upgrade.go
mandesero: Fixed.

cli/replicaset/cmd/upgrade.go
mandesero: Fixed.

cli/replicaset/cmd/upgrade.go
mandesero: Fixed.

cli/replicaset/cmd/upgrade.go
mandesero: Fixed.

cli/replicaset/cmd/upgrade.go
mandesero: Replaced with `func СollectRWROInfo(replicaset replicaset.Replicaset) (InstanceMeta, []InstanceMeta, error)`

cli/replicaset/cmd/upgrade.go
mandesero: I included this as a fallback in case Discovery is unable to determine the mode for some reason.

mandesero:
> Please, rebase on the latest commit in the master branch.

Rebased.

cli/replicaset/cmd/upgrade.go
oleg-jukovec: It seems to me that in this case it is better to do everything once. If discovery failed due to it's implementation - it's better to improve the implementation.

cli/replicaset/cmd/lua/upgrade.lua
```diff
@@ -0,0 +1,11 @@
+local ok, err
+ok, err = pcall(box.schema.upgrade)
```
oleg-jukovec, line 2: The same, but shortest.
```suggestion
local ok, err = pcall(box.schema.upgrade)
```

cli/replicaset/cmd/upgrade.go
```diff
@@ -0,0 +1,272 @@
...
+//go:embed lua/upgrade.lua
+var upgradeMasterLua string
+
+type syncInfo struct {
+	LSN uint64  `mapstructure:"lsn"`
+	IID uint32  `mapstructure:"iid"`
+	Err *string `mapstructure:"err"`
+}
+
+// Filters the given replicaset list by chosen aliases and returns the chosen
```
oleg-jukovec, line 36: It is an idiomatic way to start a comment.
```suggestion
// filterReplicasetsByAlieases filters the given replicaset list by chosen aliases and returns the chosen
```

cli/replicaset/cmd/upgrade.go
```diff
@@ -0,0 +1,272 @@
...
+	"github.com/tarantool/tt/cli/replicaset"
+	"github.com/tarantool/tt/cli/running"
+)
+
+var (
+	// List of replicaset names specified by the user for the upgrade.
+	ChosenReplicasetAliases []string
+	// Timeout period (in seconds) for waiting on LSN synchronization.
+	LsnTimeout int
+)
```
oleg-jukovec, line 20: Please, rewrite the code to avoid using of the global variables. This creates a lot of non-obvious relationships in the code that are more difficult to maintain and test. It's better to create a some type and pass objects of the type into calls.

cli/replicaset/cmd/upgrade.go
```diff
@@ -0,0 +1,272 @@
...
+		if !exists {
+			return nil, fmt.Errorf("replicaset with alias %q doesn't exist", alias)
+		}
+		chosenReplicasets = append(chosenReplicasets, rs)
+	}
+
+	return chosenReplicasets, nil
+}
+
+func Upgrade(discoveryCtx DiscoveryCtx) error {
```
oleg-jukovec, line 63: Like this, as example.
```suggestion

func Upgrade(discoveryCtx DiscoveryCtx, opts UpgrateOpts) error {
```

test/integration/replicaset/test_replicaset_upgrade.py
```diff
@@ -0,0 +1,217 @@
...
+
+        upgrade_out = out.strip().split("\n")
+        assert len(upgrade_out) == len(replicasets)
+
+        for i in range(len(replicasets)):
+            match = re.search(r"•\s*(.*?):\s*(.*)", upgrade_out[i])
+            assert match.group(1) in replicasets
+            assert match.group(2) == "ok"
+
+        # Create data (new schema)
```
oleg-jukovec, line 207:
```suggestion
        # Create data (new schema).
```

oleg-jukovec (commented):
A couple of non-critical comments.

test/integration/replicaset/test_replicaset_upgrade.py
mandesero: Fixed.

cli/replicaset/cmd/upgrade.go
mandesero: Done.

cli/replicaset/cmd/upgrade.go
mandesero: Fixed.

cli/replicaset/cmd/lua/upgrade.lua
mandesero: Done.

cli/replicaset/cmd/upgrade.go
mandesero: Extra code removed.

oleg-jukovec (approved):

cli/replicaset/cmd/lua/upgrade.lua
DerekBum, line 1: Missing newline at the end of this file.

cli/replicaset/cmd/upgrade.go
```diff
@@ -0,0 +1,256 @@
...
+	"fmt"
+	"time"
+
+	"github.com/mitchellh/mapstructure"
+	"github.com/tarantool/tt/cli/connector"
+	"github.com/tarantool/tt/cli/replicaset"
+	"github.com/tarantool/tt/cli/running"
+)
+
+// UpgradeOpts contains the options used for the upgrade process.
```
DerekBum, line 15:
```suggestion
// UpgradeOpts contains options used for the upgrade process.
```

cli/replicaset/cmd/upgrade.go
```diff
@@ -0,0 +1,256 @@
...
+var upgradeMasterLua string
+
+type syncInfo struct {
+	LSN uint64  `mapstructure:"lsn"`
+	IID uint32  `mapstructure:"iid"`
+	Err *string `mapstructure:"err"`
+}
+
+// filterReplicasetsByAliases filters the given replicaset list by chosen aliases and
+// returns the chosen replicasets. If a non-existent alias is found, it returns an error.
```
DerekBum, line 38:
```suggestion
// filterReplicasetsByAliases filters the given replicaset list by chosen aliases and
// returns chosen replicasets. If a non-existent alias is found, it returns an error.
```

cli/replicaset/cmd/upgrade.go
```diff
@@ -0,0 +1,256 @@
...
+		if !exists {
+			return nil, fmt.Errorf("replicaset with alias %q doesn't exist", alias)
+		}
+		chosenReplicasets = append(chosenReplicasets, rs)
+	}
+
+	return chosenReplicasets, nil
+}
+
+func Upgrade(discoveryCtx DiscoveryCtx, opts UpgradeOpts) error {
```
DerekBum, line 64: Maybe add a comment about the public function here?

test/integration/replicaset/single-t2-app/init.lua
DerekBum, line 1: I think there is an extra newline at the end of this file.

cli/replicaset/cmd/upgrade.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+func collectRWROInfo(replset replicaset.Replicaset) (*instanceMeta, []instanceMeta,
+	error) {
+	var master *instanceMeta = nil
+	var replicas []instanceMeta
+	for _, instance := range replset.Instances {
+		run := instance.InstanceCtx
+		fullInstanceName := running.GetAppInstanceName(run)
+		conn, err := getInstanceConnector(instance)
```
DerekBum, line 122: Im still not sure about this. What if some connections were created, but after that we fail on some of the checks below:
```
if instance.Mode == replicaset.ModeUnknow
```
or
```
if isRW && master != nil
```
or even after an error in `upgradeMaster`, `waitLSN` or `snapshot`.
Than error is returned, but none of the created connection are closed. Is it possible to achieve this?

cli/replicaset/cmd/lua/upgrade.lua
mandesero: Fixed.

cli/replicaset/cmd/upgrade.go
mandesero: Fixed.

cli/replicaset/cmd/upgrade.go
mandesero: Fixed.

cli/replicaset/cmd/upgrade.go
mandesero: Added.

test/integration/replicaset/single-t2-app/init.lua
mandesero: Fixed.

DerekBum (approved):

mandesero:
```
It was found that sometimes the replicaset name might not reach tt for some reason. I simplified the check in the test so that the test wouldn't be flaky.

For example:
```
• 5b3a3d0d-5ee2-40d1-989e-d4d68687581e: ok
• router-001: ok
• storage-001: ok
```

But should be:
```
• storage-001: ok
• storage-002: ok
• router-001: ok
```
```

Issue #911, opened 2024-07-30
`tt status` output is unsorted


locker:
**Version**

Tarantool CLI version 2.3.1, linux/amd64. commit: c8f2df1fbe03 (v2.3.1-27-gc8f2df1fbe03)

**Steps to reproduce**

I observed the issue when I started an example Aeon cluster.

1. Download and build Aeon:
   ```
   git clone git@github.com:tarantool/aeon.git
   cd aeon
   cmake . && make
   ```
2. Start the example:
   ```
   cd examples/local-cluster
   tt start
   ```
3. Check the status:
   ```
   tt status
   ```

I guess that the issue can be reproduced with any multi-instance config.

**Actual behavior**

The output of `tt status` appears to be unsorted. Moreover, the order changes every time:

```
~/src/tarantool/aeon/examples/local-cluster$ tt status
 INSTANCE                   STATUS   PID    MODE
 aeondb:aeon-storage-001-a  RUNNING  26978  RW
 aeondb:aeon-storage-002-a  RUNNING  26996  RW
 aeondb:aeon-storage-003-b  RUNNING  26985  RW
 aeondb:aeon-manager-c      RUNNING  27014  RW
 aeondb:aeon-storage-001-b  RUNNING  26983  RO
 aeondb:aeon-storage-001-c  RUNNING  26984  RO
 aeondb:aeon-router-002     RUNNING  27007  RW
 aeondb:aeon-manager-b      RUNNING  27042  RO
 aeondb:aeon-router-001     RUNNING  27033  RW
 aeondb:aeon-storage-002-b  RUNNING  26990  RO
 aeondb:aeon-storage-002-c  RUNNING  27001  RO
 aeondb:aeon-storage-003-a  RUNNING  27002  RO
 aeondb:aeon-storage-003-c  RUNNING  27019  RO
 aeondb:aeon-manager-a      RUNNING  27039  RO
~/src/tarantool/aeon/examples/local-cluster$ tt status
 INSTANCE                   STATUS   PID    MODE
 aeondb:aeon-storage-003-c  RUNNING  27019  RO
 aeondb:aeon-router-002     RUNNING  27007  RW
 aeondb:aeon-manager-c      RUNNING  27014  RW
 aeondb:aeon-storage-002-a  RUNNING  26996  RW
 aeondb:aeon-storage-002-c  RUNNING  27001  RO
 aeondb:aeon-storage-003-b  RUNNING  26985  RW
[...]

PR #939, opened 2024-09-09
status: added output sort

Japsty:
Output of `tt status` was not sorted by instance name, so added sorting in ascending order.

Closes [911](https://github.com/tarantool/tt/issues/911)


psergee (approved):

PR #945, opened 2024-09-13
install: add the ability to flexibly define the version

### LINKED ISSUE #737: Ability to install a <major>.<minor> version of tarantool
Currently as I can see there is only a possibility to install an exact version of Tarantool, specifying `<major>.<minor>.<patch>`, like `2.11.2`. 
It'll be great to have an ability to install tarantool of a latest available `<major>.<minor>` version, like you can do with pyenv for example:
```
pyenv install 3.12
```

Right now tt results in an error:
```
tt install tarantool 2.11
   • Searching in commits...
   ⨯ the hash must contain at least 7 characters
```

dmyger:
Allow specify version like:
 - 2.11 - Chooses last release patch for specified major.minor set
 - 3 - get last available release version with required major==3
 - 2.11-rc - will select in major.minor version with last release candidate mark
 And many other combination matched with version regular expression.

Closes #737

### DIFF
--- a/cli/install/install.go
+++ b/cli/install/install.go
@@ -75,8 +75,6 @@ const (
 	// We need to give permission for all to execute
 	// read,write for user and only read for others.
 	defaultDirPermissions = 0755
-
-	MajorMinorPatchRegexp = `^[0-9]+\.[0-9]+\.[0-9]+`
 )
 
 // programGitRepoUrls contains URLs of programs git repositories.
@@ -206,20 +204,11 @@ func detectOsName() (string, error) {
 // getVersionsFromRepo returns all available versions from github repository.
 func getVersionsFromRepo(local bool, distfiles string, program string,
 	repolink string) ([]version.Version, error) {
-	versions := []version.Version{}
-	var err error
 
 	if local {
-		versions, err = search.GetVersionsFromGitLocal(filepath.Join(distfiles, program))
-	} else {
-		versions, err = search.GetVersionsFromGitRemote(repolink)
+		return search.GetVersionsFromGitLocal(filepath.Join(distfiles, program))
 	}
-
-	if err != nil {
-		return nil, err
-	}
-
-	return versions, nil
+	return search.GetVersionsFromGitRemote(repolink)
 }
 
 // getCommit returns all available commits from repository.
@@ -502,30 +491,25 @@ func installTt(binDir string, installCtx InstallCtx, distfiles string) error {
 	// The tag format in tt is vX.Y.Z, but the user can use the X.Y.Z format
 	// and this option needs to be supported.
 	if ttVersion != "master" {
-		_, err := version.Parse(ttVersion)
-		if err == nil {
+		if version.IsVersion(ttVersion, false) {
 			log.Infof("Searching in versions...")
 			versions, err := getVersionsFromRepo(installCtx.Local, distfiles,
 				"tt", search.GitRepoTT)
 			if err != nil {
 				return err
 			}
-
-			versionMatches, err := regexp.Match(MajorMinorPatchRegexp, []byte(ttVersion))
+			match, err := version.MatchVersion(ttVersion, versions)
 			if err != nil {
-				return err
-			}
-			for _, ver := range versions {
-				if ttVersion == ver.Str || (versionMatches && "v"+ttVersion == ver.Str) {
-					versionFound = true
-					ttVersion = ver.Str
-					break
+				var errNotFound version.NotFoundError
+				if !errors.As(err, &errNotFound) {
+					return err
 				}
+			} else {
+				versionFound = true
+				ttVersion = match
 			}
-			if !versionFound {
-				return fmt.Errorf("%s version of tt doesn't exist", ttVersion)
-			}
-		} else {
+		}
+		if !versionFound {
 			isPullRequest, pullRequestID = util.IsPullRequest(ttVersion)
 
 			if isPullRequest {
@@ -534,13 +518,13 @@ func installTt(binDir string, installCtx InstallCtx, distfiles string) error {
 				log.Infof("Searching in commits...")
 			}
 
+			var err error
 			ttVersion, pullRequestHash, err = checkCommit(
 				ttVersion, "tt", installCtx, distfiles)
 			if err != nil {
 				return err
 			}
 		}
-
 	}
 
 	versionStr := ""
@@ -1018,24 +1002,25 @@ func installTarantool(binDir string, incDir string, installCtx InstallCtx,
 
 	// Check that the version exists.
 	if tarVersion != "master" {
-		_, err := version.Parse(tarVersion)
-		if err == nil {
+		if version.IsVersion(tarVersion, false) {
 			log.Infof("Searching in versions...")
 			versions, err := getVersionsFromRepo(installCtx.Local, distfiles, "tarantool",
 				search.GitRepoTarantool)
 			if err != nil {
 				return err
 			}
-			for _, ver := range versions {
-				if tarVersion == ver.Str {
-					versionFound = true
-					break
+			match, err := version.MatchVersion(tarVersion, versions)
+			if err != nil {
+				var errNotFound version.NotFoundError
+				if !errors.As(err, &errNotFound) {
+					return err
 				}
+			} else {
+				versionFound = true
+				tarVersion = match
 			}
-			if !versionFound {
-				return fmt.Errorf("%s version of tarantool doesn't exist", tarVersion)
-			}
-		} else {
+		}
+		if !versionFound {
 			isPullRequest, pullRequestID = util.IsPullRequest(tarVersion)
 
 			if isPullRequest {
@@ -1044,6 +1029,7 @@ func installTarantool(binDir string, incDir string, installCtx InstallCtx,
 				log.Infof("Searching in commits...")
 			}
 
+			var err error
 			tarVersion, pullRequestHash, err = checkCommit(
 				tarVersion, "tarantool", installCtx, distfiles)
 			if err != nil {
--- a/cli/uninstall/uninstall_test.go
+++ b/cli/uninstall/uninstall_test.go
@@ -189,7 +189,7 @@ func TestGetAllVersionFormats(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			ttVersions, err := getAllTtVersionFormats(tc.programName, tc.ttVersion)
 			assert.NoError(t, err)
-			assert.Equal(t, ttVersions, tc.expectedVersions)
+			assert.Equal(t, tc.expectedVersions, ttVersions)
 		})
 	}
 }
--- a/cli/version/version_match.go
+++ b/cli/version/version_match.go
@@ -1,0 +1,175 @@
+package version
+
+import (
+	"fmt"
+
+	"github.com/tarantool/tt/cli/util"
+)
+
+type requiredField uint8
+type requiredFields []requiredField
+
+const (
+	// Values according order of Fields in regex from `matchVersionParts()`.
+	requiredFieldBuildName requiredField = iota
+	requiredFieldMajor
+	requiredFieldMinor
+	requiredFieldPatch
+	requiredFieldReleaseType
+	requiredFieldReleaseNum
+	requiredFieldAdditional
+	requiredFieldHash
+	requiredFieldRevision
+
+	countOfRequiredFields
+)
+
+// NotFoundError is sentinel error for case when required version not found.
+type NotFoundError struct {
+	expected string
+}
+
+// Error implements the [error] interface.
+func (e NotFoundError) Error() string {
+	return fmt.Sprintf("not matched with: %q", e.expected)
+}
+
+// exploreMatchVersion examines the given string and retrieves the version components,
+// that define the rules for checking for version consistency.
+func exploreMatchVersion(verStr string) (Version, requiredFields, error) {
+	var version Version
+	fields := make(requiredFields, 0, countOfRequiredFields)
+
+	matches, err := matchVersionParts(verStr, false)
+	if err != nil {
+		return version, fields, err
+	}
+
+	version.BuildName = matches["buildname"]
+	if version.BuildName != "" {
+		fields = append(fields, requiredFieldBuildName)
+	}
+
+	if matches["major"] != "" {
+		if version.Major, err = util.AtoiUint64(matches["major"]); err != nil {
+			return version, fields, fmt.Errorf("can't parse Major: %w", err)
+		}
+		fields = append(fields, requiredFieldMajor)
+	}
+
+	if matches["minor"] != "" {
+		if matches["major"] == "" {
+			return version, fields, fmt.Errorf("minor version requires major to be specified")
+		}
+		if version.Minor, err = util.AtoiUint64(matches["minor"]); err != nil {
+			return version, fields, fmt.Errorf("can't parse Minor: %w", err)
+		}
+		fields = append(fields, requiredFieldMinor)
+	}
+
+	if matches["patch"] != "" {
+		if matches["minor"] == "" {
+			return version, fields, fmt.Errorf("patch version requires minor to be specified")
+		}
+		if version.Patch, err = util.AtoiUint64(matches["patch"]); err != nil {
+			return version, fields, fmt.Errorf("can't parse Patch version: %w", err)
+		}
+		fields = append(fields, requiredFieldPatch)
+	}
+
+	if matches["release"] != "" {
+		version.Release, err = newRelease(matches["release"], matches["releaseNum"])
+		if err != nil {
+			return version, fields, fmt.Errorf("can't parse Release: %w", err)
+		}
+		fields = append(fields, requiredFieldReleaseType)
+		if matches["releaseNum"] != "" {
+			fields = append(fields, requiredFieldReleaseNum)
+		}
+	} else if len(fields) > 0 {
+		// By default require 'release' version.
+		version.Release.Type = TypeRelease
+		fields = append(fields, requiredFieldReleaseType)
+	}
+
+	if matches["additional"] != "" {
+		if version.Additional, err = util.AtoiUint64(matches["additional"]); err != nil {
+			return version, fields, fmt.Errorf("can't parse Additional: %w", err)
+		}
+		fields = append(fields, requiredFieldAdditional)
+	}
+
+	version.Hash = matches["hash"]
+	if version.Hash != "" {
+		fields = append(fields, requiredFieldHash)
+	}
+
+	if matches["revision"] != "" {
+		if version.Revision, err = util.AtoiUint64(matches["revision"]); err != nil {
+			return version, fields, fmt.Errorf("can't parse Revision: %w", err)
+		}
+		fields = append(fields, requiredFieldRevision)
+	}
+
+	version.Str = verStr
+
+	return version, fields, nil
+}
+
+// compareVersions compares with two Version according list of fields.
+//
+// Return `true` if all required version components the same.
+func compareVersions(ref Version, other Version, fields requiredFields) bool {
+	for _, m := range fields {
+		isMatch := false
+		switch m {
+		case requiredFieldBuildName:
+			isMatch = ref.BuildName == other.BuildName
+		case requiredFieldMajor:
+			isMatch = ref.Major == other.Major
+		case requiredFieldMinor:
+			isMatch = ref.Minor == other.Minor
+		case requiredFieldPatch:
+			isMatch = ref.Patch == other.Patch
+		case requiredFieldReleaseType:
+			isMatch = ref.Release.Type == other.Release.Type
+		case requiredFieldReleaseNum:
+			isMatch = ref.Release.Num == other.Release.Num
+		case requiredFieldAdditional:
+			isMatch = ref.Additional == other.Additional
+		case requiredFieldHash:
+			isMatch = ref.Hash == other.Hash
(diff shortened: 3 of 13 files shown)


cli/version/version_tools.go
```diff
@@ -73,28 +73,40 @@ type Version struct {
 	BuildName  string  // Custom build name.
 }
 
-// Parse parses a version string and return the version value it represents.
-func Parse(verStr string) (Version, error) {
-	version := Version{}
-	var err error
-
-	// Part 1 (optional) -> custom build name,
+func createVersionRegexp(is_strict bool) *regexp.Regexp {
```
oleg-jukovec, line 76: camelCase for variable names only.
```suggestion
func createVersionRegexp(isStrict bool) *regexp.Regexp {
```

cli/install/install_test.go
```diff
@@ -11,6 +12,15 @@ import (
...
 )
 
+func parseVersion(verStr []string) []version.Version {
+	versions := make([]version.Version, 0, len(verStr))
+	for _, v := range verStr {
+		ver, _ := version.Parse(v)
+		versions = append(versions, ver)
+	}
+	return versions
+}
```
oleg-jukovec, line 22:
```suggestion
func parseVersion(t testing.TB, verStr []string) []version.Version {
	t.Helper()

	versions := make([]version.Version, 0, len(verStr))
	for _, v := range verStr {
		ver, err := version.Parse(v)
		require.NoError(t, err)
		versions = append(versions, ver)
	}
	return versions
}
```

cli/version/version_match_test.go
```diff
@@ -0,0 +1,235 @@
+package version
```
oleg-jukovec, line 1: I understand that the rules are broken in the `tt` from time to time, but it is better to test only the public API. At least for a new code.

```suggestion
package version_test
```

cli/version/version_tools.go
```diff
@@ -73,28 +73,40 @@ type Version struct {
...
+		`(?:-(?P<release>entrypoint|rc|alpha|beta)(?P<releaseNum>\d+)?)?` +
+		`(?:-(?P<additional>\d+))?` +
+		`(?:-(?P<hash>g[a-f0-9]+))?(?:-r(?P<revision>\d+))?(-gc64|-nogc64)?$`
+	if !is_strict {
+		matchString = strings.Replace(matchString, "{1}", "?", 3)
+	}
+	return regexp.MustCompile(matchString)
+}
+
+func matchVersionParts(version string, is_strict bool) (map[string]string, error) {
```
oleg-jukovec, line 95:
```suggestion
func matchVersionParts(version string, isStrict bool) (map[string]string, error) {
```

cli/uninstall/uninstall_test.go
```diff
@@ -145,7 +145,7 @@ func TestSearchLatestVersion(t *testing.T) {
 	}
 }
 
-func TestGetAllVersionFormats(t *testing.T) {
+func Test_getAllVersionFormats(t *testing.T) {
```
oleg-jukovec, line 148: If you make such changes, then apply them at least for the entire file, and even better for the package.
```suggestion
func TestGetAllVersionFormats(t *testing.T) {
```

cli/version/version_match.go
```diff
@@ -0,0 +1,156 @@
...
+		version.Release, err = newRelease(matches["release"], matches["releaseNum"])
+		if err != nil {
+			return version, fmt.Errorf("can't parse Release; %s", err)
+		}
+		version.requiredFields = append(version.requiredFields, CompareMatchReleaseType)
+		if matches["releaseNum"] != "" {
+			version.requiredFields = append(version.requiredFields, CompareMatchReleaseNum)
+		}
+	} else if len(version.requiredFields) > 0 {
+		// By default require 'release' version
```
oleg-jukovec, line 97:
```suggestion
		// By default require 'release' version.
```

cli/version/version_match.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	version.BuildName = matches["buildname"]
+	if version.BuildName != "" {
+		version.requiredFields = append(version.requiredFields, CompareMatchBuildName)
+	}
+
+	if matches["major"] != "" {
+		if version.Major, err = util.AtoiUint64(matches["major"]); err != nil {
+			return version, fmt.Errorf("can't parse Major; %s", err)
```
oleg-jukovec, line 68: The usual error wrapping looks like this:
```suggestion
			return version, fmt.Errorf("can't parse Major: %w", err)
```

cli/version/version_match.go
```diff
@@ -0,0 +1,156 @@
...
+)
+
+const (
+	IgnoredVersionPart = ^uint64(0)
+)
+
+type CompareResult uint16
+
+const (
+	// Values according order of Fields in regex from `matchVersionParts()`
```
oleg-jukovec, line 16:
```suggestion
	// Values according order of Fields in regex from `matchVersionParts()`.
```

cli/version/version_match.go
```diff
@@ -0,0 +1,156 @@
...
+	if matches["major"] != "" {
+		if version.Major, err = util.AtoiUint64(matches["major"]); err != nil {
+			return version, fmt.Errorf("can't parse Major; %s", err)
+		}
+		version.requiredFields = append(version.requiredFields, CompareMatchMajor)
+	}
+
+	if matches["minor"] != "" {
+		if version.Minor, err = util.AtoiUint64(matches["minor"]); err != nil {
+			return version, fmt.Errorf("can't parse Minor; %s", err)
```
oleg-jukovec, line 75: etc
```suggestion
			return version, fmt.Errorf("can't parse Minor: %w", err)
```

oleg-jukovec (changes requested):
Thank you for the patch, please add a new entry into the `CHANGELOG.md` file.

cli/version/version_match.go
```diff
@@ -0,0 +1,156 @@
...
+	"fmt"
+
+	"github.com/tarantool/tt/cli/util"
+)
+
+const (
+	IgnoredVersionPart = ^uint64(0)
+)
+
+type CompareResult uint16
```
oleg-jukovec, line 13: Since it is a public API of the package, it should be a well-commented.

cli/version/version_match.go
```diff
@@ -0,0 +1,156 @@
...
+			Release: Release{
+				Type: TypeRelease,
+				Num:  IgnoredVersionPart,
+			},
+		},
+		requiredFields: make([]CompareResult, 0, CompareMatchTotal),
+	}
+}
+
+func Explore(verStr string) (MatchVersion, error) {
```
oleg-jukovec, line 53: Since it is a public API of the package, it should be a well-commented [2] etc.

cli/install/install.go
```diff
@@ -502,28 +514,17 @@ func installTt(binDir string, installCtx InstallCtx, distfiles string) error {
...
-			for _, ver := range versions {
-				if ttVersion == ver.Str || (versionMatches && "v"+ttVersion == ver.Str) {
-					versionFound = true
-					ttVersion = ver.Str
-					break
-				}
-			}
-			if !versionFound {
-				return fmt.Errorf("%s version of tt doesn't exist", ttVersion)
+				return fmt.Errorf("version of tt doesn't exist: %s", err)
```
themilchenko, line 527:
```suggestion
				return fmt.Errorf("version of tt doesn't exist: %w", err)
```

cli/install/install.go
```diff
@@ -1018,22 +1020,18 @@ func installTarantool(binDir string, incDir string, installCtx InstallCtx,
...
-				if tarVersion == ver.Str {
-					versionFound = true
-					break
-				}
-			}
-			if !versionFound {
-				return fmt.Errorf("%s version of tarantool doesn't exist", tarVersion)
+			tarVersion, err = findVersion(versions, expectedVersion)
+			if err != nil {
+				return fmt.Errorf("version of tarantool doesn't exist: %s", err)
```
themilchenko, line 1033:
```suggestion
				return fmt.Errorf("version of tarantool doesn't exist: %w", err)
```

cli/version/version_match.go
```diff
@@ -0,0 +1,156 @@
...
+	if matches["minor"] != "" {
+		if version.Minor, err = util.AtoiUint64(matches["minor"]); err != nil {
+			return version, fmt.Errorf("can't parse Minor; %s", err)
+		}
+		version.requiredFields = append(version.requiredFields, CompareMatchMinor)
+	}
+
+	if matches["patch"] != "" {
+		if version.Patch, err = util.AtoiUint64(matches["patch"]); err != nil {
+			return version, fmt.Errorf("can't parse Patch; %s", err)
```
themilchenko, line 82:
```suggestion
			return version, fmt.Errorf("can't parse Patch version: %w", err)
```

cli/version/version_match.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+		version.requiredFields = append(version.requiredFields, CompareMatchPatch)
+	}
+
+	if matches["release"] != "" {
+		version.Release, err = newRelease(matches["release"], matches["releaseNum"])
+		if err != nil {
+			return version, fmt.Errorf("can't parse Release; %s", err)
```
themilchenko, line 90:
```suggestion
			return version, fmt.Errorf("can't parse Release tag: %w", err)
```

cli/version/version_match.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	version.Hash = matches["hash"]
+	if version.Hash != "" {
+		version.requiredFields = append(version.requiredFields, CompareMatchHash)
+	}
+
+	if matches["revision"] != "" {
+		if version.Revision, err = util.AtoiUint64(matches["revision"]); err != nil {
+			return version, fmt.Errorf("can't parse Revision; %s", err)
```
themilchenko, line 115:
```suggestion
			return version, fmt.Errorf("can't parse Revision: %w", err)
```

cli/version/version_match.go
```diff
@@ -0,0 +1,161 @@
...
+
+const NotFoundVersion = "NOT_FOUND"
+
+// exploreMatchVersion examines the given string and retrieves the version components,
+// that define the rules for checking for version consistency.
+func exploreMatchVersion(verStr string) (Version, requiredFields, error) {
+	var version Version
+	fields := make(requiredFields, 0, countOfRequiredFields)
+
+	matches, err := matchVersionParts(verStr, false)
```
themilchenko, line 35: Now I can install version with incorrect name. For example if I try to use versions like `v.2.1`, `.2.1` the output will be:

```bash
$ ./tt install tt v.2.1 
   • Searching in versions... 
   • Checking existing...     
   • Installing tt=v2.2.1     
   • Checking dependencies... 
   • Downloading tt...        
   • Building tt...           
   • Copying executable...    
   • Tt executable is installed to: "/home/milchenko/tarantool/tt/bin/tt_v2.2.1"
   • Made default by symlink "/home/milchenko/tarantool/tt/bin/tt"
   • Use the following command to add the bin_dir directory to the PATH: . <(tt env)
   • Done.    
```

Possibly it can be a feature due to `tt` try to search the latest version by giving pattern even it has incorrect syntax, but in my opinion it is more a bug. Correct me if I'm wrong.

test/integration/install/test_install.py
```diff
@@ -214,6 +215,29 @@ def test_install_tt_specific_version(tt_cmd, tmp_path):
...
 
+@pytest.mark.slow
+def test_install_tt_incomplete_version(tt_cmd, tmp_path):
+    configPath = tmp_path / config_name
+    # Create test config.
+    with open(configPath, "w") as f:
+        f.write("env:\n  bin_dir:\n  inc_dir:\n")
+
+    # Install latest tt.
+    install_cmd = (tt_cmd, "--cfg", configPath, "install", "tt", "2.1")
```
themilchenko, line 226: I think we can test many different scenarios here with different version formats and errors using `parametrize`.

test/integration/install/test_install.py
```diff
@@ -214,6 +215,29 @@ def test_install_tt_specific_version(tt_cmd, tmp_path):
     assert re.search(r"Tarantool CLI version 2.1.2", start_output)
 
 
+@pytest.mark.slow
+def test_install_tt_incomplete_version(tt_cmd, tmp_path):
```
themilchenko, line 219: We need the same tests for tarantool installation.

cli/version/version_match.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	version.BuildName = matches["buildname"]
+	if version.BuildName != "" {
+		fields = append(fields, requiredFieldBuildName)
+	}
+
+	if matches["major"] != "" {
+		if version.Major, err = util.AtoiUint64(matches["major"]); err != nil {
+			return version, fields, fmt.Errorf("can't parse Major; %w", err)
```
themilchenko, line 47: In my opinion `:` is more prettier than `;` symbol.

Besides, as I have noticed, there is a pattern in `tt` that the error wrapper starts with the error description and is complemented after `:` by the error received.

Or am I missing something and this wrap has a different purpose?

```suggestion
			return version, fields, fmt.Errorf("can't parse Major: %w", err)
```

themilchenko (commented):
Please, add note about this feature to the CHANGELOG.

cli/version/version_match.go
```diff
@@ -0,0 +1,161 @@
...
+	requiredFieldReleaseType
+	requiredFieldReleaseNum
+	requiredFieldAdditional
+	requiredFieldHash
+	requiredFieldRevision
+
+	countOfRequiredFields
+)
+
+const NotFoundVersion = "NOT_FOUND"
```
oleg-jukovec, line 27: Please, add a comment to the public API constant.

cli/version/version_match.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+		if !isMatch {
+			return false
+		}
+	}
+	return true
+}
+
+// Check is string looks like version.
+// If isStrict is true, then version <major.minor.patch> components are required.
```
oleg-jukovec, line 139:
```suggestion
// IsVersion checks is string looks like version.
// If isStrict is true, then version <major.minor.patch> components are required.
```

cli/version/version_match.go
dmyger: It seems debatable - what the correct version means.
Here you apply the rule where <major> does not important. And it found first of last version with such match.

If needed, we can set the <major> part as mandatory. But, the best way here should be more interactive behavior of the tool. It should collect all matches and ask user to make a choose.

cli/version/version_match.go
dmyger: Done.

test/integration/install/test_install.py
oleg-jukovec: `tt install` is a takes much time, so it's ok to test with integration tests some basic scenarios, but remaining scenarios should be tested with lighter tests (unit-tests).

cli/version/version_match.go
oleg-jukovec: To be honest, I don't know when such behavior might be useful. So I propose require a major if minor is set, minor if patch is set etc.

test/integration/install/test_install.py
dmyger: I add another generic case for `tt install tt <version>` because this test works fast enough. 
And used `parametrize` for easy future expansion.

cli/version/version_match.go
dmyger: Implement this logic:
- `patch` - requires `minor`
- `minor` - requires `major`

cli/version/version_match.go
```diff
@@ -0,0 +1,161 @@
...
+func IsVersion(version string, isStrict bool) bool {
+	re := createVersionRegexp(isStrict)
+	return re.MatchString(version)
+}
+
+// MatchVersion takes as input an ascending sorted list of versions.
+// In which it finds the latest version matching the search mask.
+//
+// Return string of found version or `NotFoundVersion` with error if just not found
```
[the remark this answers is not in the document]
dmyger: Done.

cli/version/version_match.go
```diff
@@ -0,0 +1,174 @@
...
+	requiredFieldRevision
+
+	countOfRequiredFields
+)
+
+// NotFoundError is sentinel error for case when required version not found.
+type NotFoundError struct {
+	expected string
+}
+
```
oleg-jukovec, line 31: Please, add a comment to the public API method.

cli/version/version_match.go
```diff
@@ -0,0 +1,174 @@
...
+	if err != nil {
+		return "", err
+	}
+	for i := len(sortedVersions) - 1; i >= 0; i-- {
+		ver := sortedVersions[i]
+		if compareVersions(reference, ver, fields) {
+			return ver.Str, nil
+		}
+	}
+	return "", NotFoundError{expected}
```
oleg-jukovec, line 173: Please, add a unit-test with the error.

cli/version/version_match.go
```diff
@@ -0,0 +1,174 @@
...
+	if matches["major"] != "" {
+		if version.Major, err = util.AtoiUint64(matches["major"]); err != nil {
+			return version, fields, fmt.Errorf("can't parse Major: %w", err)
+		}
+		fields = append(fields, requiredFieldMajor)
+	}
+
+	if matches["minor"] != "" {
+		if matches["major"] == "" {
+			return version, fields, fmt.Errorf("parse Minor requires Major: %w", err)
```
themilchenko, line 61: Something wrong with the output, the error is nil:

```bash
$ ./tt install tt .2.1 
   • Searching in versions... 
   ⨯ parse Minor requires Major: %!!(MISSING)w(<nil>)
```

oleg-jukovec:
Please, rebase to the `master` branch.

cli/version/version_match.go
```diff
@@ -0,0 +1,174 @@
...
+		if version.Minor, err = util.AtoiUint64(matches["minor"]); err != nil {
+			return version, fields, fmt.Errorf("can't parse Minor: %w", err)
+		}
+		fields = append(fields, requiredFieldMinor)
+	}
+
+	if matches["patch"] != "" {
+		if matches["minor"] == "" {
+			return version, fields, fmt.Errorf("parse Patch requires Minor: %w", err)
+		}
```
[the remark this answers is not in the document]
dmyger: Due to the way regex works, this case is not possible. This branch of checks in the code is currently unreachable. It is added only in case regex will changes.

cli/version/version_match_test.go
```diff
@@ -0,0 +1,146 @@
...
+		{"1.1", "", "not matched with: 1.1"},
+		{"3.1-0", "", "not matched with: 3.1-0"},
+		{"V1", "", "failed to parse version"},
+		{"18446744073709551616", "", "can't parse Major"},
+		{"1.18446744073709551616", "", "can't parse Minor"},
+		{"0.1.18446744073709551616", "", "can't parse Patch"},
+		{"1.2.3-rc18446744073709551616", "", "bad release number format"},
+		{"1.2.3-18446744073709551616", "", "can't parse Additional"},
+		{"1.2.3-r18446744073709551616", "", "can't parse Revision"},
+		{".2.3", "", "parse Minor requires Major"},
```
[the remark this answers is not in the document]
dmyger: Currently there is no way to specify only `patch`, without `minor`

cli/version/version_match.go
dmyger: Added the comment for `Error()` method

cli/version/version_match.go
dmyger: Yeah, here there is no `err` value

cli/version/version_match.go
dmyger: Done.

cli/version/version_match.go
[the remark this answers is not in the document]
oleg-jukovec: I think we need a prettier error for the user here. Something like:

```minor version requires major to be specified```

The error looks confusing from a user side. There are too many implementation details that the user will not understand.

File: cli/version/version_match.go [hunk 29]
```diff
@@ -0,0 +1,175 @@
...
+	countOfRequiredFields
+)
+
+// NotFoundError is sentinel error for case when required version not found.
+type NotFoundError struct {
+	expected string
+}
+
+// Error - error interface implementation for NotFoundError.
+func (e NotFoundError) Error() string {
```
oleg-jukovec, line 33: Is is usual something like:

https://pkg.go.dev/bytes#Reader.Read
```suggestion
// Error implements the [error] interface.
func (e NotFoundError) Error() string {
```

oleg-jukovec (approved):

cli/version/version_match.go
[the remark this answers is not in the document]
themilchenko: I catch that this case is not possible. I mean that the error a bit confusing. We type something like `.<patch>` and the output says `parse Minor requires Major`.

cli/version/version_match.go
dmyger: Done.

cli/version/version_match.go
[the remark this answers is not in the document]
dmyger: > when we type something like `.<patch>`

We type `.<digit>` and what we put into it nobody knows. In our understanding, it can be either a `.patch` or a `.minor`. But the current regexp considers such input as a `.minor`.

I hope that Oleg's suggestion to change the message will help solve the problem.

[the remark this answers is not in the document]
themilchenko: Ok, let it be.

themilchenko (approved):
Thank you for the patch! LGTM.

PR #951, opened 2024-09-23
tt: add support set delimiter in console

### LINKED ISSUE #727: `tt` connection console hangs during typing commands if delimiter is set
Steps to reproduce:
```
$ tt connect <uri>

> \set delimiter ;

> print('123456');
```
Console is hanging during typing `print('123456');` line. This is due to invocation of the completion handler lua code. After setting a delimiter `tarantool` expects the code to end with the specified delimiter. But `tt` built-in lua code does not have this delimiter at the end and the code is not invoked on the server side, so the console waits for 3 sec for each symbol. And it hangs completely after `;` and enter. Probably due to the same reason.

dmyger:
The “\set delimiter [marker]” command is handled by the `tt` utility, simulating the behavior of a similar command for the Tarantool console.

Closes #727

### DIFF
--- a/cli/connect/commands.go
+++ b/cli/connect/commands.go
@@ -160,7 +160,7 @@ func (command argSetCmdDecorator) Aliases() []string {
 // Run checks that there is one allowed argument and runs the command.
 func (command argSetCmdDecorator) Run(console *Console,
 	cmd string, args []string) (string, error) {
-	if len(args) != 1 || !find(command.sorted, args[0]) {
+	if len(command.sorted) > 0 && (len(args) != 1 || !find(command.sorted, args[0])) {
 		return "", fmt.Errorf("the command expects one of: %s",
 			strings.Join(command.sorted, ", "))
 	}
@@ -361,6 +361,19 @@ func setTableColumnWidthMaxFunc(console *Console,
 	return "", nil
 }
 
+// setDelimiterMarker apply delimiter to the Console object.
+func setDelimiterMarker(console *Console, cmd string, args []string) (string, error) {
+	switch len(args) {
+	case 0:
+		console.delimiter = ""
+	case 1:
+		console.delimiter = args[0]
+	default:
+		return "", fmt.Errorf("the command expects zero or single argument")
+	}
+	return "", nil
+}
+
 // switchNextFormatFunc switches to a next output format.
 func switchNextFormatFunc(console *Console, cmd string, args []string) (string, error) {
 	console.format = (1 + console.format) % formatter.FormatsAmount
@@ -451,6 +464,14 @@ var cmdInfos = []cmdInfo{
 				[]string{setTableColumnWidthMaxLong},
 				setTableColumnWidthMaxFunc,
 			),
+		),
+	},
+	cmdInfo{
+		Short: setDelimiter + " <marker>",
+		Long:  "set expression delimiter",
+		Cmd: newArgSetCmdDecorator(
+			newBaseCmd([]string{setDelimiter}, setDelimiterMarker),
+			[]string{},
 		),
 	},
 	cmdInfo{
--- a/cli/connect/const.go
+++ b/cli/connect/const.go
@@ -19,6 +19,9 @@ const setTableDialect = "\\set table_format"
 // width for tables.
 const setTableColumnWidthMaxLong = "\\set table_column_width"
 
+// setDelimiter set a custom expression delimiter for Tarantool console.
+const setDelimiter = "\\set delimiter"
+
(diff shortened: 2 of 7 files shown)


cli/connect/input_test.go
```diff
@@ -178,7 +178,71 @@ func TestAddStmtPart_luaValidator(t *testing.T) {
...
 		var completed bool
-		stmt, completed = AddStmtPart(stmt, part.str, validator)
+		stmt, completed = AddStmtPart(stmt, part.str, "", validator)
+
+		assert.Equal(t, part.expected, stmt)
+		assert.Equal(t, part.completed, completed)
+	}
+}
+
+func TestAddStmtPart_luaValidator_Delim(t *testing.T) {
```
oleg-jukovec, line 188:
```suggestion
func TestAddStmtPart_luaValidator_Delimiter(t *testing.T) {
```

oleg-jukovec (changes requested):
Please, add some integration tests to make sure that everything work fine. See:

https://github.com/tarantool/tt/blob/a46b1e2c1b626ac8d55dd0571b90ac52f89dc6c2/test/integration/connect/test_connect.py#L323-L325
and
https://github.com/tarantool/tt/blob/master/test/integration/connect/test_connect.py

dmyger:
> Please, add some integration tests

Of course, I remember about them.

cli/connect/commands.go
```diff
@@ -361,6 +361,19 @@ func setTableColumnWidthMaxFunc(console *Console,
...
+func setDelimiterMarker(console *Console, cmd string, args []string) (string, error) {
+	switch len(args) {
+	case 0:
+		console.delimiter = ""
+	case 1:
+		console.delimiter = args[0]
+	default:
+		return "", fmt.Errorf("the command expects zero or single argument")
+	}
+	return "---\n...\n", nil
```
oleg-jukovec, line 374: All `tt` commands return an empty output. So I suggest to follow the pattern since it does not break a table output:
```
localhost:3013> \set language sql
localhost:3013> \set delimiter ;
---
...

localhost:3013> \set output ttable
localhost:3013> \set delimiter ;
---
...
```

```suggestion
	return "", nil
```

cli/connect/commands.go
```diff
@@ -453,6 +466,14 @@ var cmdInfos = []cmdInfo{
 			),
 		),
 	},
+	cmdInfo{
+		Short: setDelimiter + " [marker]",
```
oleg-jukovec, line 470:
```
> \help
---
- - |
    To get help, see the Tarantool manual at https://tarantool.io/en/doc/
    To start the interactive Tarantool tutorial, type 'tutorial()' here.

    Available backslash commands:

      \set language <language>   -- set language (lua or sql)
      \set output <format>       -- set output format (lua[,line|block] or yaml)
      \set delimiter <delimiter> -- set expression delimiter
      \help                      -- show this screen
      \quit                      -- quit interactive console
...
```

```suggestion
		Short: setDelimiter + " <marker>",
```

cli/connect/commands.go
```diff
@@ -453,6 +466,14 @@ var cmdInfos = []cmdInfo{
...
+		Long:  "set a custom end-of-request marker",
```
oleg-jukovec, line 471:
```suggestion
		Long:  "set expression delimiter",
```

cli/connect/const.go
```diff
@@ -19,6 +19,9 @@ const setTableDialect = "\\set table_format"
 // width for tables.
 const setTableColumnWidthMaxLong = "\\set table_column_width"
 
+// setDelimiter set a custom end-of-request marker for Tarantool console.
```
oleg-jukovec, line 22:
```suggestion
// setDelimiter set a custom expression delimiter for Tarantool console.
```

oleg-jukovec (commented):
Thank you for the patch, just a few non-critical comments.

cli/connect/commands.go
dmyger: Done.

cli/connect/commands.go
dmyger: Done.
But I'm not sure it would be clear how it should be used.
Square brackets usually indicate that the argument is optional.
So an empty `<delimiter>` will turns off delimiter logic, here.

themilchenko (approved):
Thank you for the patch! See the improvement comment below.

cli/connect/commands.go
oleg-jukovec: The format used by Tarantool, so let it be.

oleg-jukovec (approved):

PR #961, opened 2024-09-26
ci: fix Python version for tests

dmyger:
Fixed CI environment setup with required Python version.

Follows up #727


oleg-jukovec (approved):

oleg-jukovec:
Do we need the same change in https://github.com/tarantool/tt/blob/master/.github/actions/prepare-ee-test-env/action.yml ?

themilchenko (approved):

dmyger:
> Do we need the same change in

It seems as it uses previously installed Python with right version. But, I add here, too. Just in case if tests order could be changed.

PR #979, opened 2024-10-10
reduce requirements for stop/kill/clean/logrotate

### LINKED ISSUE #880: `tt stop` does not work if 3.* configuration is missing or invalid
For instance, configuration storage server can be stopped, or local config may be invalid. This does allow `tt stop` command to stop running instances. But the `instances.yaml` file is sufficient for `tt stop` to carry out its tasks.
It also blocks stopping other applications, because `tt stop` exits after "broken" app is found. But it should just skip such apps.

elhimov:
For multi-instance applications no need to have instances scripts.
For tarantool3-based applications no need to have cluster config.

Closes #880


oleg-jukovec:
Please, rebase to the master branch & fix tests.

themilchenko (commented):
1. Please, can you add more information into a commit [test: improve test coverage](https://github.com/tarantool/tt/pull/979/commits/250755c35c8d975bbcfd801ce75643f3d6c54d5f) message with info about what was done or maybe split it on the several?

2. Is it possible to add an integration test on the fix for [reduce requirements for stop/kill/clean/logrotate](https://github.com/tarantool/tt/pull/979/commits/4d153929e2cf4f6c03db5ef599b0bddaa7ee0462)?

elhimov:
> 1. Please, can you add more information into a commit [test: improve test coverage](https://github.com/tarantool/tt/pull/979/commits/250755c35c8d975bbcfd801ce75643f3d6c54d5f) message with info about what was done or maybe split it on the several?
Sure, I'll add details.

> 2. Is it possible to add an integration test on the fix for [reduce requirements for stop/kill/clean/logrotate](https://github.com/tarantool/tt/pull/979/commits/4d153929e2cf4f6c03db5ef599b0bddaa7ee0462)?
Well, in fact the requested tests are already added as a part of the previous "test" commit (within this PR). I used this approach to be able to see the FAILED-to-PASSED transition of these tests in a more convenient way while implementing the desired functionality. OK, I'll rearrange tests to put new-functionality-tests into the "functionality" commit.

test/utils.py
```diff
@@ -217,6 +218,226 @@ def all_files_exist():
...
+        #         shutil.copytree(app_path, tmpdir)
+        #         app_name = os.path.basename(app_path)
+        #     else:
+        #         print("shutil.copy")
+        #         shutil.copy(app_path, tmpdir)
+        #         app_name = os.path.splitext(app_path)[0]
+        #     print(f"app_path={app_path}")
+        #     print(f"app_name={app_name}")
+        #     print(f"tmpdir={tmpdir}")
+        #     print("^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^")
```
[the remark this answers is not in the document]
elhimov: Done.

test/integration/restart/test_restart.py
```diff
@@ -94,3 +104,160 @@ def test_restart_no_args(tt_cmd, tmp_path):
...
+    pytest.param([], id="running:none"),
+    pytest.param([TtApp.app_target], id="running:all"),
+])
+@pytest.mark.parametrize("input", [
+    TtApp.Input("y\n", True),  # confirm (lowercase)
+    TtApp.Input("Y\n", True),  # confirm (uppercase)
+    TtApp.Input("nn\nny\ny\n", True),  # confirm preceded with the wrong answers
+    TtApp.Input("n\n", False),  # discard (lowercase)
+    TtApp.Input("N\n", False),  # discard (uppercase)
+    TtApp.Input("yy\nyn\nn\n", False),  # discard preceded with the wrong answers
```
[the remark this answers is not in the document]
elhimov: Done.

test/integration/restart/test_restart.py
```diff
@@ -94,3 +104,160 @@ def test_restart_no_args(tt_cmd, tmp_path):
...
```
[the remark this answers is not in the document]
elhimov: Done.

test/integration/restart/test_restart.py
```diff
@@ -94,3 +104,160 @@ def test_restart_no_args(tt_cmd, tmp_path):
...
+                    inst_names=["router", "master", "replica", "stateboard"])
+@pytest.mark.parametrize("running_targets", [
+    pytest.param([], id="running:none"),
+    pytest.param([TtApp.app_target], id="running:all"),
+    pytest.param([TtApp.Target("master")], id="running:master"),
+    pytest.param([TtApp.Target("master"), TtApp.Target("router")], id="running:master,router"),
+])
+@pytest.mark.parametrize("extra_setup", [
+    None,
+    # post_start_make_identical_subdir,
```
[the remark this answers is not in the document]
elhimov: Removed.

test/integration/kill/test_kill.py
```diff
@@ -0,0 +1,170 @@
...
+    pytest.param([], id="running:none"),
+    pytest.param([TtApp.app_target], id="running:all"),
+])
+@pytest.mark.parametrize("input", [
+    TtApp.Input("y\n", True),  # confirm (lowercase)
+    TtApp.Input("Y\n", True),  # confirm (uppercase)
+    TtApp.Input("nn\nny\ny\n", True),  # confirm preceded with the wrong answers
+    TtApp.Input("n\n", False),  # discard (lowercase)
+    TtApp.Input("N\n", False),  # discard (uppercase)
+    TtApp.Input("yy\nyn\nn\n", False),  # discard preceded with the wrong answers
```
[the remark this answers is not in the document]
elhimov: Done.

test/integration/clean/test_clean.py
```diff
@@ -0,0 +1,133 @@
...
+@pytest.mark.skipif(platform.system() == "Darwin", reason="Tests are failed by timeout")
+@pytest.mark.skipif(True, reason="Parsing of multiple answers from stdin is not supported.")
+@pytest.mark.tt_app(app_name="multi_inst_app",
+                    inst_names=["router", "master", "replica", "stateboard"])
+@pytest.mark.parametrize("input", [
+    TtApp.Input("y\ny\nY\nY\n", [True, True, True, True]),  # confirm all
+    TtApp.Input("n\nn\nN\nN\n", [False, False, False, False]),  # discard all
+    TtApp.Input("y\nn\nN\nY\n", [True, False, False, True]),  # mix #1
+    TtApp.Input("n\ny\nY\nN\n", [False, True, True, False]),  # mix #2
+    TtApp.Input("ny\ny\nyn\nn\nY\nN\n", [True, False, True, False]),  # mix #3
```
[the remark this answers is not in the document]
elhimov: Removed as multiple answers test is not supported currently.

test/integration/clean/test_clean.py
```diff
@@ -0,0 +1,133 @@
...
+            assert inst_id not in out
+
+
+# Multi-instance with auto-confirmation (short option).
+@pytest.mark.skipif(platform.system() == "Darwin", reason="Tests are failed by timeout")
+@pytest.mark.tt_app(app_name="multi_inst_data_app",
+                    inst_names=["router", "master", "replica", "stateboard"])
+@pytest.mark.parametrize("extra_setup", [
+    None,
+    # post_start_make_identical_subdir,
```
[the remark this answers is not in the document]
elhimov: Done.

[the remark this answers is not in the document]
elhimov: Done.

test/utils.py
```diff
@@ -217,6 +219,185 @@ def all_files_exist():
...
+
+    def clean(self, target=app_target, **kwargs):
+        return self.__run_tt("clean", target, **kwargs)
+
+    def logrotate(self, target=app_target):
+        return self.__run_tt("logrotate", target)
+
+
+# Values to be used with pytest.mark.parametrize to check various input at confirmation prompt.
+tt_app_params_confirmation = [
```
dmyger, line 328: 1) By the name of array is not clear what this array should do, and how it will be working. Seems a bit complex to support those tests.
2) `utils.py` does not seem the best place to keep tests arguments.

test/utils.py
```diff
@@ -217,6 +219,185 @@ def all_files_exist():
     return wait_event(timeout, all_files_exist, interval)
 
 
+class TtApp(object):
+    # Helper type to represent target. Commands (start, stop, etc.) treat it as following:
+    # None - no target at all (ex: tt start)
+    # Target(None) - app name only (ex: tt start some_app)
+    # Target('some_inst') - app:inst target (ex: tt start some_app:some_inst)
+    Target = namedtuple('Target', ['inst'])
```
dmyger, line 227: What are the benefits to use `namedtuple` construction? Especially for single field tuple.

test/utils.py
```diff
@@ -217,6 +219,185 @@ def all_files_exist():
...
+    @classmethod
+    def is_instance_of(cls, inst, *targets):
+        for target in targets:
+            assert target is None or isinstance(target, cls.Target)
+            if target is None or target.inst is None or target.inst == inst:
+                return True
+        return False
+
+    # Helper type to handle user input
+    Input = namedtuple('Input', ['str', 'is_confirm'])
```
dmyger, line 240: Is it required to put this type inside the class?
What kind of dependencies do we have here?
Maybe it is better to make such a construction independent?
What makes `namedtuple` better than a conventional dictionary like `dict[str, bool]`?

test/utils.py
elhimov: I've rearranged tests and dropped `Target`.

test/utils.py
elhimov: I've rearranged tests and dropped `Input`.

elhimov: Moved to separate file.

test/utils.py
elhimov: 1. Move it closer to where it's used. Hope along with comment it is more clear of how its working. It's just various values for one of test parameter that is passed to test via `pytest.mark.parametrize` machinery, so I see no problem to support those tests.
2. The same parameter set is used in multiple places, so initially I put it here just to avoid duplication. However, I'm agree it's bad place for it, so finally I just duplicate it where it's used.

cli/running/running.go
```diff
@@ -362,29 +348,41 @@ func loadInstanceConfig(configPath, instName string,
...
-	instances := []InstanceCtx{}
+	// log.Debugf("Collecting instances from application directory %q", appDir)
 	if !util.IsDir(appDir) {
-		return instances, fmt.Errorf("%q doesn't exist or not a directory", appDir)
+		return nil, fmt.Errorf("%q doesn't exist or not a directory", appDir)
 	}
 
 	appDirFiles, err := collectAppDirFiles(appDir)
+	// log.Errorf("appDirFiles: %v", appDirFiles)
+	// log.Error("^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^")
```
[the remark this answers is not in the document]
elhimov: Done.

cli/running/running.go
```diff
@@ -362,29 +348,41 @@ func loadInstanceConfig(configPath, instName string,
...
+				"cluster config %q is found, but instances config (instances.yml) is missing",
+				appDirFiles.clusterCfgPath)
+		} else {
+			if appDirFiles.defaultLuaPath != "" {
+				return []InstanceCtx{{
+					InstanceScript: appDirFiles.defaultLuaPath,
+					AppName:        filepath.Base(appDir),
+					InstName:       filepath.Base(appDir),
+					AppDir:         appDir,
+					SingleApp:      true}}, nil
```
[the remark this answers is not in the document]
elhimov: Not the change of this request. I just tabulated the original strings here.

test/ttapp.py
```diff
@@ -140,3 +143,13 @@ def wait_data_files(tt_app, instances):
 
 def make_identical_subdir(tt_app):
     os.mkdir(tt_app.app_path(os.path.basename(tt_app.app_dir)))
+
+
+def remove_app_script(tt_app):
+    time.sleep(0.5)
```
themilchenko, line 149: Why do we need this sleep with magic number?

test/ttapp.py
themilchenko, line 1: Maybe it is better to name it like `tt_helper` like it is already exist for Cartridge: `cartridge_helper`?

test/ttapp.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def wait_data_files(tt_app, instances):
+    files = []
+    for inst in instances:
+        if "router" in inst:
+            continue
+        files.append(tt_app.log_path(inst, utils.log_file))
+        files.append(tt_app.lib_path(inst, '*.snap'))
+        # files.append(tt_app.lib_path(inst, '*.xlog'))
```
themilchenko, line 140: Do we need this comment?

test/integration/stop/test_stop.py
```diff
@@ -4,10 +4,9 @@
 import pytest
 import ttapp
 
-from utils import (get_tarantool_version, pid_file, run_command_and_get_output,
-                   run_path, wait_file)
+import utils
```
themilchenko: Why do we need this change?

elhimov: In my opinion it's more convenient, because once you need any function from utils you just use it, no need to add the corresponding item into `from utils import ...` line (I see no problem to write `utils.<some_func>` instead of just `<some_func>`, moreover for me the former even looks better because it explicitly says that this function is external). Also after modifying `from util import ...` you would need to make sure it passes flaky:isort verification.
Besides I'm using this approach (I mean `import utils`) in the similar `test_*.py` modules that was added (not modified) in this PR.

test/ttapp.py
elhimov: The quick-and-dirty way to make sure the script is removed after it have been read. The better way is using the file that is created during the instance script execution as an indicator that script have been read already. I considered the latter initially but forgot to switch to it finally.

test/ttapp.py
elhimov: Forget to uncomment, thanks!

test/integration/clean/test_clean.py
```diff
@@ -0,0 +1,157 @@
...
+
+import pytest
+import ttapp
+
+import utils
+
+tarantool_major_version, tarantool_minor_version = utils.get_tarantool_version()
+
+skip_cluster_cond = tarantool_major_version < 3
+skip_cluster_reason = "skip cluster instances test for Tarantool < 3"
```
themilchenko, line 12: We already have this condition in function in `test/utils.py`:

https://github.com/tarantool/tt/blob/6a9c221cbac8f909a17e1595498b79ff6e33b1fd/test/utils.py#L552-L554

test/integration/stop/test_stop.py
themilchenko: Ok, fine.

test/integration/stop/test_stop.py
```diff
@@ -126,3 +145,161 @@ def test_stop_no_prompt(tt_cmd, tmpdir_with_cfg):
     finally:
         stop_cmd = [tt_cmd, "stop", "-y", app_name]
         run_command_and_get_output(stop_cmd, cwd=tmpdir_with_cfg)
+
+
+def check_stop(tt_app, running_targets, post_start, target, input, is_confirm, *args):
+    print(f"do_check: args: {args} target={target} input={input}")
```
themilchenko, line 72: Is it a debug print? Do we need it?

test/integration/stop/test_stop.py
```diff
@@ -126,3 +145,161 @@ def test_stop_no_prompt(tt_cmd, tmpdir_with_cfg):
...
+    # Prepare the application considering the test parameters.
+    orig_status = ttapp.setup(tt_app, running_targets, post_start)
+
+    # Do stop.
+    print(f"do_check: stop: target={target} args: {args}")
```
themilchenko, line 77: Here too.

test/integration/kill/test_kill.py
```diff
@@ -0,0 +1,187 @@
...
+
+import utils
+
+tarantool_major_version, tarantool_minor_version = utils.get_tarantool_version()
+
+skip_cluster_cond = tarantool_major_version < 3
+skip_cluster_reason = "skip cluster instances test for Tarantool < 3"
+
+# Values to be used to parametrize input at confirmation prompt.
+confirmation_input_params = [
```
dmyger, line 15: And why isn't enough just one test that checks all these options, why do you need to repeat the same for each test?

test/integration/clean/test_clean.py
```diff
@@ -0,0 +1,157 @@
...
+@pytest.mark.tt_app(**mark_kwargs_multi_inst)
+@pytest.mark.parametrize('running_targets, stop_targets', [
+    pytest.param([], [], id="running:none/none"),
+    pytest.param(['app'], [], id="running:all/none"),
+    pytest.param(['app'], ['app'], id="running:all/all"),
+    pytest.param(['app:master'], ['app:master'], id="running:master/master"),
+    pytest.param(['app'], ['app:master'], id="running:all/master"),
+    pytest.param(['app'], ['app:master', 'app:router'], id="running:all/master,router"),
+])
+@pytest.mark.parametrize("extra_setup", [
```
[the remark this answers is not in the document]
elhimov: I'm not sure I understand your concern. In any test we have "input" and "output" (in a broad sense). Here the presence of the instance script is a part of the test "input" and "output" should not depend on its existence.

> If nothing changed, why it matters to test?

Considering the above this is incorrect statement, because "input" is changed.

test/integration/kill/test_kill.py
elhimov: > Note: On CI the duration of such a sleep is very flaky.

Agree. I'm working to drop it.

test/ttapp.py
```diff
@@ -0,0 +1,144 @@
...
+    wait_pid_files(tt_app, instances)
+    wait_box_status(tt_app, instances, ['loading', 'running'])
+
+
+def wait_log_files(tt_app, instances):
+    files = list(map(lambda x: tt_app.log_path(x, utils.log_file), instances))
+    assert utils.wait_files(5, files)
+
+
+def make_identical_subdir(tt_app):
```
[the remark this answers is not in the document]
elhimov: No, it's not. Will be removed.

test/integration/kill/test_kill.py
elhimov: > And why isn't enough just one test that checks all these options, why do you need to repeat the same for each test?

There are 2 parameterized tests: against multi-instance app and against cluster app. Each of them checks various combinations of "input" parameters, such as initially running instances and targets kill is requested for. What does mean "for each test"? Please, clarify.

test/integration/stop/test_stop.py
elhimov: Missed it. Thanks!

test/integration/clean/test_clean.py
elhimov: It seems I just copy-pasted the wrong approach. I'll switch to this one.

test/ttapp.py
```diff
@@ -0,0 +1,144 @@
...
+    def clean(self, *args, **kwargs):
+        return self.exec('clean', *args, **kwargs)
+
+    def logrotate(self, *args):
+        return self.exec('logrotate', *args)
+
+    def status(self, *args):
+        rc, out = self.exec('status', *args)
+        assert rc == 0
+        return utils.extract_status(out)
```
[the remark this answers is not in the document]
elhimov: Some time ago there was sense, but now it's gone indeed. Thanks, for pointing out! Will be fixed.

test/integration/clean/test_clean.py
```diff
@@ -0,0 +1,157 @@
...
+
+import utils
+
+tarantool_major_version, tarantool_minor_version = utils.get_tarantool_version()
+
+skip_cluster_cond = tarantool_major_version < 3
+skip_cluster_reason = "skip cluster instances test for Tarantool < 3"
+
+
+def check_clean(tt_app, running_targets, stop_targets, extra_setup, target, *args):
```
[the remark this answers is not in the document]
elhimov: Well... I agree that over-something is bad in general. The question is where we are? Already "over" or still "under"? And what is the alternative? How it affects the rest of context?

This function checks the correctness of `tt clean`. That expected prompt/messages/warnings are displayed. That affected only requested instances. This is _the essential part_ of `tt clean` functionality that must be checked at any execution of `tt clean`, so it looks quite reasonable to have such a function. The same applied to other-universal-functions.

Q: How we were living w/o it?
A: We had just a single test that checks just a couple of `tt clean` scenario (and frankly speaking checking was oversimplified)

This PR significantly extends the coverage of such instance-commands like start/stop/restart/kill/clean/logrotate by checking various combinations of initial states (which instances are running at the moment command is executed) and command target (target is omitted, app-target, instance-target). W/o this function (and similar for other commands) we would need another "over-" -- over-duplication which is much worse in terms of support and development as for me. 

One more note specific to clean-universal-function: in addition to other-universal-functions parameters (that specify already running instances and command target) in case of 'clean' we need separate parameter `stop_targets` because we need to differentiate instances that were running and then stopped and instances that haven't been started at all. One may find this part where initial setup is done a little tricky but the main part (as well as other-universal-functions) is quite simple and straightforward.

test/ttapp.py
elhimov: That was temporary solution. I'm working to drop it.

test/ttapp.py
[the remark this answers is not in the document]
elhimov: Done.

test/integration/clean/test_clean.py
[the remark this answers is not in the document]
elhimov: To improve readability code that performs initial setup was separated from the checking itself.

test/ttapp.py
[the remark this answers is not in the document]
elhimov: Done

test/integration/stop/test_stop.py
elhimov: Done

test/integration/stop/test_stop.py
elhimov: Done

test/integration/clean/test_clean.py
elhimov: Done

test/integration/kill/test_kill.py
dmyger: I see this variable mentioned 11 times. In my mind, this functionality should be checked once in a corresponding test that checks the logic of entering an answer, not its result.

test/integration/clean/test_clean.py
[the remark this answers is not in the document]
dmyger: > the instance script is a part of the test "input" and "output" should not depend on its existence

That's what I find so strange: changes in inputs don't lead to changes in outputs.

dmyger (approved):

test/integration/kill/test_kill.py
elhimov: > I see this variable mentioned 11 times. In my mind, this functionality should be checked once in a corresponding test

`tt` implements multiple commands and some of them in some circumstances require user confirmation. This PR introduces tests that check various scenarios with 3 commands: restart, stop, kill. Various scenarios here means various combinations of different application type (simple app, multi-instance app, cluster app) with various user answers. These scenarios are distributed among tests based on command, then on application type. I used this kind of distribution based on current test layout that groups tests by commands. Do you suggest to rearrange these tests so single test checks user interaction against all command and application types? If so then what's the benefit of such arrangement? It may eliminate multiple references to this parameter set at a price of another parameter set duplication (like set representing collection of commands or set representing collection of application types).

> checks the logic of entering an answer, not its result

Sorry, still don't understand. What does it mean? What is "the logic of entering an answer" and how it should be checked? And what does mean "not its result"? What is the result? Is it the way execution goes after confirmation or discarding? If so and it should not be used, then how "the logic of entering an answer" should be checked?

It seems there is some key difference in conception regarding these tests, so I just can't understand your point. Would you please explain in other words.

My vision is scenario-based:
There are tons of scenarios `tt` might be used and we need to cover as much as we can with tests (considering CI limitations). This PR introduces tests that cover considerable amount of scenarios for 6 commands: start/stop/restart/kill/logrotate/clean. Arrangement of tests based on current hierarchy, i.e. at first level tests are grouped by the `tt` command they check.

test/integration/clean/test_clean.py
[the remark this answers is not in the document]
elhimov: > That's what I find so strange: changes in inputs don't lead to changes in outputs.

It was the goal of this PR. Previously such change in input did lead to dramatic changes in output because every command mentioned in PR title failed w/o instance script (in case of multi-inst app).

themilchenko:
Some checks on CI has failed.

test/ttapp.py
themilchenko: `sleep` still exists.

test/ttapp.py
themilchenko: What do you think about it?

test/ttapp.py
elhimov: Yes, it was the last one to fix. Now it's done.

test/ttapp.py
elhimov: Done

themilchenko (approved):
Thank you!

PR #980, opened 2024-10-11
cd: add Ubuntu Noble

Totktonada:
We have a `tt` package in the `linux-deb/static` repository and it is suitable for Ubuntu Noble as well. However, sometimes it is convenient to have the same focal/jammy/noble setup, where the `ubuntu/${codename}` repository is enabled.


oleg-jukovec (approved):
Thank you!

PR #985, opened 2024-10-15
pack: added .packignore

Japsty:
commit template


cli/pack/common.go
```diff
@@ -277,14 +277,30 @@ func getDestAppDir(bundleEnvPath, appName string,
 
 // copyApplications copies applications from current env to the result bundle.
 func copyApplications(bundleEnvPath string, packCtx *PackCtx,
-	cliOpts, newOpts *config.CliOpts) error {
-	var err error
+	cliOpts, newOpts *config.CliOpts, toIgnore map[string]struct{}) error {
```
psergee, line 280: The name of this function is "copy applications". One of the arguments is a set of paths to do not copy. Do we exclude the same paths for all applications? If so, this is not what was expected.

Japsty: Changed the implementation, now files are deleted from the directory after full copying by a separate function 'removeIgnoredFiles'

cli/pack/testdata/bundle_with_packignore/instances_enabled
```diff
@@ -0,0 +1 @@
+../../../../test/integration/pack/test_bundles/bundle1/instances_enabled
```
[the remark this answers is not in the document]
Japsty: These files are for the new unit test in "common_test.go"

cli/pack/ignore.go
```diff
@@ -0,0 +1,100 @@
...
+		ignore, err := shouldIgnore(relPathUnix, patterns)
+		if err != nil {
+			return err
+		}
+
+		if ignore {
+			if info.IsDir() {
+				err = os.RemoveAll(path)
+				if err != nil {
+					return fmt.Errorf("failed to remove directory %q: %v", path, err)
```
psergee, line 88:
```suggestion
					return fmt.Errorf("failed to remove directory %q: %s", path, err)
```

cli/pack/ignore.go
```diff
@@ -0,0 +1,100 @@
...
+				}
+				return filepath.SkipDir
+			} else {
+				err = os.Remove(path)
+				if err != nil {
+					return fmt.Errorf("failed to remove file %q: %v", path, err)
```
psergee, line 94:
```suggestion
					return fmt.Errorf("failed to remove file %q: %s", path, err)
```

cli/pack/ignore.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// shouldIgnore checks if the given file path matches any of the ignore patterns.
+func shouldIgnore(path string, patterns map[string]struct{}) (bool, error) {
+	for pattern := range patterns {
+		pattern = filepath.ToSlash(pattern)
+		filePath := filepath.ToSlash(path)
+
+		if strings.HasSuffix(pattern, "/") {
+			if strings.HasPrefix(filePath, pattern) {
```
psergee, line 47: How is it possible for filePath, which is a relative path, to have a prefix with first '/'?
It would be nice to have unit-tests for this function.

test/integration/pack/test_bundles/bundle_with_packignore/instances_enabled/app2/.packignore
```diff
@@ -0,0 +1,4 @@
+var/lib/inst1
+var/log
+var/run/inst2/app2.pid
```
psergee, line 3: run_dir should not be copied, even without the .packignore file.

cli/pack/common.go
```diff
@@ -285,6 +285,17 @@ func copyApplications(bundleEnvPath string, packCtx *PackCtx,
 		}
 		inst := instances[0]
 		appPath := inst.AppDir
+		if !inst.IsFileApp {
+			ignorePatterns, err := readPackIgnore(appPath)
+			if err != nil {
+				return fmt.Errorf("failed to read .packignore: %v", err)
+			}
+
+			if err = removeIgnoredFiles(appPath, ignorePatterns); err != nil {
```
psergee, line 294: Looks like we remove files before copying the application. This approach will surprise the user.
We should skip copying such files, not removing them from the source directory.

cli/pack/common_test.go
```diff
@@ -1068,6 +1068,26 @@ func Test_prepareBundle(t *testing.T) {
 				{assert.FileExists, "app/tt.yaml"},
 			},
 		},
+		{
+			name: "Packing env with packignore",
```
psergee, line 1072: There are no tests for ignoring app files/dirs.

Issue #900, opened 2024-07-10
`tt uninstall` does not work if no symlink in `bin_dir`


psergee:
How to reproduce:
```
$ ls -l bin/
total 22512
-rwxr-xr-x. 1 user user 23048344 Jul 10 16:44 tt_94ba971

$ tt uninstall tt 94ba971
   • Removing binary...       
   ⨯ failed to resolve symlink /home/user/tmp/bin/tt: lstat /home/user/tmp/bin/tt: no such file or directory
```

Japsty:
I would like to work on this issue. Could you please assign it to me?

PR #997, opened 2024-10-31
uninstall: fix uninstalling  if no symlink in bin_dir

Japsty:
tt uninstall does not work if there is no symbolic link in bin_dir, so a check for the presence of a symbolic link has been added.

Closes: [900](https://github.com/tarantool/tt/issues/900)

### DIFF
--- a/cli/uninstall/uninstall.go
+++ b/cli/uninstall/uninstall.go
@@ -61,19 +61,29 @@ func remove(program string, programVersion string, directory string,
 	} else if err != nil {
 		return false, fmt.Errorf("there was some problem locating %s", path)
 	}
-	// Get path where symlink point.
-	resolvedPath, err := util.ResolveSymlink(linkPath)
-	if err != nil {
-		return false, fmt.Errorf("failed to resolve symlink %s: %s", linkPath, err)
-	}
+
 	var isSymlinkRemoved bool
-	// Remove symlink if it points to program.
-	if strings.Contains(resolvedPath, fileName) {
-		err = os.Remove(linkPath)
-		if err != nil {
-			return false, err
+
+	_, err = os.Lstat(linkPath)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return false, fmt.Errorf("failed to access %q: %s", linkPath, err)
 		}
-		isSymlinkRemoved = true
+		isSymlinkRemoved = false
+	} else {
+		// Get path where symlink point.
+		resolvedPath, err := util.ResolveSymlink(linkPath)
+		if err != nil {
+			return false, fmt.Errorf("failed to resolve symlink %q: %s", linkPath, err)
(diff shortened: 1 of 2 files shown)


cli/uninstall/uninstall.go
```diff
@@ -61,19 +61,30 @@ func remove(program string, programVersion string, directory string,
...
+
 	var isSymlinkRemoved bool
-	// Remove symlink if it points to program.
-	if strings.Contains(resolvedPath, fileName) {
-		err = os.Remove(linkPath)
+
+	_, err = os.Lstat(linkPath)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return false, fmt.Errorf("failed to access %s: %s", linkPath, err)
```
psergee, line 70:
```suggestion
			return false, fmt.Errorf("failed to access %q: %s", linkPath, err)
```

cli/uninstall/uninstall.go
```diff
@@ -61,19 +61,30 @@ func remove(program string, programVersion string, directory string,
...
+		}
+		isSymlinkRemoved = false
+	} else {
+		// Get path where symlink point.
+		resolvedPath, err := util.ResolveSymlink(linkPath)
 		if err != nil {
-			return false, err
+			return false, fmt.Errorf("failed to resolve symlink %s: %s", linkPath, err)
```
psergee, line 77:
```suggestion
			return false, fmt.Errorf("failed to resolve symlink %q: %s", linkPath, err)
```

cli/uninstall/uninstall.go
```diff
@@ -61,19 +61,30 @@ func remove(program string, programVersion string, directory string,
...
+		}
+
+		// Remove symlink if it points to program.
+		if strings.Contains(resolvedPath, fileName) {
+			err = os.Remove(linkPath)
```
psergee, line 82:
```suggestion
			if err = os.Remove(linkPath); err != nil {
```

cli/uninstall/uninstall.go
```diff
@@ -61,19 +61,30 @@ func remove(program string, programVersion string, directory string,
...
-	// Remove symlink if it points to program.
-	if strings.Contains(resolvedPath, fileName) {
-		err = os.Remove(linkPath)
+
+	_, err = os.Lstat(linkPath)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return false, fmt.Errorf("failed to access %s: %s", linkPath, err)
+		}
+		isSymlinkRemoved = false
```
psergee, line 72: Looks like, if there is no symlink, we report like it is not removed. Please add a test to check the message in case of missing symlink, but existing binary.

psergee (commented):
Thank you for the patch!

psergee (approved):
lgtm

PR #998, opened 2024-11-06
deps: bump go-openssl to v1.1.1

oleg-jukovec:
The latest version fixes a svacer issues.

### DIFF
--- a/.github/workflows/full-ci.yml
+++ b/.github/workflows/full-ci.yml
@@ -24,7 +24,7 @@ jobs:
       (github.event_name == 'pull_request' &&
         github.event.action == 'synchronize' &&
         contains(github.event.pull_request.labels.*.name, 'full-ci'))
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     strategy:
       matrix:
         tarantool-version: ["1.10", "2.10", "3.0"]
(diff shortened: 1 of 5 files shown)


dmyger (approved):
Won't there be a problem with 22 Ubuntu like there used to be?

oleg-jukovec:
> Won't there be a problem with 22 Ubuntu like there used to be?

I think we can wait for problems to solve it. `tt-ee` is testing and building under ubuntu-22.04 and all is fine.

DerekBum (approved):
What about this file? Is it used in those workflows? Don't we need to update Ubuntu version here as well?
https://github.com/tarantool/tt/blob/bb933927617a81ef5ff6398ae05a6e2969fd2977/ci/Dockerfile#L1-L6

oleg-jukovec:
> What about this file? Is it used in those workflows? Don't we need to update Ubuntu version here as well?
> 
> https://github.com/tarantool/tt/blob/bb933927617a81ef5ff6398ae05a6e2969fd2977/ci/Dockerfile#L1-L6

https://github.com/tarantool/tt/blob/bb933927617a81ef5ff6398ae05a6e2969fd2977/ci/Dockerfile#L63-L69

It installs OpenSSL 3.0.8 directly, so it should work fine.

oleg-jukovec:
Yeah, it works:

https://github.com/tarantool/tt/actions/runs/11719705206/job/32643535109
https://github.com/tarantool/tt/actions/runs/11719705206/job/32643535654

PR #1024, opened 2024-11-17
connect: add initial custom evaler support

psergee:
Add support for customizing the way user input is processed. The `--evaler` option is added to specify the Lua code, which accepts `cmd` var containing user input for current connection. The evaler Lua code can be loaded from file.
Autocompletion is disabled if custom evaler is set.


test/integration/connect/test_connect.py
```diff
@@ -3088,3 +3089,94 @@ def test_set_delimiter(
...
+                                                            "test_app.lua")
+    copy_data(tmpdir, [test_app_path])
+    start_app(tt_cmd, tmpdir, "test_app")
+
+    try:
+        file = wait_file(os.path.join(tmpdir, "test_app"), "ready", [])
+        assert file != ""
+
+        with open(os.path.join(tmpdir, "evaler.lua"), 'w') as f:
+            f.write("return box.execute(cmd)")
```
DerekBum, line 3154: Why do you need to create a correct file for error cases? Plus it's not used here. Maybe delete this lines, or check the error for correct filepath, but incorrect command in the file.

DerekBum (commented):
Thanks for the patch! Just couple of comments.

Don't we need to add more info about `tt connect --evaler` argument? To let users know that it can also work with lua scripts from file (and files should be passed as `@path/to/file`).

askalt (approved):

test/integration/connect/test_connect.py
psergee: This is left after test code copy-paste. Removed.

DerekBum (approved):

PR #1030, opened 2024-11-20
`tt replicaset upgrade`: support remote replica set

mandesero:
This patch adds the ability to update the schema on a remote replica set.

-------

### Example

Start vshard app.

<details>
<summary>config.yaml</summary>

```yaml
credentials:
  users:
    client:
      password: 'secret'
      roles: [super]
    replicator:
      password: 'secret'
      roles: [replication]
    storage:
      password: 'secret'
      roles: [sharding]

iproto:
  advertise:
    peer:
      login: replicator
    sharding:
      login: storage

sharding:
  bucket_count: 3000

groups:
  storages:
    app:
      module: storage
    sharding:
      roles: [storage]
    replication:
      failover: manual
    replicasets:
      storage-001:
        leader: storage-001-a
        instances:
          storage-001-a:
            iproto:
              listen:
                - uri: localhost:3301
          storage-001-b:
            iproto:
              listen:
                - uri: localhost:3302
      storage-002:
        leader: storage-002-a
        instances:
          storage-002-a:
            iproto:
              listen:
                - uri: localhost:3303
          storage-002-b:
            iproto:
              listen:
                - uri: localhost:3304
  routers:
    app:
      module: router
    sharding:
      roles: [router]
    replicasets:
      router-001:
        instances:
          router-001-a:
            iproto:
              listen:
                - uri: localhost:3305
```

</details>

To update the schema of this cluster, you need to update each replica set individually. Simply select one instance of each replica set and run:
```bash
$ tt replicates upgrade tcp://client:secret@127.0.0.1:3301
  • storage-001: ok
$ tt replicates upgrade tcp://client:secret@127.0.0.1:3304
  • storage-002: ok
$ tt replicates upgrade tcp://client:secret@127.0.0.1:3305
  • router-001: ok
``` 

------

Part of #968

Issue #1014, opened 2024-11-15
`modules:` prepare local modules list


dmyger:
Each entry of module list is treated as a directory with a collection of modules. Each module is located in its own separate subdirectory, one level down. Any duplicates of modules with the same name must be removed, and the original order of module declaration must be preserved (so the first module declaration is used, duplicates are ignored).

All directories of the first level within the specified directory of modules should be scanned, and within each of them a search for either the main executable file or the manifest.yaml file is performed. If one of them is found, the name of the directory will be the potential name of the command executed in `tt`.

The result is a filtered list of directories where found files meet the first requirement of a module.

Blocked with #1012 and #1013 - expecting the list of modules directories is pre-prepared for work.

oleg-jukovec:
Part of #1015

PR #1036, opened 2024-11-26
modules: find module with TT_CLI_MODULES_PATH

### LINKED ISSUE #1013: `modules:` add support find module with `TT_CLI_MODULES_PATH` variable
The `TT_CLI_MODULES_PATH` environment variable specifies a list of module paths separated by a colon `“:”`. The variable must be set in the `SDK` when using the `env.sh` configuration file. It can also be set independently and must contain a list of folders where modules are located. The logic of working with this variable should be similar to the `PATH` system variable.

Lists of modules are combined into one common list:
- The modules declared in the configuration file `tt.yaml` come first
- then the modules defined via the environment variable are added.

Any duplicates must be removed, while maintaining the original order of module declarations.

Blocked with #1012 here we are expecting a list of at `modules/directories` in the `tt.yaml`.

dmyger:
Closes #1013 

@TarantoolBot document
Title: Use environment variable to find modules.

The TT_CLI_MODULES_PATH environment variable specifies a list of
module's paths separated by a colon “:”.
It can be set externally from tt and must contain a list of
folders where modules are located.
The logic of working with this variable is similar to the PATH
system variable.

Lists of modules are combined into one common list:

- The modules declared in the configuration file tt.yaml come first
- then the modules defined via the environment variable are added.
Any duplicates must be removed, while maintaining the original
order of module declarations.

**Usage example**
Set variable:
```sh
export TT_CLI_MODULES_PATH=/ext/path/modules:${TT_CLI_MODULES_PATH}
```

### DIFF
--- a/cli/modules/modules_test.go
+++ b/cli/modules/modules_test.go
@@ -1,3 +1,5 @@
+// FIXME: Create new tests https://github.com/tarantool/tt/issues/1039
+
 package modules
 
 import (
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 - `tt.yaml`: allows to specify a list of modules directories.
+- Environment variable TT_CLI_MODULES_PATH can be used to specify
+  an extra path with modules.
 
 ### Changed
 
--- a/cli/modules/modules.go
+++ b/cli/modules/modules.go
@@ -46,12 +46,18 @@ func fillSubCommandsInfo(cmd *cobra.Command, modulesInfo *ModulesInfo) {
 // GetModulesInfo collects information about available modules (both external and internal).
 func GetModulesInfo(cmdCtx *cmdcontext.CmdCtx, rootCmd *cobra.Command,
 	cliOpts *config.CliOpts) (ModulesInfo, error) {
-	modulesDir, err := getExternalModulesDir(cmdCtx, cliOpts)
+	modulesDirs, err := getConfigModulesDirs(cmdCtx, cliOpts)
 	if err != nil {
 		return nil, err
 	}
 
-	externalModules, err := getExternalModules(modulesDir)
+	modulesEnvDirs, err := getEnvironmentModulesDirs()
+	if err != nil {
+		return nil, err
+	}
+	modulesDirs = append(modulesDirs, modulesEnvDirs...)
+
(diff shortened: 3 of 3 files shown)


cli/modules/modules_test.go
```diff
@@ -1,3 +1,6 @@
+// FIXME: Create new tests https://github.com/tarantool/tt/issues/1039
```
oleg-jukovec, line 1: Why not create new tests now?

dmyger: At the moment, we don't have any public method for unit tests. And no completed functionality for integration tests.

oleg-jukovec (approved):
Please, add a description to the commit messages, see:

https://github.com/tarantool/tt/pull/1030
https://github.com/tarantool/doc/issues/4661

It will help to notify a doc team about changes.

oleg-jukovec (approved):

cli/modules/modules_test.go
```diff
@@ -1,3 +1,6 @@
...
+//go:build exclude
```
oleg-jukovec, line 2: This is much worse. We can't break backward compatibility. So old tests should work.

dmyger: According to RFC we don't keep backward compatibility with old module's system.

oleg-jukovec: Anyway why does it break here if we add an optional environment variable that is not used in tests?

dmyger: Restore run this test.

oleg-jukovec (changes requested):
I think we could add an integration test that a module could be found in a path passed with the variable, not just in "modules". Could we?

dmyger:
> I think we could add an integration test

Yes, integration tests will need to be added. But they won't work at the current stage, as there is nothing to run to test yet. There is a task assigned to the tests #1039.

oleg-jukovec:
> > I think we could add an integration test
> 
> Yes, integration tests will need to be added. But they won't work at the current stage, as there is nothing to run to test yet. There is a task assigned to the tests #1039.

Why not? In theory, the directory should add a module path, so we could add a test in which modules (in the current implementation) will be located at the path. Or am I missed something?

dmyger:
> Or am I missed something?

Creating an integration test requires that there should be a difference in external behaviour in the operation of the utility, and this has not yet been implemented. 

Unit testing, in theory, allows for such verification. But we don't test implementation of private methods, while public methods don't have implementation until #1014 will be done.

oleg-jukovec (approved):

PR #932, opened 2024-08-31
generate: autoupdate

TarantoolBot:
Automated changes by [update](https://github.com/tarantool/tt/blob/master/.github/workflows/update.yml) job.

PR #1045, opened 2024-12-02
Revert "generate: autoupdate"

oleg-jukovec:
Reverts tarantool/tt#932


oleg-jukovec:
The problem is not in the commit, but somewhere in the tarantool 3.3 update.

PR #1056, opened 2024-12-06
test: fix check resolving Rocks path

dmyger:
The issues with the `TestGetRocksRepoPath` was when run Unit test with
applied Tarantool `env.sh` and `TT_CLI_REPO_ROCKS` environment variable.


patapenka-alexey (approved):
LGTM.

PR #1058, opened 2024-12-09
play: support app:instance

### LINKED ISSUE #1022: play: support app:instance
The `tt play` command should support connection to a target instance by `app:instance` as `tt connect` command.

patapenka-alexey:
This patch adds support connection to a target instance by `app:instance` for `tt play` command.

Closes #1022


cli/cmd/play.go
```diff
@@ -78,6 +79,29 @@ func internalPlayModule(cmdCtx *cmdcontext.CmdCtx, args []string) error {
...
+		}
+
+		_, err := os.Stat(runningCtx.Instances[0].BinaryPort)
+		if err != nil {
+			return util.InternalError(
+				"Internal error: application binary port does not exist: %s",
+				version.GetVersion, err)
+		}
+		args[0] = runningCtx.Instances[0].BinaryPort
+	}
```
oleg-jukovec, line 103: Please, do it in the connect manner. It will help us to have the same behavior for all modules:

https://github.com/tarantool/tt/blob/96db0688623bf0a8072b3c985eb20c3d1b5df617/cli/cmd/connect.go#L142-L187

```suggestion
	// FillCtx returns error if no instances found.
	var runningCtx running.RunningCtx
	var err error
	if err = running.FillCtx(cliOpts, cmdCtx, &runningCtx, args, false); err == nil {
		_, err := os.Stat(runningCtx.Instances[0].BinaryPort)
		if err != nil {
			return util.InternalError(
				"Internal error: application binary port does not exist: %s",
				version.GetVersion, err)
		}
		args[0] = runningCtx.Instances[0].BinaryPort
	} else if libconnect.IsCredentialsURI(args[0]) {
		// Fetch credentials from the URI.
	} else if libconnect.IsBaseURI(args[0]) {
		// Fetch credentials from ENV.
	} else {
		// Return wrapped err.
	}
```

patapenka-alexey: Updated.

patapenka-alexey: Updated.

cli/cmd/play.go
```diff
@@ -78,6 +79,43 @@ func internalPlayModule(cmdCtx *cmdcontext.CmdCtx, args []string) error {
...
+				version.GetVersion, err)
+		}
+
+		args[0] = runningCtx.Instances[0].BinaryPort
+
+	} else if libconnect.IsCredentialsURI(args[0]) {
+		if playUsername != "" || playPassword != "" {
+			return errors.New("username and password are specified with" +
+				" flags and a URI")
+		}
```
oleg-jukovec, line 105:
```suggestion
	} else if libconnect.IsCredentialsURI(args[0]) {
		if playUsername != "" || playPassword != "" {
			return errors.New("username and password are specified with" +
				" flags and a URI")
		}
		playUsername = ?
		playPassword = ?
```

cli/cmd/play.go
```diff
@@ -78,6 +79,43 @@ func internalPlayModule(cmdCtx *cmdcontext.CmdCtx, args []string) error {
...
+		if playUsername == "" {
+			playUsername = os.Getenv(libconnect.TarantoolUsernameEnv)
+		}
+		if playPassword == "" {
+			playPassword = os.Getenv(libconnect.TarantoolPasswordEnv)
+		}
+	} else {
+		return util.InternalError(
+			"Internal error: could not parse: %s",
+			version.GetVersion, args[0])
```
oleg-jukovec, line 116:
```suggestion
		return util.InternalError(
			"could not resolve URI or application: %s",
			version.GetVersion, args[0])
```

test/integration/play/test_play.py
```diff
@@ -1,14 +1,60 @@
...
+
+def is_cluster_app_supported(tt_cmd, tmpdir):
+    ok, major, minor, patch = get_version(tt_cmd, tmpdir)
+    assert ok
+    return major >= 3
+
+
+def skip_if_cluster_app_unsupported(tt_cmd, tmpdir):
+    if not is_cluster_app_supported(tt_cmd, tmpdir):
+        pytest.skip("Tarantool 3.0 or above required")
```
oleg-jukovec, line 52: Do we have helpers for it in utils? We should have it, or it should be added to a some common place.

test/integration/play/test_play.py
```diff
@@ -182,3 +228,116 @@ def test_play_creds(tt_cmd, tmp_path, opts, test_instance):
...
+        for instance in instances:
+            master_run_path = os.path.join(tmpdir, app_name, run_path, instance)
+            file = wait_file(master_run_path, control_socket, [])
+            assert file != ""
+            file = wait_file(master_run_path, BINARY_PORT_NAME, [])
+            assert file != ""
+            file = wait_file(os.path.join(tmpdir, app_name), 'configured', [])
+            assert file != ""
+
+        # Play to the instances
```
oleg-jukovec, line 330:
```suggestion
        # Play to the instances.
```

test/integration/play/test_play.py
```diff
@@ -182,3 +228,116 @@ def test_play_creds(tt_cmd, tmp_path, opts, test_instance):
...
+
+    cmd = [tt_cmd, "play", "remote_instance_cfg", test_xlog_name]
+    rc, cmd_output = run_command_and_get_output(cmd, cwd=test_instance._tmpdir)
+    assert rc == 1
+    assert "application binary port does not exist" in cmd_output
+
+
+def test_play_to_cluster_app(tt_cmd):
+    if platform.system() == "Darwin":
+        pytest.skip("/set platform is unsupported by test")
```
oleg-jukovec, line 298: We do not "/set" anything here.

test/integration/play/test_play.py
```diff
@@ -1,14 +1,60 @@
...
+        text=True
+    )
+    if instance_process.returncode == 0:
+        stdout = instance_process.stdout
+        full = stdout.splitlines()[0]
+        for word in re.split(r'\s', full):
+            matched = re.match(r'^\d+\.\d+\.\d+', word)
+            if matched:
+                print("Matched:")
+                print(matched)
```
oleg-jukovec, line 38: Please, remove the debug prints.
```suggestion
                print("Matched:")
                print(matched)
```

cli/running/running.go
```diff
@@ -680,7 +680,8 @@ func FillCtx(cliOpts *config.CliOpts, cmdCtx *cmdcontext.CmdCtx,
 	var err error
 
 	if len(args) > 1 && cmdCtx.CommandName != "run" && cmdCtx.CommandName != "connect" &&
-		cmdCtx.CommandName != "add" && cmdCtx.CommandName != "remove" {
+		cmdCtx.CommandName != "add" && cmdCtx.CommandName != "remove" &&
+		cmdCtx.CommandName != "play" {
```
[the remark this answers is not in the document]
patapenka-alexey: Or we should pass only `args[0]` as strings array to the function. Updated.

test/integration/play/test_play.py
patapenka-alexey: Used directly from `connect.py`.

oleg-jukovec: It's better to move it into utils.py than.

cli/cmd/play.go
```diff
@@ -78,6 +79,45 @@ func internalPlayModule(cmdCtx *cmdcontext.CmdCtx, args []string) error {
...
+		_, err := os.Stat(runningCtx.Instances[0].BinaryPort)
+		if err != nil {
+			return util.InternalError(
+				"Internal error: application binary port does not exist: %s",
+				version.GetVersion, err)
+		}
+
+		args[0] = runningCtx.Instances[0].BinaryPort
+
+	} else if libconnect.IsCredentialsURI(args[0]) {
```
oleg-jukovec, line 101:
```suggestion
		args[0] = runningCtx.Instances[0].BinaryPort
	} else if libconnect.IsCredentialsURI(args[0]) {
```

cli/cmd/play.go
```diff
@@ -78,6 +79,45 @@ func internalPlayModule(cmdCtx *cmdcontext.CmdCtx, args []string) error {
...
+		if playUsername != "" || playPassword != "" {
+			return errors.New("username and password are specified with" +
+				" flags and a URI")
+		}
+		_, user, pass := libconnect.ParseCredentialsURI(args[0])
```
oleg-jukovec, line 106: Why do we not update a passed URI?

patapenka-alexey: Hm, [cause netbox module accepts URI with credentials](https://www.tarantool.io/en/doc/latest/platform/configuration/configuration_code/#listen-uri).

oleg-jukovec: So why do we set username/password here? It is a bit confusing. Let cut credentials from the URI.

patapenka-alexey: Ok.

oleg-jukovec (approved):

test/integration/play/test_play.py
```diff
@@ -182,3 +188,115 @@ def test_play_creds(tt_cmd, tmp_path, opts, test_instance):
...
+    test_xlog_name = "test.snap"
+    # Copy the test application to the "run" directory.
+    test_app_path = os.path.join(os.path.dirname(__file__), app_name)
+    tmp_app_path = os.path.join(tmpdir, app_name)
+    shutil.copytree(test_app_path, tmp_app_path)
+    # The test file.
+    empty_file_path = os.path.join(os.path.dirname(__file__), "test_file", empty_file)
+    test_xlog_path = os.path.join(os.path.dirname(__file__), "test_file", test_xlog_name)
+    # Copy test data into temporary directory.
+    copy_data(tmpdir, [empty_file_path, test_xlog_path])
```
dmyger, line 272: Do we really need a special function for this?
Just, I doubt it will be in demand at the global `utils` level.
```suggestion
    for path in (empty_file_path, test_xlog_path):
        shutil.copy(path, dst)    
```

test/utils.py
```diff
@@ -615,3 +617,86 @@ def wait_pid_disappear(file, target_pid):
...
                 break
     assert not found
+
+
+def copy_data(dst, file_paths):
+    for path in file_paths:
+        shutil.copy(path, dst)
+
+
+def get_version(tt_cmd, tmpdir):
```
dmyger, line 627: Please move all `tt_cmd` related functions to the `tt_helper.py` file. They are not general utilities on in particular.

test/utils.py
```diff
@@ -615,3 +617,86 @@ def wait_pid_disappear(file, target_pid):
...
+    return major >= 3
+
+
+def is_tuple_format_supported(tt_cmd, tmpdir):
+    ok, major, minor, patch = get_version(tt_cmd, tmpdir)
+    assert ok
+    return major > 3 or (major == 3 and minor >= 2)
+
+
+def is_tarantool_major_one():
```
dmyger, line 667: We have `get_tarantool_version()` why we need one more method with similar functionality?
I know you just moved methods here, but since we found the problem, it's better to correct it in the right direction right away.
Please see how it is used in the rest of the tests.

test/utils.py
dmyger, line 627: 1. Why do you need `tmpdir`? `tt_cmd` is an absolute path, and `version` request does not depend on where it will be running.
2. It seems that we have two functions - one for Tarantool `get_tarantool_version` and this new one. And it would be wonderful to have the same response format, at least. What are the benefits of returning `boolean` status and `patch` number?

patapenka-alexey: Done.

patapenka-alexey: Agree. `get_version()` has been removed, all other `tt` helpers moved to the `utils.py` with applied `get_tarantool_version()`.

test/utils.py
patapenka-alexey: Updated.

test/utils.py
```diff
@@ -565,6 +567,59 @@ def is_tarantool_ee():
...
+    cmd = ["tarantool", "--version"]
+    instance_process = subprocess.run(
+        cmd,
+        stderr=subprocess.STDOUT,
+        stdout=subprocess.PIPE,
+        text=True
+    )
+    if instance_process.returncode == 0:
+        return "Tarantool 1." in instance_process.stdout
+    return False
```
dmyger, line 600: Why just don't use `get_tarantool_version`?
```suggestion
   major, minor = get_tarantool_version()
    return major == 1
```

patapenka-alexey: Yeap, we can. Updated.

dmyger (approved):

PR #1059, opened 2024-12-09
coredump: support options in `pack` subcommand

### LINKED ISSUE #763: coredump: allow to choose tarantool executable to pack
My scenario:

```sh
$ ./tools/tarabrt.sh -e src/tarantool -c /tmp/Dbgm3j/core
```

I would like to do the same using `tt coredump`, but I see no analogue for the `-e` (executable) option.

```sh
$ tt coredump pack -h
pack tarantool coredump into tar.gz archive

USAGE
  tt coredump pack <COREDUMP> [flags]
 

FLAGS
  -h, --help   help for pack
```

I would also highlight that [tarabrt.sh](https://github.com/tarantool/tarantool/blob/d7b72654799272e541882bea3e16ea79b1b7a36d/tools/tarabrt.sh) has other options that are not present in `tt coredump pack`.

```sh
$ ./tools/tarabrt.sh -h
./tools/tarabrt.sh - Tarantool Automatic Bug Reporting Tool

This tool collects all required artefacts (listed below) and packs them into
a single archive with unified format:
  - /checklist      - the plain text file with the list of tarball contents
  - /version        - the plain text file with `tarantool --version' output
  - /tarantool      - the executable binary file produced the core dump
  - /coredump       - the core dump file produced by the executable
  - /etc/os-release - the plain text file with OS identification data
  - all shared libraries loaded (even via dlopen(3)) at the crash moment

SYNOPSIS

  ./tools/tarabrt.sh [-h] [-c core] [-d dir] [-e executable] [-p procID] [-t datetime]

Supported options are:
  -c COREDUMP                   Use file COREDUMP as a core dump to examine.
                                See core(5) for more info.

  -d DIRECTORY                  Create the resulting archive with the artefacts
                                within DIRECTORY.

  -e TARANTOOL                  Use file TARANTOOL as the executable file for
                                examining with a core dump COREDUMP. If PID is
                                specified, the one from /proc/PID/exe is chosen
                                (see proc(5) for more info). If TARANTOOL is
                                omitted, /usr/bin/tarantool is chosen.

  -p PID                        PID of the dumped process, as seen in the PID
                                namespace in which the given process resides
                                (see %p in core(5) for more info). This flag
                                have to be set when tarabrt.sh is used as
                                kernel.core_pattern pipeline script.

  -t DATETIME                   Time of dump, expressed as seconds since the
                                epoch, 1970-01-01 00:00:00 +0000 (UTC).

  -h                            Shows this message and exit.

USAGE

  - Manual usage. User can simply pack all necessary artefacts by running the
    following command.
    $ ./tools/tarabrt.sh -c ./core -d /tmp

  - Automatic usage. If user faces the failures often, one can set this script
    as a pipe receiver in kernel.core_pattern syntax.
    # sysctl -w kernel.core_pattern="|./tools/tarabrt.sh -d /var/core -p %p -t %t"
```

elhimov:
`-e,--executable path`  Tarantool executable path
`-p,--pid pid`   PID of the dumped process
`-t,--time time`  Time of dump (seconds since the Epoch)

Closes #763

### DIFF
--- a/.github/workflows/full-ci.yml
+++ b/.github/workflows/full-ci.yml
@@ -51,6 +51,8 @@ jobs:
         run: sudo systemctl kill mono-xsp4 || true
 
       - name: Integration tests
+        env:
+          TT_ENABLE_COREDUMP_TESTS: '1'
         run: mage integrationfull
 
   full-ci-ce-linux-arm64:
@@ -83,6 +85,8 @@ jobs:
         run: mage unit:fullSkipDocker
 
       - name: Integration tests
+        env:
+          TT_ENABLE_COREDUMP_TESTS: '1'
         run: mage integrationfullskipdocker
 
   full-ci-sdk:
@@ -139,6 +143,8 @@ jobs:
         run: sudo systemctl kill mono-xsp4 || true
 
       - name: Integration tests
+        env:
+          TT_ENABLE_COREDUMP_TESTS: '1'
         run: mage integrationfull
 
   full-ci-macOS:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,15 +14,20 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   an extra path with modules.
 - `tt play`: support connection to a target instance by `application` name
   or `application:instance` name.
+- `tt coredump pack`: add options to customize coredump packing:
+  * `-e (--executable)`: specify Tarantool executable path.
+  * `-p (--pid)`: specify PID of the dumped process.
+  * `-t (--time)`: specify time of dump (seconds since the Epoch).
 
 ### Changed
 
 - `tt stop/kill/clean/logrotate`: no longer need:
    * Instances scripts for multi-instance applications.
    * Cluster config for tarantool3-based cluster applications.
-
 - `tt logrotate`: don't exit at non-running instance, just warn and proceed
   with the other instances, like `tt stop` and `tt kill` do.
+- `tt coredump pack`: by default tarantool executable path is obtained from
+  `PATH` instead of using the hardcoded path `/usr/bin/tarantool`.
 
 ### Fixed
 
--- a/cli/cmd/coredump.go
+++ b/cli/cmd/coredump.go
@@ -6,9 +6,17 @@ import (
 	"github.com/tarantool/tt/cli/util"
 )
 
+var (
+	coredumpPackExecutable      string
+	coredumpPackPID             uint
+	coredumpPackTime            string
+	coredumpPackOutputDirectory string
+	coredumpInspectSourceDir    string
+)
+
 // NewCoredumpCmd creates coredump command.
 func NewCoredumpCmd() *cobra.Command {
-	var coredumpCmd = &cobra.Command{
+	var cmd = &cobra.Command{
 		Use:   "coredump",
 		Short: "Perform manipulations with the tarantool coredumps",
 	}
@@ -17,12 +25,28 @@ func NewCoredumpCmd() *cobra.Command {
 		Use:   "pack COREDUMP",
 		Short: "pack tarantool coredump into tar.gz archive",
 		Run: func(cmd *cobra.Command, args []string) {
-			if err := coredump.Pack(args[0]); err != nil {
+			err := coredump.Pack(args[0],
+				coredumpPackExecutable,
+				coredumpPackOutputDirectory,
+				coredumpPackPID,
+				coredumpPackTime,
+			)
+			if err != nil {
 				util.HandleCmdErr(cmd, err)
 			}
 		},
 		Args: cobra.ExactArgs(1),
 	}
+	packCmd.Flags().StringVarP(&coredumpPackExecutable, "executable", "e", "",
+		"Tarantool executable path")
+	packCmd.Flags().StringVarP(&coredumpPackOutputDirectory, "directory", "d", "",
+		"Directory the resulting archive is created in")
+	packCmd.Flags().StringVarP(&coredumpPackTime, "time", "t", "",
+		"Time of dump, expressed as seconds since the Epoch (1970-01-01 00:00 UTC)")
+	packCmd.Flags().UintVarP(&coredumpPackPID, "pid", "p", 0,
+		"PID of the dumped process, as seen in the PID namespace in which\n"+
+			"the given process resides (see %p in core(5) for more info). This flag\n"+
+			"is to be used when tt is used as kernel.core_pattern pipeline script")
 
 	var unpackCmd = &cobra.Command{
 		Use:   "unpack ARCHIVE",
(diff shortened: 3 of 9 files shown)


cli/cmd/coredump.go
```diff
@@ -17,12 +25,28 @@ func NewCoredumpCmd() *cobra.Command {
...
+	packCmd.Flags().StringVarP(&coredumpPackExecutable, "executable", "e", "",
+		"Tarantool executable path")
+	packCmd.Flags().StringVarP(&coredumpPackOutputDirectory, "directory", "d", "",
+		"Directory the resulting archive is created in")
+	packCmd.Flags().StringVarP(&coredumpPackTime, "time", "t", "",
+		"Time of dump, expressed as seconds since the Epoch (1970-01-01 00:00 UTC)")
+	packCmd.Flags().UintVarP(&coredumpPackPID, "pid", "p", 0,
+		"PID of the dumped process, as seen in the PID namespace in which\n"+
+			"the given process resides (see %p in core(5) for more info). This flag\n"+
+			"is to be used when tt is used as kernel.core_pattern pipeline script")
```
patapenka-alexey, line 49:
```suggestion
		"the given process resides (see %p in core(5) for more info). This flag\n"+
		"is to be used when tt is used as kernel.core_pattern pipeline script")
```

.github/workflows/full-ci.yml
```diff
@@ -51,7 +51,11 @@ jobs:
         run: sudo systemctl kill mono-xsp4 || true
 
       - name: Integration tests
-        run: mage integrationfull
+        env:
+          TT_ENABLE_COREDUMP_TESTS: '1'
+        run: |
+          sudo apt install -y gdb
```
oleg-jukovec, line 57: We have two actions:

https://github.com/tarantool/tt/tree/master/.github/actions/prepare-ce-test-env
https://github.com/tarantool/tt/tree/master/.github/actions/prepare-ee-test-env

For installing some stuff.

CHANGELOG.md
```diff
@@ -14,9 +14,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
 - `tt play`: support connection to a target instance by `application` name
   or `application:instance` name.
+- `tt coredump pack`: add several options to customize dump processing:
+  * `-e (--executable)`: specify Tarantool executable path.
+  * `-p (--pid)`: specify PID of the dumped process.
+  * `-t (--time)`: specify time of dump (seconds since the Epoch).
 
 ### Changed
 
+- `tt coredump pack`: by default tarantool executable path is obtained from
```
oleg-jukovec, line 24: Should it use a tarantool binary from the tt environment or not?

cli/cmd/coredump.go
oleg-jukovec: The command:
`goimports -w cli/cmd/coredump.go`

may help to format the code and resolve all problems in formatting.

patapenka-alexey (commented):
Thank you for the patch!

.github/workflows/full-ci.yml
elhimov: Done.

cli/cmd/coredump.go
elhimov: @patapenka-alexey , this is how go decorates multiline string pattern (I suppose the indentation is used to point out that this is continuation of the string.
@oleg-jukovec , it seems vscode extension does this formatting for me. Thanks for info, anyway.

test/utils.py
```diff
@@ -518,11 +518,12 @@ def is_valid_tarantool_installed(
     return True
 
 
-def get_tarantool_version():
+def get_tarantool_version(env=None):
```
dmyger, line 521: Let me suggest copying the logic from `tt-ee` to look up `tarantool` executable via the environment variable `TARANTOOL_BIN`.
This would simplify parameter passing, it would be enough to set the right variable, without additional manipulations with `PATH` and function arguments extension.
And, it will also bring the same logic to how the `tarantool` search works.

test/integration/coredump/test_coredump.py
```diff
@@ -54,7 +65,7 @@ def apport_crash_to_coredump(crash):
         resource.setrlimit(resource.RLIMIT_CORE, (resource.RLIM_INFINITY, rlim_core_hard))
     # Crash tarantool.
     cmd = ["tarantool", "-e", "require('ffi').cast('char *', 0)[0] = 42"]
-    rc, output = run_command_and_get_output(cmd, cwd=coredump_tmpdir)
+    rc, output = run_command_and_get_output(cmd, cwd=coredump_dir, env=env)
```
dmyger, line 68: In this case, you will lose other environment settings that may be needed in some situations.
The right way would be to make a copy of the current environment and add new settings to it.
But, I suggest you make a copy of `tarantool` search logic from the `tt-ee` project, then you don't need to make such changes here, and nor add a new `fixture` `coredump_alt`.

test/integration/coredump/test_coredump.py
```diff
@@ -19,31 +19,42 @@
...
+        assert rc == 0
+        return outdir / 'CoreDump'
 
-    def apport_crash_to_coredump(crash):
-        apport_unpack_dir = coredump_tmpdir / 'apport-unpack'
-        rc, output = run_command_and_get_output(['apport-unpack', crash, apport_unpack_dir])
-        return apport_unpack_dir / 'CoreDump'
+    def coredump_systemd(core_source, outdir):
+        core = outdir / core_source.stem
+        rc, _ = run_command_and_get_output(['coredumpctl', 'dump', f'--output={core}'])
```
dmyger, line 42: If you don't need `output` why you keep the same function here?
```suggestion
        rc = subprocess.run(['coredumpctl', 'dump', f'--output={core}']).returncode
```
This way the output of the command will not be captured and deleted, which will then appear in the log in case of problems with the test, and this in itself can help in finding problems.

patapenka-alexey (approved):
Thank you for the patch!

CHANGELOG.md
elhimov: It definitely should, but currently it's not. Let me resolve this in the scope of another issue (#1069).

oleg-jukovec:
Please, add a TarantoolBot message into the commit.

test/integration/coredump/test_coredump.py
elhimov: Adjusted to use executable path instead of changing environment.
Considering that we need 2 different coredumps in the scope of single test session I see no way to avoid using another fixture.

test/utils.py
elhimov: Adjusted to use executable path instead of changing environment. `tt-ee` logic as-is is not suitable here.

elhimov:
> Please, add a TarantoolBot message into the commit.

Done

dmyger (approved):

oleg-jukovec (approved):

PR #1063, opened 2024-12-10
coredump: fix inspect failure for tarantool-ee coredumps

### LINKED ISSUE #1062: `tt coredump inspect` fails for tarantool-ee if source directory option (`-s`) is omitted
Steps to reproduce:

1. Make sure coredump limit is set properly.
   ```bash
   $ ulimit -c unlimited
   ```
1. Setup any tarantool-ee
   ```bash
   $ tar xvf tarantool-enterprise-sdk-gc64-3.2.0-0-r40.linux.x86_64.tar.gz
   $ source tarantool-enterprise/env.sh
   ```
1. Crash tarantool
   ```bash
   $ tarantool -e "require('ffi').cast('char *', 0)[0] = 42"
   ```
1. Pack coredump
   ```bash
   $ tt coredump pack core
   ```
1. Inspect coredump
   ```bash
   $ tt coredump inspect tarantool-core-N-*.tar.gz
   ⨯ inspect script execution failed: exit status 1
   ```

elhimov:
Closes #1062

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- `tt coredump inspect`: fails for tarantool-ee coredump archive if the source
+  directory is missing.
+
 ## [2.6.0] - 2024-11-29
 
 The release introduces `upgrade` and `downgrade` subcommands for
--- a/cli/coredump/scripts/gdb.sh
+++ b/cli/coredump/scripts/gdb.sh
@@ -79,7 +79,7 @@ SOURCES=${SOURCES:-${WORKDIR}/sources}
 # the wrapper more complex by additional checks, so this activity
 # is left for the user.
 if [ ! -d "${SOURCES}" ]; then
-	REGEX='Tarantool \d+\.\d+\.\d+(-(alpha|beta|rc)[0-9]+|(-entrypoint))?-\d+-g\K[a-f0-9]+'
+	REGEX='Tarantool( Enterprise)? \d+\.\d+\.\d+(-(alpha|beta|rc)[0-9]+|(-entrypoint))?-\d+-g\K[a-f0-9]+'
 	REVISION=$(grep -oP "$REGEX" "$VERSION")
 	cat <<SOURCES
 ================================================================================


CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
 ### Fixed
+- `tt coredump inspect`: fails for tarantool-ee coredump archive if the source
```
patapenka-alexey, line 15: Please, add an empty line after `Fixed`.

elhimov: Done.

patapenka-alexey (approved):
Thank you for the patch. LGTM!

test/integration/coredump/test_coredump.py
```diff
@@ -38,7 +38,7 @@ def apport_crash_to_coredump(crash):
     if not core_pattern.startswith('|'):
         core_wildcard = core_pattern.strip().split('%')[0] + '*'
         if not os.path.isabs(core_wildcard):
-            core_wildcard = coredump_tmpdir / core_wildcard
+            core_wildcard = os.path.join(coredump_tmpdir.as_posix(), core_wildcard)
```
dmyger, line 41: What problem is fixed here?

elhimov: Previous version just didn't work, because core_wildcard is expected to be string later
```python
glob.glob(core_wildcard)
```
but it was not.

Thought it was me who wrote the previous version, I have no idea why it turned out like this.
The only version -- it was the last-minute-change. Tests where real coredump are generated are not the part of CI, neither general not full-ci (one can run them explicitly by specifying the proper env), so it seems I just haven't check them after that last-minute-change.

dmyger: As for me, it's a more easy and clear way to write `glob.glob(str(core_wildcard))`. Than such an odd expression when it's not quite clear at a glance exactly what he's doing.
The question was mostly why need `.as_posix()`?

elhimov: I think I got your point. However considering that:
- This code is changed in the next PR (#1059)
- This change doesn't relate directly to the issue

in order to avoid wasting time I dropped this change completely from this PR.

dmyger (approved):

Issue #1050, opened 2024-12-03
aeon: connect implementation


oleg-jukovec:
The implementation of the command should contain:

* A dummy server implementation that could be used in tests (maybe written on python?);
* A `tt aeon connect URI` command implementation and test for it;

The command should be able to connect to aeon and execute command on it via gRPC.

We already have an implementation of console in the `tt connect` so the deal should be only in the gRPC connection only.

PR #1065, opened 2024-12-12
aeon: connect implementation

dmyger:
Mock server Implement some base methods for integration tests.

Part of #1050


cli/aeon/client.go
```diff
@@ -0,0 +1,232 @@
...
+// ResultType is a custom type to format output with console.Formatter interface.
+type ResultType struct {
+	data  map[string][]any
+	count int
+}
+
+// ResultError wraps pb.Error to implement console.Formatter interface.
+type ResultError struct {
+	*pb.Error
+}
```
oleg-jukovec, line 34: It is better to move the types and methods into a separate files (and test for the public API too).

cli/aeon/decode.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/tarantool/tt/cli/aeon/pb"
+)
+
+/*
+decodeValue convert a value obtained from protobuf into a value that can be used as an
+argument to Tarantool functions.
+
+Copy from https://github.com/tarantool/aeon/blob/master/aeon/grpc/server/pb/decode.go
+*/
+func decodeValue(val *pb.Value) (any, error) {
```
oleg-jukovec, line 19: Just to keep the same code style.
```suggestion
// decodeValue convert a value obtained from protobuf into a value that can be used as an
// argument to Tarantool functions.
// 
// Copy from https://github.com/tarantool/aeon/blob/master/aeon/grpc/server/pb/decode.go
func decodeValue(val *pb.Value) (any, error) {
```

cli/cmd/aeon.go
```diff
@@ -5,24 +5,26 @@ import (
 	"fmt"
 
 	"github.com/spf13/cobra"
-	aeon "github.com/tarantool/tt/cli/aeon/cmd"
+	aeon "github.com/tarantool/tt/cli/aeon"
+	aeon_cmd "github.com/tarantool/tt/cli/aeon/cmd"
```
oleg-jukovec, line 9:
```suggestion
	aeoncmd "github.com/tarantool/tt/cli/aeon/cmd"
```

cli/console/console.go
```diff
@@ -0,0 +1,265 @@
...
+
+	"github.com/apex/log"
+	"golang.org/x/term"
+
+	"github.com/tarantool/go-prompt"
+)
+
+const (
+	maxLivePrefixIndent = 15
+	// see https://github.com/tarantool/tarantool/blob/b53cb2aeceedc39f356ceca30bd0087ee8de7c16/src/box/lua/console.c#L265
```
oleg-jukovec, line 20:
```suggestion
	// See https://github.com/tarantool/tarantool/blob/b53cb2aeceedc39f356ceca30bd0087ee8de7c16/src/box/lua/console.c#L265
```

cli/console/console.go
```diff
@@ -0,0 +1,265 @@
...
+type ConsoleOpts struct {
+	// Handler is the implementation of command processor.
+	Handler Handler
+
+	// History if specified than save input commands with it.
+	History HistoryKeeper
+
+	// Format options set how to formatting result.
+	Format Format
+}
```
oleg-jukovec, line 39:
```suggestion
type ConsoleOpts struct {
	// Handler is the implementation of command processor.
	Handler Handler
	// History if specified than save input commands with it.
	History HistoryKeeper
	// Format options set how to formatting result.
	Format Format
}
```

cli/console/format.go
oleg-jukovec, line 1: Please, add tests.

cli/console/handler.go
```diff
@@ -0,0 +1,23 @@
...
+	// Complete checks the input and return available variants to continue typing.
+	Complete(input prompt.Document) []prompt.Suggest
+
+	// Execute accept input to perform actions defined by client implementation.
+	// Expecting that result type implements Formatter interface.
+	Execute(input string) any
+
+	// Close notify handler to terminate execution and close any opened streams.
+	Close()
+}
```
oleg-jukovec, line 23:
```suggestion
// Handler is a auxiliary abstraction to isolate the console from
// the implementation of a particular instruction processor.
type Handler interface {
	// Title return name of instruction processor instance.
	Title() string
	// Validate the input string.
	Validate(input string) bool
	// Complete checks the input and return available variants to continue typing.
	Complete(input prompt.Document) []prompt.Suggest
	// Execute accept input to perform actions defined by client implementation.
	// Expecting that result type implements Formatter interface.
	Execute(input string) any
	// Close notify handler to terminate execution and close any opened streams.
	Close()
}
```

cli/console/handler.go
```diff
@@ -0,0 +1,23 @@
...
+	Title() string
+
+	// Validate the input string.
+	Validate(input string) bool
+
+	// Complete checks the input and return available variants to continue typing.
+	Complete(input prompt.Document) []prompt.Suggest
+
+	// Execute accept input to perform actions defined by client implementation.
+	// Expecting that result type implements Formatter interface.
```
oleg-jukovec, line 18: So why does it return `any`?

cli/console/history.go
```diff
@@ -0,0 +1,136 @@
...
+
+	"github.com/tarantool/tt/cli/util"
+)
+
+const (
+	DefaultHistoryFileName = ".tarantool_history"
+	DefaultHistoryLines    = 10000
+)
+
+type History struct {
```
oleg-jukovec, line 21: Missed comments.

cli/console/history.go
oleg-jukovec, line 1: And missed tests.

oleg-jukovec:
Please, add a CHANGELOG.md entry and a TarantoolBot comment too.

cli/console/format.go
dmyger: We don't care about return type here, it's only important that the Console should be able to display it on the screen.
And this can be any object providing one of the following interfaces:
- Formatter
- Stringer
- error

But even otherwise, an error will be printed that formatting of this type is not supported, which will be seen on test time.

cli/console/handler.go
dmyger: We don't care about return type here, it's only important that the Console should be able to display it on the screen.
And this can be any object providing one of the following interfaces:
- Formatter
- Stringer
- error

But even otherwise, an error will be printed that formatting of this type is not supported, which will be seen on test time.

oleg-jukovec: Please, add a comment about it into the interface method.

cli/console/format.go
```diff
@@ -0,0 +1,47 @@
...
+
+	} else if so, ok := data.(fmt.Stringer); ok {
+		// Then checking is it has String method.
+		fmt.Println(so.String())
+
+	} else if eo, ok := data.(error); ok {
+		// Then checking is it has Error method.
+		fmt.Println("Error:\n", eo.Error())
+
+	} else {
```
oleg-jukovec, line 43:
```suggestion
	} else if so, ok := data.(fmt.Stringer); ok {
		// Then checking is it has String method.
		fmt.Println(so.String())
	} else if eo, ok := data.(error); ok {
		// Then checking is it has Error method.
		fmt.Println("Error:\n", eo.Error())
	} else {
```

cli/aeon/generate-pb.sh
```diff
@@ -0,0 +1,23 @@
...
+EOF
+fi
+
+if [ ! -x "$(command -v protoc-gen-go)" ]; then
+	go install google.golang.org/protobuf/cmd/protoc-gen-go@latest
+fi
+
+if [ ! -x "$(command -v protoc-gen-go-grpc)" ]; then
+	go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest
+fi
```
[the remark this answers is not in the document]
oleg-jukovec: Please, add a new README.md file into the `aeon` and a link to it from the main `README.md` file.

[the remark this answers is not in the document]
dmyger: Done.

cli/console/handler.go
dmyger: Done.

cli/aeon/client.go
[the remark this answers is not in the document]
dmyger: Client tested with integration test.
Add test on format output.

oleg-jukovec:
Please, rebase to the master branch to test the CI.

cli/aeon/results.go
```diff
@@ -0,0 +1,53 @@
...
+	if len(r.names) == 0 {
+		return ""
+	}
+	output, err := formatter.MakeOutput(f.Mode, r.asYaml(), f.Opts)
+	if err != nil {
+		return fmt.Sprintf("can't format output: %s;\nResults:\n%v", err, r)
+	}
+	return output
+}
+
```
oleg-jukovec, line 50: A comment missed.

cli/aeon/results_export_test.go
```diff
@@ -0,0 +1,16 @@
+package aeon
+
+import "github.com/tarantool/tt/cli/aeon/pb"
+
+func NewResultType(names []string, rows []ResultRow) ResultType {
+	return ResultType{
+		names: names,
+		rows:  rows,
+	}
+}
```
oleg-jukovec, line 10: Maybe we could just to add the constructor into the public API?

cli/console/history.go
```diff
@@ -0,0 +1,137 @@
...
+	"regexp"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/tarantool/tt/cli/util"
+)
+
+const (
+	DefaultHistoryFileName = ".tarantool_history"
```
oleg-jukovec, line 17: Do a separate history files are used for Tarantool and aeon?

cli/console/format.go
oleg-jukovec: Looks still relevant.

cli/aeon/client.go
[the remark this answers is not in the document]
oleg-jukovec: Where exactly?

cli/aeon/results_export_test.go
oleg-jukovec: This all sounds like a lot of type coupling. Why do we need to return a private type (with a public name) from a public interface? How to implement the interface outside the module?

cli/console/history.go
oleg-jukovec: I think we need .aeon_history here since the commands may be different in the future.

dmyger: Done.

cli/console/format.go
dmyger: Done.

cli/aeon/client.go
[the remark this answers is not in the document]
dmyger: Done, added `format_test.go`.

cli/aeon/results_export_test.go
dmyger: > Why do we need to return a private type (with a public name) from a public interface?

In this solution, the object instance itself is passed outside. But with the ensuring that its contents are protected.
I don't see a reason why it might be necessary to change the internal state of such an object from the outside.
It is enough to call its implemented methods.
As for example, the same concept was used in the `Console` struct in `tt/cli/connect/console.go` file.

> How to implement the interface outside the module?

For what purposes might it be necessary to implement an interface outside the module?
Outside the module, its representation of its internal data, does not make any explicit sense.

dmyger: Make types as private.

oleg-jukovec (approved):

patapenka-alexey (approved):

PR #1076, opened 2024-12-26
replicaset downgrade: make version a positional argument

### LINKED ISSUE #1073: replicaset downgrade: specify version as positional argument instead of mandatory option
Currently version is to be specified with mandatory(!) option `--version`. It looks natural and more convenient to use positional argument in such a case. Usage will be:
```bash
tt replicaset downgrade  (<APP_NAME> | <URI>) VERSION
```

elhimov:
Closes #1073

@TarantoolBot
title: replicaset downgrade: make version a positional argument

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `tt coredump pack`: if `-e` option is omitted first search tarantool
   executable in tt environment then in `PATH` instead of using the hardcoded
   path `/usr/bin/tarantool`.
+- `tt replicaset downgrade`: drop option `-v` (`--version`). Pass version as a
+  positional argument rather than option.
 
 ### Fixed
 
--- a/cli/cmd/connect.go
+++ b/cli/cmd/connect.go
@@ -129,18 +129,17 @@ func makeConnOpts(network, address string, connCtx connect.ConnectCtx) connector
 	}
 }
 
-// resolveConnectOpts tries to resolve the first passed argument as an instance
-// name to replace it with a control socket or as a URI with/without
-// credentials.
+// resolveConnectOpts tries to resolve target argument to replace it
+// with a control socket or as a URI with/without credentials.
 // It returns connection options and the remaining args.
 func resolveConnectOpts(cmdCtx *cmdcontext.CmdCtx, cliOpts *config.CliOpts,
-	connectCtx *connect.ConnectCtx, args []string) (
-	connOpts connector.ConnectOpts, newArgs []string, err error) {
+	connectCtx *connect.ConnectCtx, target string) (
+	connOpts connector.ConnectOpts, err error) {
 
-	newArgs = args[1:]
 	// FillCtx returns error if no instances found.
 	var runningCtx running.RunningCtx
-	if fillErr := running.FillCtx(cliOpts, cmdCtx, &runningCtx, args, false); fillErr == nil {
+	fillErr := running.FillCtx(cliOpts, cmdCtx, &runningCtx, []string{target}, false)
+	if fillErr == nil {
 		if len(runningCtx.Instances) > 1 {
 			err = fmt.Errorf("specify instance name")
 			return
@@ -159,19 +158,19 @@ func resolveConnectOpts(cmdCtx *cmdcontext.CmdCtx, cliOpts *config.CliOpts,
 				connector.UnixNetwork, runningCtx.Instances[0].ConsoleSocket, *connectCtx,
 			)
 		}
-	} else if libconnect.IsCredentialsURI(args[0]) {
+	} else if libconnect.IsCredentialsURI(target) {
 		if connectCtx.Username != "" || connectCtx.Password != "" {
 			err = fmt.Errorf("username and password are specified with" +
 				" flags and a URI")
 			return
 		}
-		newURI, user, pass := libconnect.ParseCredentialsURI(args[0])
+		newURI, user, pass := libconnect.ParseCredentialsURI(target)
 		network, address := libconnect.ParseBaseURI(newURI)
 		connectCtx.Username = user
 		connectCtx.Password = pass
 		connOpts = makeConnOpts(network, address, *connectCtx)
 		connectCtx.ConnectTarget = newURI
-	} else if libconnect.IsBaseURI(args[0]) {
+	} else if libconnect.IsBaseURI(target) {
 		// Environment variables do not overwrite values.
 		if connectCtx.Username == "" {
 			connectCtx.Username = os.Getenv(libconnect.TarantoolUsernameEnv)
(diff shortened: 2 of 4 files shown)


cli/cmd/replicaset.go
```diff
@@ -90,33 +88,34 @@ func newUpgradeCmd() *cobra.Command {
 
 // newDowngradeCmd creates a "replicaset downgrade" command.
 func newDowngradeCmd() *cobra.Command {
+	validateVersion := func(i int) cobra.PositionalArgs {
+		return func(cmd *cobra.Command, args []string) error {
+			var versionPattern = regexp.MustCompile(`^\d+\.\d+\.\d+$`)
+			if args[i] == "" {
+				return errors.New("need to specify the version to downgrade " +
+					"use --version (-v) option")
```
patapenka-alexey, line 96:
```suggestion
```

cli/cmd/replicaset.go
```diff
@@ -90,33 +88,34 @@ func newUpgradeCmd() *cobra.Command {
...
+			} else if !versionPattern.MatchString(args[i]) {
+				return errors.New("--version (-v) must be in the format " +
```
patapenka-alexey, line 98:
```suggestion
				return errors.New("Version must be in the format " +
```

CHANGELOG.md
```diff
@@ -29,6 +29,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `tt coredump pack`: if `-e` option is omitted first search tarantool
   executable in tt environment then in `PATH` instead of using the hardcoded
   path `/usr/bin/tarantool`.
+- `tt replicaset downgrade`: make version a positional argument rather than
+  using the mandatory option `-v` (`--version`)
```
patapenka-alexey, line 33:
```suggestion
  using the mandatory option `-v` (`--version`).
```

patapenka-alexey (commented):
Thanks for the patch. Please, fix a few things.

CHANGELOG.md
elhimov: Fixed.

cli/cmd/replicaset.go
elhimov: Thanks, I missed it. Fixed.

cli/cmd/replicaset.go
```diff
@@ -634,14 +630,14 @@ func internalReplicasetDowngradeModule(cmdCtx *cmdcontext.CmdCtx, args []string)
 	}, replicasetcmd.DowngradeOpts{
 		ChosenReplicasetAliases: chosenReplicasetAliases,
 		Timeout:                 lsnTimeout,
-		DowngradeVersion:        downgradeVersion,
+		DowngradeVersion:        args[1],
```
dmyger, line 633: You were trying to avoid overuse of `args`, but why not use here the named variable `downgradeVersion` and set it at the `PreRun` step?

cli/cmd/connect.go
```diff
@@ -129,18 +129,17 @@ func makeConnOpts(network, address string, connCtx connect.ConnectCtx) connector
...
-// name to replace it with a control socket or as a URI with/without
-// credentials.
+// resolveConnectOpts tries to resolve target argument to replace it
+// with a control socket or as a URI with/without credentials.
 // It returns connection options and the remaining args.
 func resolveConnectOpts(cmdCtx *cmdcontext.CmdCtx, cliOpts *config.CliOpts,
-	connectCtx *connect.ConnectCtx, args []string) (
-	connOpts connector.ConnectOpts, newArgs []string, err error) {
+	connectCtx *connect.ConnectCtx, target string) (
+	connOpts connector.ConnectOpts, err error) {
```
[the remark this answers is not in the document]
elhimov: Well... I don't like them too (even with small functions), but I'm trying to simplify review by making as little change as possible and that's why I kept this part intact.

cli/cmd/replicaset.go
elhimov: It seems there is some misunderstanding.
My main concern is to avoid using args as an _argument in a signature_ of a function that is not command handler, because when I see a function that expects `args` it is unclear the exact sequence of `args` that are expected. And even if it's documented, it would look like unnecessary constraint. Keeping `args` argument only in command handlers forces us to put the knowledge about every arg in `args` only in corresponding handler and don't spread out it to helper functions.
Considering the above your suggestion is the next step to improve readability. We may 'resolve' `args` to named variables in the very beginning of the handler and then refer them instead of `args`. I would prefer to declare it locally rather than using global variables (which would needed if declaring it in `PreRun`). I know that all (or at least the most of) optional arguments (flags in Cobra's terms) are represented with global variables, but it is forced by Cobra's flags parsing machinery which is not applied to positional arguments.

elhimov: I've made change to illustrate what I mean

CHANGELOG.md
```diff
@@ -29,6 +29,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `tt coredump pack`: if `-e` option is omitted first search tarantool
   executable in tt environment then in `PATH` instead of using the hardcoded
   path `/usr/bin/tarantool`.
+- `tt replicaset downgrade`: make version a positional argument rather than
+  using the mandatory option `-v` (`--version`).
```
oleg-jukovec, line 33: So maybe remove the `-v` flag?

elhimov: Do you mean this CHANGELOG entry shouldn't mention it?

oleg-jukovec: Uh, I see, you already remove it. Yes, a changelog entry should mention it so - it is not obvious at now.

oleg-jukovec: And a TarantoolBot comment too.

patapenka-alexey (approved):

oleg-jukovec:
Please, fix a linter errors.

cli/cmd/replicaset.go
```diff
@@ -479,7 +475,7 @@ type replicasetCtx struct {
 }
 
 // replicasetFillCtx fills the replicaset command context.
-func replicasetFillCtx(cmdCtx *cmdcontext.CmdCtx, ctx *replicasetCtx, args []string,
+func replicasetFillCtx(cmdCtx *cmdcontext.CmdCtx, ctx *replicasetCtx, target string,
```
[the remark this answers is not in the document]
elhimov: It seems there is some misunderstanding.

I see no problem in _passing_ `args`, I see the problem in _expecting_ `args` _in a non-handler function_. Thus I see no problem in `replicasetFillCtx(args)` (i.e. _passing_ args to helper function) if it occurs within corresponding handler function, because handler "knows" what's in `args`, and if it "knows" that `args` is actually a sequence of targets in this particular command and `replicasetFillCtx` has signature `replicasetFillCtx(targets []string)`, i.e. it expects the sequence of targets then I see no problem just call `replicasetFillCtx(args)`.

I do see the problem in `func replicasetFillCtx(args []string)` (i.e. expecting args in helper function) because it means that `replicasetFillCtx` must "know" about `args` sequence of all the commands it is used in. That means also that if any variation in `args` sequence among managed commands take place we should also pass something to `replicasetFillCtx` to identify a command within `replicasetFillCtx`. As for me it looks ugly and unfortunately we already do that -- pass `cmdCtx` to `replicasetFillCtx` as well as we pass it to `running.FillCtx`.

In my opinion there should be only one function that "knows" about command `args` -- the corresponding command handler.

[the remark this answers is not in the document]
elhimov: > I don't see a reason why the method `cmd.resolveConnectOpts()` does not use `ConnectTarget` and requires extra `target` argument.

Actually `args`-related adjustment is a kind of auxiliary change in the scope of this PR, so I'm just trying to keep it as small and as trivial as possible. It looks like the change you suggested deserves separate issue/PR.

dmyger (approved):

Totktonada:
Please, create the documentation request manually: it seems that it is not created automatically due to the lowercased `title` word.

https://github.com/tarantool/docbot/blob/bda9b762eafa5181f378287619471c3bff4c6142/docbot/settings.py#L11

https://github.com/tarantool/docbot/blob/bda9b762eafa5181f378287619471c3bff4c6142/docbot/processors.py#L23-L24

Totktonada:
Filed https://github.com/tarantool/doc/issues/4731.

PR #1084, opened 2025-01-16
aeon: fix repo URL for Aeon submodule

dmyger:
Use HTTPS instead Git protocol to avoid extra authorization.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -9,5 +9,5 @@
 	ignore = dirty
 [submodule "cli/aeon/protos"]
 	path = cli/aeon/protos
-	url = git@github.com:tarantool/aeon-api-protos.git
+	url = https://github.com/tarantool/aeon-api-protos.git
 	branch = master


.gitmodules
```diff
@@ -1,13 +1,13 @@
 [submodule "third_party/luarocks"]
 	path = cli/rocks/third_party/luarocks
-	url = https://github.com/tarantool/luarocks.git
+	url = ../luarocks.git
```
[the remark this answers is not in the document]
dmyger: 1. One of the benefits of relative path - it does not depend on server where repositories located.
If we move projects on the GitLab nothing will be required to change.
2. As I remember, the public projects will not move.

Totktonada, line 3: If we're going to use relative paths, I would use `../../tarantool/luarocks.git` instead of `../luarocks.git`. This way one can fork just one repository and the clone still works.

dmyger: Done.
But let's wait a solution, about should we use relative path or not.

Totktonada (approved):
For me it is OK, but I see that a discussion about using relative paths is ongoing. It depends on your plans where to host the repositories and I can't judge here.

oleg-jukovec (approved):
Anyway, we can fix relative paths it if something breaks in the future after migration to GitLab.

dmyger (commented):
Due to test issues, did the absolute path, just don't spend time to figure out why they don't work.

PR #1087, opened 2025-01-20
lib: move url parsing to connect package

dmyger:
Moves the common URL parsing logic from the `cmd` package to the
`connect` package.

Part of #TNTP-1081


oleg-jukovec:
Could we close it or we need for something?

PR #1096, opened 2025-02-04
pack: support .packignore

### LINKED ISSUE #812: Allow to ignore paths on artifact pack
There are some cases when I have files that may be useful, but shouldn't be a part of artifact. Now I can "filter" then through `tt.post-build`, but it is inconvenient to use use it together with `tt build`.

For example, I have some local rock that my application needs to use. I store it in the `custom_rocks` folder and install with `tt.pre-build` script. I don't need `custom_rocks/myrock-scm-1.all.rock` file to be a part of the artifact since application use one that's already unpacked to its `.rocks/share/...`. I can remove it from the artifact with `tt.post-build`, but then the rock will be removed in case I run `tt build` as well --- and I don't want it to be gone. Similar things will happen even if I don't use a path to build an artifact (for example, `doc` folder).

So it would be nice to provide some file to specify the list of files to be ignored on the packing stage. (Maybe we can reuse `.gitignore`, I haven't thought about it that much.) We also need to consider the case when the file is required on `tt.pre-build` or `tt-post-build` stage, but shouldn't be a part of the artifact, as I had described above.

elhimov:
In some cases, there are files that may be useful, but should not be part of artifact. The ability to add files and directories to the .packignore file has been added, which allows you to ignore these files and directories when packing.

Closes #812

### DIFF
--- a/cli/pack/common.go
+++ b/cli/pack/common.go
@@ -1,6 +1,7 @@
 package pack
 
 import (
+	"errors"
 	"fmt"
 	"io/fs"
 	"os"
@@ -33,6 +34,8 @@ const (
 	versionLuaFileName = "VERSION.lua"
 
 	rocksManifestPath = ".rocks/share/tarantool/rocks/manifest"
+
+	ignoreFile = ".packignore"
 )
 
 var (
@@ -50,6 +53,8 @@ var (
 		),
 	}
 )
+
+type skipFilter func(srcInfo os.FileInfo, src string) bool
 
 type RocksVersions map[string][]string
 
@@ -76,9 +81,8 @@ func skipDefaults(srcInfo os.FileInfo, src string) bool {
 }
 
 // appArtifactsFilters returns a slice of skip functions to avoid copying application artifacts.
-func appArtifactsFilters(cliOpts *config.CliOpts, srcAppPath string) []func(
-	srcInfo os.FileInfo, src string) bool {
-	filters := make([]func(srcInfo os.FileInfo, src string) bool, 0)
+func appArtifactsFilters(cliOpts *config.CliOpts, srcAppPath string) []skipFilter {
+	filters := make([]skipFilter, 0)
 	if cliOpts.App == nil {
 		return filters
 	}
@@ -102,9 +106,8 @@ func appArtifactsFilters(cliOpts *config.CliOpts, srcAppPath string) []func(
 }
 
 // ttEnvironmentFilters prepares a slice of filters for tt environment directories/files.
-func ttEnvironmentFilters(packCtx *PackCtx, cliOpts *config.CliOpts) []func(
-	srcInfo os.FileInfo, src string) bool {
-	filters := make([]func(srcInfo os.FileInfo, src string) bool, 0)
+func ttEnvironmentFilters(packCtx *PackCtx, cliOpts *config.CliOpts) []skipFilter {
+	filters := make([]skipFilter, 0)
 	if cliOpts == nil {
 		return filters
 	}
@@ -139,10 +142,9 @@ func ttEnvironmentFilters(packCtx *PackCtx, cliOpts *config.CliOpts) []func(
 }
 
 // previousPackageFilters returns filters for the previously built packages.
-func previousPackageFilters(packCtx *PackCtx) []func(
-	srcInfo os.FileInfo, src string) bool {
+func previousPackageFilters(packCtx *PackCtx) []skipFilter {
 	pkgName := packCtx.Name
-	return []func(srcInfo os.FileInfo, src string) bool{
+	return []skipFilter{
 		func(srcInfo os.FileInfo, src string) bool {
 			name := srcInfo.Name()
 			if strings.HasPrefix(name, pkgName) {
@@ -159,13 +161,18 @@ func previousPackageFilters(packCtx *PackCtx) []func(
 
 // appSrcCopySkip returns a filter func to filter out artifacts paths.
 func appSrcCopySkip(packCtx *PackCtx, cliOpts *config.CliOpts,
-	srcAppPath string) func(srcinfo os.FileInfo, src, dest string) (bool, error) {
+	srcAppPath string) (func(srcinfo os.FileInfo, src, dest string) (bool, error), error) {
 	appCopyFilters := appArtifactsFilters(cliOpts, srcAppPath)
 	appCopyFilters = append(appCopyFilters, ttEnvironmentFilters(packCtx, cliOpts)...)
 	appCopyFilters = append(appCopyFilters, previousPackageFilters(packCtx)...)
 	appCopyFilters = append(appCopyFilters, func(srcInfo os.FileInfo, src string) bool {
 		return skipDefaults(srcInfo, src)
 	})
+	if f, err := ignoreFilter(util.GetOsFS(), filepath.Join(srcAppPath, ignoreFile)); err == nil {
+		appCopyFilters = append(appCopyFilters, f)
+	} else if !errors.Is(err, fs.ErrNotExist) {
+		return nil, fmt.Errorf("failed to load %q: %w", ignoreFile, err)
+	}
 
 	return func(srcinfo os.FileInfo, src, dest string) (bool, error) {
 		for _, shouldSkip := range appCopyFilters {
@@ -174,7 +181,7 @@ func appSrcCopySkip(packCtx *PackCtx, cliOpts *config.CliOpts,
 			}
 		}
 		return false, nil
-	}
+	}, nil
 }
 
 // getAppNamesToPack generates application names list to pack.
@@ -430,7 +437,10 @@ func copyAppSrc(packCtx *PackCtx, cliOpts *config.CliOpts, srcAppPath, dstAppPat
 		return err
 	}
 
-	skipFunc := appSrcCopySkip(packCtx, cliOpts, resolvedAppPath)
+	skipFunc, err := appSrcCopySkip(packCtx, cliOpts, resolvedAppPath)
+	if err != nil {
+		return err
+	}
 
 	// Copying application.
 	log.Debugf("Copying application source %q -> %q", resolvedAppPath, dstAppPath)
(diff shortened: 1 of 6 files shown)


cli/pack/common.go
```diff
@@ -157,6 +161,126 @@ func previousPackageFilters(packCtx *PackCtx) []func(
...
+	} else {
+		if strings.HasPrefix(pattern, "\\!") || strings.HasPrefix(pattern, "\\#") {
+			pattern = pattern[1:]
+		}
+	}
+
+	expr := pattern
+	expr = strings.ReplaceAll(expr, "/**/", "/([^/]+/)*")
+	// expr = strings.ReplaceAll(expr, "**/", "([^/]+/)*")
+	// expr = strings.ReplaceAll(expr, "/**", "[^/]*")
```
patapenka-alexey, line 183:
```suggestion
```

elhimov: Done.

cli/pack/common.go
```diff
@@ -157,6 +162,108 @@ func previousPackageFilters(packCtx *PackCtx) []func(
...
+		expr = basepath + expr
+	} else {
+		expr = "/?([^/]+/)*" + expr
+	}
+
+	return expr, negate, onlyDirs
+}
+
+// ignoreFilter returns filter that excludes files based on the patterns.
+func ignoreFilter(fsys fs.FS, ignoreFile string) skipFilter {
```
dmyger, line 196: Is it possible to split such a long function into several smaller functions.
I'd like to see self-documenting code with talking names of small functions.

cli/util/osfs.go
```diff
@@ -0,0 +1,22 @@
...
+	return osfs
+}
+
+func (fs osFS) Open(name string) (fs.File, error) {
+	return os.Open(name)
+}
+
+func (fs osFS) ReadFile(name string) ([]byte, error) {
+	return os.ReadFile(name)
+}
```
oleg-jukovec, line 22: To be honest, I didn't get the idea. Why do we need this stuff?

elhimov: There is interface fs.FS (as well as extended ones like fs.ReadFileFS), but there is no default implementation. Here it is. It should be used wherever you need to be able to substitute non-default FS. The most obvious scenario is using mock FS for testing (like this PR do). Say we have some function that expect path as an argument and we need to test it. To be able to use mock FS in such a case we need to add `fsys FS` argument to this function and replace `os.Open/os.ReadFile` within function to `fsys.Open/fsys.ReadFile`.

elhimov: Application code call the mentioned function like
```go
someFunction(utils.GetOsFS(), somePath)
```
while tests are able to substitute mockFS
```go
someFunction(someMockFS, somePath)
```

oleg-jukovec: Probably you could to use library calls instead of a custom implementation:

https://pkg.go.dev/os#DirFS
https://pkg.go.dev/os#Root.FS

cli/util/osfs.go
```diff
@@ -0,0 +1,22 @@
...
+	"os"
+)
+
+type osFS struct{}
+
+var osfs osFS
+
+func GetOsFS() fs.FS {
+	return osfs
+}
```
oleg-jukovec, line 14: The global variable is confusing here. Please, add a comments for the type.
```suggestion

func GetOsFS() fs.FS {
	return osFS{}
}
```

oleg-jukovec: I don't think there are enough allocations here for us to worry about creating the object.

The global variable is more confusing - it suggests the idea of ​​replacing the global context in tests.

elhimov: OK, fixed as suggested.

oleg-jukovec:
Please, rebase on the master branch.

cli/util/osfs.go
```diff
@@ -0,0 +1,20 @@
+package util
+
+import (
+	"io/fs"
+	"os"
+)
+
+type osFS struct{}
+
+func GetOsFS() fs.FS {
```
oleg-jukovec, line 10: Please, add a comment about purpose of the thing.

cli/util/osfs.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+type osFS struct{}
+
+// Default implementation of fs.FS interface. Interface fs.FS should be added as an argument
+// to any function where you need to be able to substitute non-default FS. The most obvious
+// scenario is using mock FS for testing. In such a case general code might use this default
+// implementation, while the tests will use some mock FS implementation (like fstest.MapFS).
+
+func GetOsFS() fs.FS {
```
oleg-jukovec, line 15: It would be in godoc style:
```suggestion
// GetOsFS returns a default implementation of fs.FS interface. Interface fs.FS should be
// added as an argument to any function where you need to be able to substitute non-default
// FS. The most obvious scenario is using mock FS for testing. In such a case general code
// might use this default implementation, while the tests will use some mock FS
// implementation (like fstest.MapFS).
func GetOsFS() fs.FS {
```
The beginning is very important here + there is should not be an empty line between comment and a function.
```
GetOsFS returns
```

oleg-jukovec (approved):

cli/util/util.go
```diff
@@ -1024,3 +1024,12 @@ func CollectWALFiles(paths []string) ([]string, error) {
 
 	return collectedFiles, nil
 }
+
+// Map is a generic slice mapper function.
+func Map[T, V any](ts []T, fn func(T) V) []V {
```
dmyger, line 1029: Is it used?

elhimov: Yes (see ignore_test.go)

cli/pack/ignore.go
```diff
@@ -0,0 +1,132 @@
...
+	cleanPattern, dirOnly, isNegate := splitIgnorePattern(pattern)
+
+	// Translate pattern to regex expression.
+	expr := cleanPattern
+	// Turn escaped '*' and '?' to their hex representation to simplify the translation.
+	expr = turnEscapedToHexCode(expr, '*')
+	expr = turnEscapedToHexCode(expr, '?')
+	// Escape symbols that designate themselves in pattern, but have special meaning in regex.
+	for _, s := range []string{"(", ")", "{", "}", "+"} {
+		// Do unescape first to avoid double escaping of the ones that are already escaped
```
patapenka-alexey, line 50:
```suggestion
		// Do unescape first to avoid double escaping of the ones that are already escaped.
```

elhimov: Done.

patapenka-alexey (approved):

elhimov:
A few corrections have been applied:
- Add comments in unit tests for better understanding of how they work
- Add new unit tests
- Fixup of the latest commit (there were duplicated entries in CHANGELOG)

cli/util/util.go
oleg-jukovec: Let's put this code into the test code then.

elhimov: Done.

cli/pack/ignore_test.go
```diff
@@ -0,0 +1,893 @@
...
+func Test_createIgnorePattern_basic(t *testing.T) {
+	checkIgnorePattern(t, ignorePatternBasicCases)
+}
+
+func Test_createIgnorePattern_negate(t *testing.T) {
+	toNegate := func(tc ignorePatternCase) ignorePatternCase {
+		result := tc
+		result.pattern = "!" + tc.pattern
+		result.isNegate = true
+		return result
```
dmyger, line 374:
```suggestion
		tc.pattern = "!" + tc.pattern
		tc.isNegate = true
		return tc
```
Up to you, I just don't see a reason why we need an extra copy.

cli/pack/ignore_test.go
```diff
@@ -0,0 +1,893 @@
...
+				assert.NoFileExists(t, path.Join(dst, name))
+			}
+			for _, name := range tc.copied {
+				assert.FileExists(t, path.Join(dst, name))
+			}
+		})
+	}
+}
+
+// Prepare basic test set for ignoreFilter function
```
dmyger, line 487:
```suggestion
// Prepare basic test set for ignoreFilter function.
```
Here and in other comments, please add a point at the end.

cli/pack/ignore_test.go
elhimov: I mentioned it in your [suggestion PR](https://github.com/tarantool/tt/pull/1119)

> I doubt that you mean that, but it seems this way we change the original test set at each transformation which requires us to keep in mind what is in origin at any time, keep test orders or something similar. In my opinion this approach complicates the readability and here "the less code" doesn't mean "the better".

dmyger: Here we are working with a copy, not an original test case.

elhimov: Oops, sorry, my bad. "There" I missed that the whole map copied and "here"... no excuses, just facepalm (

cli/pack/ignore.go
```diff
@@ -0,0 +1,132 @@
...
+	"path/filepath"
+	"regexp"
+	"slices"
+	"strings"
+)
+
+type ignorePattern struct {
+	re       *regexp.Regexp
+	dirOnly  bool
+	isNegate bool
```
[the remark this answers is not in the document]
elhimov: Added some comments.

cli/pack/ignore_test.go
elhimov: Fixed.

cli/pack/ignore_test.go
elhimov: Fixed.

cli/pack/ignore_test.go
```diff
@@ -0,0 +1,986 @@
...
+				assert.Truef(t, p.re.MatchString(s), "%q doesn't match %s", s, p.re.String())
+			}
+			for _, s := range tc.expectedMismatches {
+				assert.False(t, p.re.MatchString(s), "%q matches %s", s, p.re.String())
+			}
+		})
+	}
+}
+
+// Prepare basic test set for createIgnorePattern function
```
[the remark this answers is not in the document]
elhimov: Fixed this comment and added detailed comments about preparation.

Issue #1064, opened 2024-12-11
`tt connect` console ignores syntax errors


locker:
**Bug description**

In contrast to the built-in Tarantool `console` module, `tt connect` console doesn't raise any error if the input string contains syntax errors. Instead it returns `null`.

**Version**

Tarantool CLI version 2.6.0, linux/amd64. commit: 96db0688623b (v2.6.0-10-g96db0688623b)

**Steps to reproduce**

1. Create a test instance.
   ```sh
   tt init
   tt create single_instance --name test
   tt start
   ```

2. Connect to the console.
   ```sh
   tt connect test:instance001
   ```
3. Type an invalid Lua string.
   ```
   hello)
   ```

**Actual behavior**

```
test:instance001> hello)
---
- null
...
```

**Expected behavior**

```
test:instance001> hello)
---
- error: '[string "hello)"]:1: ''='' expected near '')'''
...
```

PR #1110, opened 2025-02-24
changelog: fix entries about Lua parse error

d-enk:
Entries about Lua parse error was rebased into wrong place.

Follow up #1064


oleg-jukovec (approved):
Thank you for the fix!

dmyger (approved):

PR #1112, opened 2025-03-03
tls: build without openssl linkage

### LINKED ISSUE #1101: build: return ability to build without OpenSSL linking
We already had the ability to build `tt` without linking with `OpenSSL` and with CGO_ENABLED=0:
https://github.com/tarantool/tt/blob/648f98a29f597175a59f7f1166541d030eb7b87d/magefile.go#L67

But it was broken with the `go-tarantool/v2` by directly usage of the `go-tlsdialer`. We could return the ability to build without OpenSSL with a build tag inside `tt` (and optional usage of the `go-tlsdialer`). See `go-tarantool/v1` as an example:

https://github.com/tarantool/go-tarantool/blob/v1.12.2/ssl.go#L1-L2
https://github.com/tarantool/go-tarantool/blob/6843830d225bd5dc7c5ec50acc6f02bb73ab5c56/ssl_disable.go#L1-L2

patapenka-alexey:
Closes #1101

### DIFF
--- a/cli/cluster/cluster.go
+++ b/cli/cluster/cluster.go
@@ -8,9 +8,9 @@ import (
 	"time"
 
 	"github.com/tarantool/go-tarantool/v2"
-	"github.com/tarantool/go-tlsdialer"
 	libcluster "github.com/tarantool/tt/lib/cluster"
 	"github.com/tarantool/tt/lib/connect"
+	"github.com/tarantool/tt/lib/dial"
 )
 
 const (
@@ -96,35 +96,30 @@ func collectTarantoolConfig(collectors libcluster.CollectorFactory,
 			network, address = connect.ParseBaseURI(endpoint.Uri)
 		}
 		addr := fmt.Sprintf("%s://%s", network, address)
-		if endpoint.Params.Transport == "" || endpoint.Params.Transport != "ssl" {
-			opts = append(opts, tarantoolOpts{
-				addr: addr,
-				dialer: tarantool.NetDialer{
-					Address:  addr,
-					User:     endpoint.Login,
-					Password: endpoint.Password,
-				},
-				opts: tarantool.Opts{
-					SkipSchema: true,
-				},
-			})
-		} else {
-			opts = append(opts, tarantoolOpts{
-				addr: addr,
-				dialer: tlsdialer.OpenSSLDialer{
-					Address:     addr,
-					User:        endpoint.Login,
-					Password:    endpoint.Password,
-					SslKeyFile:  endpoint.Params.SslKeyFile,
-					SslCertFile: endpoint.Params.SslCertFile,
-					SslCaFile:   endpoint.Params.SslCaFile,
-					SslCiphers:  endpoint.Params.SslCiphers,
-				},
-				opts: tarantool.Opts{
-					SkipSchema: true,
-				},
-			})
+
+		dialer, err := dial.New(dial.Opts{
+			Address:         addr,
+			User:            endpoint.Login,
+			Password:        endpoint.Password,
+			SslKeyFile:      endpoint.Params.SslKeyFile,
+			SslCertFile:     endpoint.Params.SslCertFile,
+			SslCaFile:       endpoint.Params.SslCaFile,
+			SslCiphers:      endpoint.Params.SslCiphers,
+			SslPassword:     endpoint.Params.SslPassword,
+			SslPasswordFile: endpoint.Params.SslPasswordFile,
+			Transport:       endpoint.Params.Transport,
+		})
+		if err != nil {
+			return nil, err
 		}
+
+		opts = append(opts, tarantoolOpts{
+			addr:   addr,
+			dialer: dialer,
+			opts: tarantool.Opts{
+				SkipSchema: true,
+			},
+		})
 	}
 
 	var connectionErrors []error
--- a/lib/dial/dial.go
+++ b/lib/dial/dial.go
@@ -1,0 +1,37 @@
+package dial
+
+import (
+	"fmt"
+
+	"github.com/tarantool/go-tarantool/v2"
+)
+
+// New creates new dialer according to the options.
+func New(opts Opts) (tarantool.Dialer, error) {
+	transport, err := Parse(opts.Transport)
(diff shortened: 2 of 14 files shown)


lib/dial/opts.go
```diff
@@ -0,0 +1,41 @@
...
+const (
+	TransportDefault Transport = ""
+	TransportPlain   Transport = "plain"
+	TransportSSL     Transport = "ssl"
+	TransportInvalid Transport = "invalid"
+)
+
+func ConvertTransport(tr string) Transport {
+	switch tr {
+	case "":
```
oleg-jukovec, line 19: Or we could do something like that to avoid duplication:
```suggestion
	case TransportDefault.String():
```

lib/dial/opts.go
```diff
@@ -0,0 +1,41 @@
...
+	User            string
+	Password        string
+	SslKeyFile      string
+	SslCertFile     string
+	SslCaFile       string
+	SslCiphers      string
+	SslPassword     string
+	SslPasswordFile string
+	Transport       string
+}
```
oleg-jukovec, line 41: Please, move it into opts.go file.

lib/dial/opts.go
```diff
@@ -0,0 +1,41 @@
...
+	case "":
+		return TransportDefault
+	case "plain":
+		return TransportPlain
+	case "ssl":
+		return TransportSSL
+	default:
+		return TransportInvalid
+	}
+}
```
oleg-jukovec, line 28: Please, move it into transport.go.

cli/cluster/cluster.go
```diff
@@ -96,35 +96,29 @@ func collectTarantoolConfig(collectors libcluster.CollectorFactory,
...
+			Address:         addr,
+			User:            endpoint.Login,
+			Password:        endpoint.Password,
+			SslKeyFile:      endpoint.Params.SslKeyFile,
+			SslCertFile:     endpoint.Params.SslCertFile,
+			SslCaFile:       endpoint.Params.SslCaFile,
+			SslCiphers:      endpoint.Params.SslCiphers,
+			SslPassword:     endpoint.Params.SslPassword,
+			SslPasswordFile: endpoint.Params.SslPasswordFile,
+			Transport:       endpoint.Params.Transport, // [ssl|plain]
```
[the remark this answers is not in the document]
patapenka-alexey: Cause this is `draft` PR.

cli/cluster/cluster.go
```diff
@@ -96,35 +96,29 @@ func collectTarantoolConfig(collectors libcluster.CollectorFactory,
...
+		})
+		if err != nil {
+			return nil, err
 		}
+		opts = append(opts, tarantoolOpts{
+			addr:   addr,
+			dialer: dialer,
+			opts: tarantool.Opts{
```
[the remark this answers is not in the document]
patapenka-alexey: `go fmt` removes this space.

lib/dial/transport.go
```diff
@@ -0,0 +1,38 @@
...
+
+import "fmt"
+
+const (
+	transportDefault string = ""
+	transportPlain   string = "plain"
+	transportSsl     string = "ssl"
+)
+
+//go:generate stringer -type Transport -trimprefix Transport -linecomment
```
oleg-jukovec, line 11: Please, add a test for the String() function to make sure that everything work fine:

https://github.com/tarantool/go-discovery/blob/master/eventtype_test.go

You could also add tarantport_string.go to the commit. The stringer-generated code usually included into a commit.

lib/dial/dial.go
```diff
@@ -0,0 +1,37 @@
+package dial
+
+import (
+	"fmt"
+
+	"github.com/tarantool/go-tarantool/v2"
+)
+
+// Dial creates new dialer according to the options.
+func New(opts Opts) (tarantool.Dialer, error) {
```
oleg-jukovec, line 10: We could add unit-tests here. Something like:

```Go
func TestNew(t *testing.T) {
    cases := []struct{
        name     string
        opts        dial.Opts
        expected tarantool.DIaler
    } {
        {
            Name: "simple",
            Opts: dial.Opts{foo},
            expected: tarantool.NetDialer{foo},
        }
        ...
    }
    
    for _, tc := range cases {
        dialer, err := dial.New(tc.opts)

        require.NoError(t, err)
        require.Equal(t, tc.expected, dialer)
    } 
}
```

For simplicity we can test this only with ssl enabled. But you can make two different test files. See
https://github.com/tarantool/go-tarantool/blob/461ad486865e0f401a46568a688ec0851b2aa020/ssl_test.go#L1-L2

lib/dial/transporttype_test.go
oleg-jukovec, line 1: Please, rename it -> `transport_test.go`

lib/dial/transport.go
```diff
@@ -0,0 +1,50 @@
...
+func (t Transport) String() string {
+	switch t {
+	case TransportDefault:
+		return transportDefault
+	case TransportPlain:
+		return transportPlain
+	case TransportSsl:
+		return transportSsl
+	default:
+		panic("Unknown Transport")
```
oleg-jukovec, line 48: Please, avoid panics. Return something like a usual String() generated by a stringer return:
```suggestion
		return fmt.Sprintf("Transport(%d)", t)
```

lib/dial/transporttype_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+func TestTransportType_String(t *testing.T) {
+	cases := []struct {
+		Transport Transport
+		Expected  string
+	}{
+		{TransportDefault, ""},
+		{TransportPlain, "plain"},
+		{TransportSsl, "ssl"},
+	}
```
oleg-jukovec, line 17: Don't forget to add a case with unsupported value here.

oleg-jukovec (approved):
LGTM, but required a few non-critical improvments.

lib/dial/transporttype_test.go
oleg-jukovec: ```Transport(15).String()```

AlexandrLitkevich (approved):

PR #1107, opened 2025-02-19
feat: added support for new {{ metricsPort }} construct

MaratKarimov:
The configuration template generation functionality has been extended:
- Previously, the {{ port }} construct was available for generating cluster configuration files with different ports.
- A new construct, {{ metricsPort }}, has been added for generating the port of the metrics/metrics-export-role module.

This allows for more flexible configuration of ports for instances.

PR #1120, opened 2025-03-10
feat: added support for new {{ metricsPort }} construct

dmyger:
The configuration template generation functionality has been extended:

- Previously, the `{{ port }}` construct was available for generating cluster configuration files with different ports.
- A new construct, `{{ metricsPort }}`, has been added for generating the port of the metrics/metrics-export-role module.

This allows for more flexible configuration of ports for instances.

Based on PR: https://github.com/tarantool/tt/pull/1107


oleg-jukovec (approved):
There is 2 commits instead 1.

dmyger:
> There is 2 commits instead 1.

Squashed

AlexandrLitkevich (approved):

PR #1123, opened 2025-03-10
cluster: moved cmd.UriOpts to connect

dmyger:
Part of #TNTP-1561

### DIFF
--- a/lib/connect/uri.go
+++ b/lib/connect/uri.go
@@ -1,10 +1,15 @@
 package connect
 
 import (
+	"errors"
+	"fmt"
+	"net/url"
 	"os"
 	"path/filepath"
 	"regexp"
+	"strconv"
 	"strings"
+	"time"
 
 	"github.com/tarantool/tt/cli/connector"
 )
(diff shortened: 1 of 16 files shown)


lib/connect/uri.go
```diff
@@ -114,3 +157,109 @@ func ParseCredentialsURI(str string) (string, string, string) {
 
 	return newStr, credentialsSlice[0], credentialsSlice[1]
 }
+
+// ParseUriOpts parses options from a URI from a URL.
+func ParseUriOpts(uri *url.URL) (UriOpts, error) {
```
oleg-jukovec, line 162: Do we need the public method?

lib/connect/uri.go
```diff
@@ -114,3 +157,109 @@ func ParseCredentialsURI(str string) (string, string, string) {
...
+			return opts, err
+		}
+	}
+
+	return opts, nil
+}
+
+// ParseUrl returns a URL, nil if string could be recognized as a URL,
+// otherwise nil, an error.
+func ParseUrl(str string) (*url.URL, error) {
```
oleg-jukovec, line 233: Do we need the public method?

lib/connect/uri.go
```diff
@@ -114,3 +157,109 @@ func ParseCredentialsURI(str string) (string, string, string) {
...
+		return nil, err
+	}
+	if uri.Scheme != "" && uri.Host != "" && uri.Opaque == "" {
+		return uri, nil
+	}
+	return nil, fmt.Errorf("specified string can not be recognized as URL")
+}
+
+// CreateUriOpts parse URL string and creates appropriated UriOpts.
+func CreateUriOpts(url string) (UriOpts, error) {
```
oleg-jukovec, line 253: Or do we actually only need this method? At first glance it looks like this will be enough for us now (probably we will need to change some signatures of methods/functions to accept connect.UriOpts instead of url.URL).

oleg-jukovec (changes requested):
Please remove unnecessary public functions or add tests for them.

AlexandrLitkevich (approved):

lib/connect/uri.go
dmyger: Just to back compatible with current code in `cluster`.
It's actively using in `cli/cluster/cmd`. Firstly in one function get `url.URL` from string, then in other function convert it to the `UriOps`.
I don't want to keep copy of this function in private `cluster`'s scope, so there is the best way - do refactoring `cluster`'s code to remove such behavior.

lib/connect/uri.go
dmyger: Remove from public.

lib/connect/uri.go
dmyger: Refactored.

lib/connect/uri.go
dmyger: Done.

oleg-jukovec (approved):
Thank you for the patch!

PR #1126, opened 2025-03-13
Use Cobra functionality to validate arguments

### LINKED ISSUE #1125: Use Cobra functionality to validate arguments.
Cobra's Command has built-in functionality to perform arguments validation. Some commands use it, but some doesn't. One of the reason might be in that Cobra's validation errors are displayed as ordinary text and often look unnoticeable while general logger errors are displayed as red bold text. Consider making Cobra's errors look similar to general logger errors.

elhimov:
Closes #1125

### DIFF
--- a/cli/cmd/binaries.go
+++ b/cli/cmd/binaries.go
@@ -12,6 +12,12 @@ import (
 	"golang.org/x/exp/slices"
 )
 
+var binariesSupportedPrograms = []string{
+	search.ProgramCe,
+	search.ProgramEe,
+	search.ProgramTt,
+}
+
 // NewBinariesCmd creates binaries command.
 func NewBinariesCmd() *cobra.Command {
 	var binariesCmd = &cobra.Command{
@@ -42,6 +48,7 @@ You will need to choose version using arrow keys in your console.
 				internalSwitchModule, args)
 			util.HandleCmdErr(cmd, err)
 		},
+		Args: cobra.MatchAll(cobra.MaximumNArgs(2), binariesSwitchValidateArgs),
 	}
 	var listCmd = &cobra.Command{
 		Use:   "list",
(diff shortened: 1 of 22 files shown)


cli/cmd/replicaset.go
```diff
@@ -813,19 +813,25 @@ func getOrchestrator() (replicaset.Orchestrator, error) {
 	return orchestrator, nil
 }
 
-// internalReplicasetRebootstrapModule is a "rebootstrap" command for the replicaset module.
-func internalReplicasetRebootstrapModule(cmdCtx *cmdcontext.CmdCtx, args []string) error {
+// replicasetRebootstrapValidateArgs validates "rebootstrap" command arguments.
+func replicasetRebootstrapValidateArgs(cmd *cobra.Command, args []string) error {
 	if len(args) > 1 {
-		return util.NewArgError("only one instance supported for re-bootstrap")
+		return fmt.Errorf("only one instance supported for re-bootstrap")
```
patapenka-alexey, line 819:
```suggestion
		return errors.New("only one instance supported for re-bootstrap")
```

cli/cmd/replicaset.go
```diff
@@ -813,19 +813,25 @@ func getOrchestrator() (replicaset.Orchestrator, error) {
...
 	}
 	if len(args) < 1 {
-		return util.NewArgError("instance for rebootstrap is not specified")
+		return fmt.Errorf("instance for rebootstrap is not specified")
```
patapenka-alexey, line 822: The same.

cli/cmd/replicaset.go
```diff
@@ -813,19 +813,25 @@ func getOrchestrator() (replicaset.Orchestrator, error) {
...
+	}
+	if !strings.Contains(args[0], string(running.InstanceDelimiter)) {
+		return fmt.Errorf("an instance name is not specified. Please use app:instance format")
```
patapenka-alexey, line 825: The same.

cli/cmd/replicaset.go
elhimov: Done.

cli/cmd/replicaset.go
elhimov: Done.

cli/cmd/replicaset.go
```diff
@@ -813,19 +813,25 @@ func getOrchestrator() (replicaset.Orchestrator, error) {
...
+	return nil
+}
 
+// internalReplicasetRebootstrapModule is a "rebootstrap" command for the replicaset module.
+func internalReplicasetRebootstrapModule(cmdCtx *cmdcontext.CmdCtx, args []string) error {
 	appName, instName, found := strings.Cut(args[0], string(running.InstanceDelimiter))
 	if !found {
-		return util.NewArgError(
-			"an instance name is not specified. Please use app:instance format.")
+		panic("Instance delimiter is missing unexpectedly.")
```
patapenka-alexey, line 834:
```suggestion
		return util.NewArgError("Instance delimiter is missing unexpectedly.")
```
Why should we trigger panic here?

elhimov: It's a kind of assert. Function that validate arguments must make this `panic` unreachable.

patapenka-alexey (approved):

cli/cmd/replicaset.go
elhimov: I've removed this `panic` just to keep consistent code (there are no such "assertions" in other places).

cli/cmd/root.go
```diff
@@ -75,35 +75,51 @@ func GetModulesInfoPtr() *modules.ModulesInfo {
...
-type LogHandler struct{}
+// setWriter sets the underlying writer and returns the original one.
+func (h *LogHandler) setWriter(w io.Writer) io.Writer {
+	orig := h.baseHandler.Writer
+	h.baseHandler.Writer = w
+	return orig
+}
 
-var defaultLogHandler = &LogHandler{}
+// LogErrorWriterDecorator is used to decorate messages sent to writer
```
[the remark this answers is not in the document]
elhimov: Fixed.

patapenka-alexey (approved):

cli/cmd/binaries.go
```diff
@@ -57,44 +64,42 @@ You will need to choose version using arrow keys in your console.
...
-	switch len(args) {
-	case 2:
-		switchCtx.Version = args[1]
-		switchCtx.ProgramName = args[0]
-		if !slices.Contains(supportedPrograms, switchCtx.ProgramName) {
-			return fmt.Errorf("not supported program: %s", switchCtx.ProgramName)
-		}
-	case 1:
+
+	if len(args) > 0 {
```
dmyger, line 85: As for me, code with `switch len(args) {` was a bit clear to understand and support.

cli/download/download.go
```diff
@@ -16,7 +16,7 @@ import (
 
 type DownloadCtx struct {
 	// SDK version to download.
-	version string
+	Version string
```
dmyger, line 19: Please update comment for the field, too. It should start with the field name.

cli/cmd/binaries.go
oleg-jukovec: Yes, it is. Switch instead of multiple if-else it is a Golang pattern.

cli/download/download.go
elhimov: Done.

cli/cmd/binaries.go
elhimov: As for me this approach is simpler. 
Consider them as 2 independent if-else. Each of them just deals with single corresponding parameter.
We don't need to validate program here because its validation was moved to the dedicated validation function (as well as we don't need to duplicate that validation code with error handling in two cases). All we need here is to deal with 2 parameters and it could be done with 2 independent if-else.

oleg-jukovec (approved):

cli/cmd/binaries.go
dmyger: Let's close the thread so we don't waste time arguing.
Your arguments still don't answer for me why unreadable code is better than readable code.

dmyger (approved):

cli/cmd/binaries.go
elhimov: I'm not trying to argue that unreadable is better than readable. Obviously it's not. But readability might strongly depends on point of view. So I'm just trying to explain different point of view, from which it seems more readable for me (and easier to maintain as well).

PR #1127, opened 2025-03-13
 Add a command the start TCM

AlexandrLitkevich:


test/integration/tcm/test_tcm.py
```diff
@@ -0,0 +1,77 @@
...
+            tcm.wait()
+            break
+        if output:
+            print(output)
+
+    assert tcm.poll() is not None
+
+
+def test_tcm_start_with_watchdog_success(tt_cmd):
+    tcmIsBunary = is_tarantool_ee()
```
patapenka-alexey, line 44:
```suggestion
    skip_if_tarantool_ce()
```

AlexandrLitkevich: Thanks, done

patapenka-alexey (approved):

tcm/pidFile.pid
```diff
@@ -0,0 +1 @@
+44947
```
AlexandrLitkevich, line 1: must be deleted

cli/tcm/watchdog.go
dmyger, line 1: 1) What is the difference of this solution from `cli/running/watchdog.go`?
2) Is it possible here reuse an existing one?
3) May be it would be better to improve current solution to make them more reusable, than create a new?

test/integration/tcm/test_tcm.py
```diff
@@ -0,0 +1,70 @@
...
+        output = tcm.stdout.readline().strip()
+
+        if not output and tcm.poll() is not None:
+            break
+        if "TCM_CLUSTER_CONNECTION_RATE_LIMIT" in output:
+            tcm.terminate()
+            tcm.wait()
+            break
+        if output:
+            print(output)
```
dmyger, line 34:
```suggestion
from utils import wait_for_lines_in_output
# ...
    wait_for_lines_in_output(tcm.stdout, ["TCM_CLUSTER_CONNECTION_RATE_LIMIT"])
    tcm.terminate()
    tcm.wait()
```

test/integration/tcm/test_tcm.py
```diff
@@ -0,0 +1,70 @@
...
+
+    tcm = Popen(
+        cmd,
+        text=True,
+        encoding="utf-8",
+        stdout=PIPE,
+        stderr=PIPE,
+    )
+
+    while True:
```
dmyger, line 53: Please grep `Python` tests for using `wait_for_lines_in_output` utils function.

cli/tcm/watchdog.go
AlexandrLitkevich: The existing watchdog is tightly coupled with tarantool, it runs instances which is not required with TCM.Whether it is worth tweaking I think needs to be discussed.

test/integration/tcm/test_tcm.py
AlexandrLitkevich: Thanks

test/integration/tcm/test_tcm.py
AlexandrLitkevich: Done

PR #1154, opened 2025-04-17
search: refactoring search code

dmyger:
Split search code into separate files for each search type [git, local, sdk].

Part of #TNTP-1094

- [ ] Don't forget about TarantoolBot in a commit message (see [example][tarantoolbot-example])
- [ ] Tests (see [documentation][go-testing] for a testing package)
- [ ] Changelog (see [documentation][keepachangelog] for changelog format)
- [ ] Documentation (see [documentation][go-doc] for documentation style guide)

Related issues:

Related to #XXX

Part of #XXX

Closes #XXX


oleg-jukovec:
Please, update the PR title and message.

cli/search/search.go
```diff
@@ -37,178 +32,24 @@ type SearchCtx struct {
...
-
-	sort.Stable(version.VersionSlice(versions))
-
-	return versions, nil
+// NewSearchCtxWithMock creates a new SearchCtx with a mock handlers for testing.
+func NewSearchCtxWithMock(informer platformInformer, doer apiDoer) SearchCtx {
+	sCtx := NewSearchCtx()
+	sCtx.platformInformer = informer
+	sCtx.apiDoer = doer
+	return sCtx
```
oleg-jukovec, line 52: To be honest, I am against the approach. It should be possible to test a public API through this public API without additional methods. Otherwise, there is no point in all this stuff and you can test a private API.

It is better to change the current public API so that it is oriented to dependency injection and you will not require the test-helper method.

cli/search/search.go
```diff
@@ -37,178 +32,24 @@ type SearchCtx struct {
...
 }
 
-// GetVersionsFromGitRemote returns sorted versions list from specified remote git repo.
-func GetVersionsFromGitRemote(repo string) ([]version.Version, error) {
-	versions := []version.Version{}
-
-	if _, err := exec.LookPath("git"); err != nil {
-		return nil, fmt.Errorf("'git' is required for 'tt search' to work")
+// NewSearchCtx creates a new SearchCtx with default production values.
+func NewSearchCtx() SearchCtx {
```
oleg-jukovec, line 41: There are no tests for the call.

cli/search/search.go
```diff
@@ -235,271 +76,34 @@ func printVersion(bindir string, program string, versionStr string) {
...
-	cliOpts *config.CliOpts, program string) error {
-	var repo string
-	versions := []version.Version{}
-
-	if program == ProgramCe {
-		repo = GitRepoTarantool
-	} else if program == ProgramTt {
-		repo = GitRepoTT
-	}
+func SearchVersions(searchCtx SearchCtx, cliOpts *config.CliOpts) error {
```
oleg-jukovec, line 79: We could add tests here too. Since we are already here.

cli/search/search_git.go
```diff
@@ -0,0 +1,198 @@
...
+	err = cmd.Run()
+	if err != nil {
+		return "", fmt.Errorf("unable to get commits: git clone failed: %w", err)
+	}
+
+	return GetCommitFromGitLocal(tempRepoPath, input)
+}
+
+// GetVersionsFromGitLocal returns sorted versions list from specified local git repo.
+func GetVersionsFromGitLocal(repo string) (version.VersionSlice, error) {
```
oleg-jukovec, line 153: And maybe for some calls here too.

cli/search/search_sdk.go
```diff
@@ -0,0 +1,80 @@
...
+		if bundle.Version.Str == expectedVersion {
+			return bundle, nil
+		}
+	}
+	return BundleInfo{}, fmt.Errorf("expected version %s not found", expectedVersion)
+}
+
+// GetEeBundleInfo returns the available EE SDK bundle for user's OS,
+// corresponding to the passed expected version argument.
+func GetEeBundleInfo(cliOpts *config.CliOpts, local bool, devBuild bool,
```
oleg-jukovec, line 73: etc

AlexandrLitkevich (approved):

cli/search/search.go
dmyger: Yes, on the one hand it is a good idea.
But it's a bit beyond the scope of the task at hand, and requires much more resources to implement.
The main problem with this method is that it is overloaded with functionality and it searches both Git repositories and the cloud. To test it, you will have to do complex tests, and you'll have the desire to do it with "dependency injection", to be able do mock for tests.
Purely technically this method was adapted from the old code, with minimal changes in the architecture, and here, in fact, it was only added a check on TCM.
But overall, I agree that we need to keep tidying up the code. If necessary, we can create a separate task for this.

cli/search/search_git.go
dmyger: That seems a bit outside the scope of the given task, doesn't it?
TCM support does not require working with a Git repository, so this refactoring was cosmetic in order to improve the readability of the code.
But now we can see obvious problems in the code that require further work on them. I propose to create new tasks for this.

oleg-jukovec:
@dmyger tests are red.

cli/search/search_git.go
oleg-jukovec: Ok

oleg-jukovec (changes requested):
Tests are red.

dmyger:
> Please, update the PR title and message.

Done.

dmyger:
> tests are red

Now they are "green"

oleg-jukovec (approved):
Please, add example of the command run into the PR message.

dmyger:
> Please, add an example of the command run into the PR message.

Done

PR #1165, opened 2025-05-20
connect: support ipv6 uri

### LINKED ISSUE #1164: Failed to connect over IPv6 address

elhimov:

I didn't forget about:

- [x] Don't forget about TarantoolBot in a commit message (see [example][tarantoolbot-example])
- [x] Tests (see [documentation][go-testing] for a testing package)
- [x] Changelog (see [documentation][keepachangelog] for changelog format)

Related issues:

Closes #1164


CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `tt cat|play <DIR>` with directories handles only `.snap` or `.xlog` files.
 - `rs vshard bootstrap`: ignore an error and retry within `timeout` flag
   period.
+- `tt connect|replicaset`: fixed using of IPv6 in instance URI.
```
dmyger, line 35: Do these changes only affect `connect` and `replicaset`?

test/integration/connect/test_connect.py
```diff
@@ -2679,68 +2650,64 @@ def test_output_format_tables_dialects(tt_cmd, tmpdir_with_cfg):
         assert file != ""
 
         # Connect to the instance.
-        uris = ["localhost:3013"]
+        uris = ["localhost:3013", "tcp://localhost:3013"]
```
dmyger, line 2653: It seems that better did it with `pytest.mark.parametrize`, does not it?

test/integration/replicaset/test_replicaset_roles_remove.py
```diff
@@ -9,10 +9,7 @@
     get_group_by_replicaset_name, get_group_replicaset_by_instance_name,
     parse_status, parse_yml, start_application, stop_application)
 
-from utils import get_tarantool_version, read_kv, run_command_and_get_output
-
-tarantool_major_version, tarantool_minor_version = get_tarantool_version()
-
+import utils
```
dmyger, line 12: It is best to import only what is required, not the whole thing. As it was before.

test/integration/replicaset/test_replicaset_vshard.py
```diff
@@ -8,6 +8,7 @@
                               cartridge_username)
 from replicaset_helpers import eval_on_instance, parse_status, stop_application
 
+import utils
```
dmyger, line 11: The same.

test/integration/replicaset/test_replicaset_vshard.py
```diff
@@ -121,8 +122,7 @@ def have_buckets_created():
     assert wait_event(10, have_buckets_created)
 
 
-@pytest.mark.skipif(tarantool_major_version < 3,
-                    reason="skip centralized config test for Tarantool < 3")
+@utils.skipif_cluster_app_unsupported
```
dmyger, line 125: Let's not introduce new entities to implement `fixtures`

test/integration/clean/test_clean.py
```diff
@@ -174,7 +171,7 @@ def test_clean_cluster_auto_yes(tt):
 )
 
 
-@pytest.mark.skipif(skip_cluster_cond, reason=skip_cluster_reason)
+@utils.skipif_cluster_app_unsupported
```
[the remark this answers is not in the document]
elhimov: Sorry, but I don't catch the point. I used the way [described](https://docs.pytest.org/en/stable/how-to/skipping.html#id1) in pytest documentation (please, see after line "You can share skipif markers between modules"). Is it not right? As for me, it looks as simple as possible.

dmyger (changes requested):

lib/connect/uri.go
```diff
@@ -105,7 +105,8 @@ type UriOpts struct {
 func IsBaseURI(str string) bool {
 	// tcp://host:port
 	// host:port
-	tcpReStr := `(tcp://)?([\w\\.-]+:\d+)`
+	hostReStr := `(?:([\w\\.-]+)|(\[[\w:]+\]))`
```
dmyger, line 108: Here we are checking for URLs, not IP addresses. Why did you decide to add IPv6 checking here?

test/utils.py
```diff
@@ -599,24 +595,47 @@ def is_tarantool_major_one():
...
-        pytest.skip("Tuple format is supported")
+def is_raft_supported():
+    major, minor = get_tarantool_version()
+    return major >= 2
 
 
-def skip_if_tuple_format_unsupported():
-    if not is_tuple_format_supported():
-        pytest.skip("Tuple format is unsupported")
+skipif_tarantool_ce = pytest.mark.skipif(
```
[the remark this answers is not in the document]
elhimov: Again it seems there is misunderstanding. The provided link describe using external fixture in skip marker, but I'm not trying to do it, I just share and reuse skip marker (w/o introducing new fixture). Please, see [here](https://docs.pytest.org/en/stable/how-to/skipping.html#id1) after the line:

> You can share skipif markers between modules

test/integration/replicaset/test_replicaset_roles_remove.py
elhimov: As far as I know there are 2 approaches while importing with their own pros&cons and none of them is the best. The reason why I'm using this one is because with "only-required" import with every file I had to
1. change import statement
2. pass the script through isort to make sure it will pass CI
3. use it

and it is annoying, while with `import utils` I just use it.

I agree however that for the certain module either of them should be used, not both. I'll update.

lib/connect/uri.go
elhimov: > Here we are checking for URLs, not IP addresses. Why did you decide to add IPv6 checking here?

IP address is a part of URI (prior to this change IPv4 was handled here as well).

test/integration/clean/test_clean.py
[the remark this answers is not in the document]
dmyger: OK, I see their suggestion, but there's also a recommendation there to put such definitions in a separate file. And `utils` is obviously a bad option here, since it's basically a rubbish dump that shouldn't be there.

Besides, such way of solution has two disadvantages:
1) These are just alias wrappers for the `skip` function, meaning they hide the real nature of such action. You can't see which ones there are. They won't be in the output of:
```sh
pytest --markers
pytest --fixtures
```

2) You have to have them imported before using them, unlike a registered `markers`, which can be used immediately.

So, there is a better way to create `markers` that are responsible for their own work. Here are links to the same official documentation:
- https://docs.pytest.org/en/stable/example/simple.html#control-skipping-of-tests-according-to-command-line-option
- https://docs.pytest.org/en/stable/example/markers.html#custom-marker-and-command-line-option-to-control-test-runs
- https://docs.pytest.org/en/stable/how-to/fixtures.html#using-markers-to-pass-data-to-fixtures
- https://docs.pytest.org/en/stable/how-to/mark.html

You can check [pytest.ini](https://github.com/tarantool/tt/blob/master/test/pytest.ini#L4) for defined _markers_.

Any way, up to you. But only not in `utils` package. If you want an "alias" way, let's create a separate file for them.

dmyger (changes requested):

patapenka-alexey (approved):
Thank you for the patch.

test/integration/clean/test_clean.py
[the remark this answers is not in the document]
elhimov: Dropped this change (will be applied in another PR).

test/integration/connect/test_connect.py
elhimov: Dropped this change (will be applied in another PR).

test/integration/replicaset/test_replicaset_roles_remove.py
elhimov: Dropped this change (will be applied in another PR).

test/integration/replicaset/test_replicaset_vshard.py
elhimov: Dropped this change (will be applied in another PR).

test/integration/replicaset/test_replicaset_vshard.py
elhimov: Dropped this change (will be applied in another PR).

test/utils.py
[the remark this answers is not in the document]
elhimov: Dropped this change (will be applied in another PR).

lib/connect/uri.go
elhimov: Added unit tests.

elhimov:
> I suggest to update unit-test [here](https://github.com/tarantool/tt/blob/4d96391361760b155c6ed58ec0f8250e070789cb/lib/connect/uri_test.go#L36) and [here](https://github.com/tarantool/tt/blob/4d96391361760b155c6ed58ec0f8250e070789cb/lib/connect/uri_test.go#L198).

Done

CHANGELOG.md
elhimov: You are right. `lib/connect `is used differently in these commands, but the applied change affects them as well. (also it affects `play` command). I updated commit message and changelog.

PR #1171, opened 2025-06-06
play: extend error message

patapenka-alexey:
This patch extends error message if space to play data is unavailable or user does not have permission to work with it, because the `net.box` module does not have means to distinguish between these errors.

Closes #TNTP-2366


oleg-jukovec (changes requested):
Please, describe in the commit and PR message why we could not resolve the problem (I guess we couldn't detect a specific error someway). It would be nice if you add a comment into the code about it, up to you.

oleg-jukovec (approved):

PR #1173, opened 2025-06-16
ci: use pre-commit hooks to static code check

dmyger:
On `CI` github action uses scripts for static code checks from `pre-commit` hooks.

Related issues:

Closes #TNTP-2939


.github/actions/static-code-check/action.yml
```diff
@@ -4,31 +4,28 @@ description: "Performs static code checks."
...
-      - name: Go Linter
-        uses: golangci/golangci-lint-action@v3
-        with:
-          version: v1.63.4
-          args: --config=golangci-lint.yml --out-${NO_FUTURE}format colored-line-number
-          skip-cache: true
-
-      - name: Python Linter
-        run: python3 -m ruff check test
-        shell: bash
```
oleg-jukovec: Now linters run only on changes?
Let's leave full linter runs here if so.

dmyger: Added two different steps in `.github/actions/static-code-check/action.yml` to check fresh diff with strict rules and all files with a bit reduced set of rules.
And `.pre-commit-config.yaml` has appropriate linters settings.

.github/actions/static-code-check/action.yml
```diff
@@ -4,31 +4,38 @@ description: "Performs static code checks."
...
+        echo "Found base branch with hash: '${HASH}'"
+        echo "BASE_BRANCH=${HASH}" >> ${GITHUB_ENV}
 
-      - name: Go Linter
-        uses: golangci/golangci-lint-action@v3
-        with:
-          version: v1.63.4
-          args: --config=golangci-lint.yml --out-${NO_FUTURE}format colored-line-number
-          skip-cache: true
+    # FIXME: Enable golangci-lint-diff after PR #1173 is merged.
```
oleg-jukovec, line 28: So, we need a one more PR?

golint-precommit.yml
```diff
@@ -44,14 +45,14 @@ linters:
...
 
-    wsl:
-      # See https://golangci-lint.run/usage/linters/#wsl
-      # And https://github.com/bombsimon/wsl/blob/master/doc/configuration.md
-      allow-assign-and-call: false
-      allow-multiline-assign: false
-      force-case-trailing-whitespace: 2
-      allow-cuddle-used-in-block: true
-      force-err-cuddling: true
+    # wsl:
```
oleg-jukovec, line 48: Why is it disabled?

oleg-jukovec (changes requested):

.github/actions/static-code-check/action.yml
dmyger: Yes, the problem is that in order to successfully pass the `full` checks, several corrections were made to a number of source files. This, in turn, triggers the `diff` check, and these checks are stricter. Therefore, the old files need to be merged with the strict checks turned off, and then turned on again. The second PR #1178 has been added to the chain for this.

golint-precommit.yml
dmyger: Using the `wsl` linter conflicts with the `gofumpt` formatter. At the same time, it is not possible to achieve compatibility through the linter settings. The formatter, on the other hand, does not allow you to adjust its behaviour at all — everything is hardcoded there.

When both are enabled, one formats the code in its own way, then the other reformats it differently, which causes the first one to generate an error again in the next pass. It's a cyclical dependency.

_For example_, I encountered the following problem: the formatter expects that after performing any action and subsequently checking the result for errors, there should be no empty lines. And it always finds such places correctly. At the same time, `wsl` often makes mistakes and does not always find such places correctly. Issues arise in the code when the check is a bit different from the simple case `if(err != nil){...}`, where there may be something more complex like `if...else` or `error` been checked for type. In such cases, the formatter still considers that this should go together without an extra empty line, while the linter move such a check in a separate logical block.

patapenka-alexey (approved):

PR #1175, opened 2025-06-18
logrotate: fix log rotation

### LINKED ISSUE #1174: logrotate: log file descriptor is not released after log rotation

elhimov:
Watchdog launches tarantool in a way that produce log to stderr and thus tarantool itself does nothing to handle log rotation, it is handled entirely at watchdog side.
Prior to this patch `os.File` was used in `Cmd.Stdout`, when watchdog launches tarantool binary and according to golang [documentation](https://pkg.go.dev/os/exec#Cmd) (see `Exec.Cmd.Stdout`) in this special case output from the child process is connected directly to that file. It means that even if it is closed at watchdog side, its child process (tarantool) keeps using the original one, i.e. it proceeds with logging to the original file even after renaming it.
Now `Logger` is used as `Cmd.Stdout`, and according to the mentioned documentation separate goroutine reads output from the child process over a pipe and delivers that data to the `Logger`. It allows proper handling of log rotation at watchdog side.

@TarantoolBot document
Title: `tt logrotate` properly release descriptor of the old log file.

I didn't forget about:

- [x] Don't forget about TarantoolBot in a commit message (see [example][tarantoolbot-example])
- [x] Tests (see [documentation][go-testing] for a testing package)
- [x] Changelog (see [documentation][keepachangelog] for changelog format)

Related issues:

Closes #1174

### DIFF
--- a/cli/running/base_instance.go
+++ b/cli/running/base_instance.go
@@ -110,8 +110,8 @@ func StdErrOpt(writer io.Writer) InstanceOption {
 func StdLoggerOpt(logger ttlog.Logger) InstanceOption {
 	return func(inst *baseInstance) error {
 		inst.logger = logger
-		inst.stdOut = logger.Writer()
-		inst.stdErr = logger.Writer()
+		inst.stdOut = logger
+		inst.stdErr = logger
 		return nil
 	}
 }
--- a/cli/ttlog/logger.go
+++ b/cli/ttlog/logger.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"sync"
 )
 
 const (
@@ -28,7 +29,7 @@ type Logger interface {
 	Println(v ...any)
 	Fatal(v ...any)
 	Fatalf(format string, v ...any)
-	Writer() io.Writer
+	Write(p []byte) (int, error)
 
 	// Rotate re-opens a log file.
 	Rotate() error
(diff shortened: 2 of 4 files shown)


cli/running/base_instance.go
```diff
@@ -110,8 +110,8 @@ func StdErrOpt(writer io.Writer) InstanceOption {
 func StdLoggerOpt(logger ttlog.Logger) InstanceOption {
 	return func(inst *baseInstance) error {
 		inst.logger = logger
-		inst.stdOut = logger.Writer()
-		inst.stdErr = logger.Writer()
+		inst.stdOut = logger
+		inst.stdErr = logger
```
oleg-jukovec, line 114: https://github.com/tarantool/tt/blob/da042d0ad022712948851a7bf2acfc8f9561b35f/cli/running/running.go#L761-L762

But stdOut may be overwritten. Is it problem?

oleg-jukovec (changes requested):

oleg-jukovec (approved):

oleg-jukovec:
Tests are red.

cli/ttlog/logger.go
```diff
@@ -102,12 +111,22 @@ func (logger *fileLogger) GetOpts() LoggerOpts {
 	return logger.opts
 }
 
+// Write implements io.Writer interface
+func (logger *fileLogger) Write(p []byte) (int, error) {
+	logger.mu.Lock()
+	defer logger.mu.Unlock()
+	return logger.Logger.Writer().Write(p)
```
oleg-jukovec: It is more canonical formatting.
```suggestion
	defer logger.mu.Unlock()

	return logger.Logger.Writer().Write(p)
```

elhimov: Corrected.

PR #1184, opened 2025-07-08
spell: turn on pre-commit auto spell checker

dmyger:
Turning on automatic static code analyzer: check words misspell for code files.

Closes #TNTP-3105

Come with second part in #1185


oleg-jukovec (approved):

.github/actions/static-code-check/action.yml
```diff
@@ -25,12 +25,12 @@ runs:
         echo "Found base branch with hash: '${HASH}'"
         echo "BASE_BRANCH=${HASH}" >> ${GITHUB_ENV}
 
-    - name: pre-commit checks (diff)
-      uses: pre-commit/action@v3.0.1
-      env:
-        SKIP: golangci-lint-full
-      with:
-        extra_args: --all-files --from-ref=${{ env.BASE_BRANCH }} --to-ref=HEAD --hook-stage=manual
+    # - name: pre-commit checks (diff)
```
elhimov, line 28: It looks a bit strange that this PR disables this job and the next PR (#1185) enables it back and it is the only change #1185 does. Why don't we just keep it intact here, so #1185 would not be needed at all.

.cspell_project-words.txt
```diff
@@ -1,48 +1,68 @@
...
+containsf
+debugf
+demoter
+disbalance
+distfile*
+drwxr
+equalf
+errf
+etcd*
+etcdserver
```
elhimov, line 22: Looks unnecessary as previous entry matches it.

.cspell_project-words.txt
```diff
@@ -1,48 +1,68 @@
...
-testfull
-TNTP
-ttable
-venv
+svacer
+tarantool*
+tlsdialer
+tmpl
+tnt+
+tntp
```
elhimov, line 61: Looks unnecessary as previous entry matches it.

lib/cluster/integration_test.go
```diff
@@ -145,7 +145,7 @@ func startEtcd(t *testing.T, opts etcdOpts) integration.LazyCluster {
 func etcdPut(t *testing.T, etcd *clientv3.Client, key, value string) {
 	t.Helper()
 	var (
-		presp *clientv3.PutResponse
+		presp *clientv3.PutResponse // spell-checker:ignore presp
```
elhimov: In 'cli/cluster/integration_test.go' the same identifier changed to `pResp`. Please consider single approach.

lib/cluster/integration_test.go
```diff
@@ -273,10 +273,10 @@ func TestConnectEtcd(t *testing.T) {
 			etcdPut(t, etcd, "foo", "bar")
 
 			ctx, cancel := context.WithTimeout(context.Background(), timeout)
-			gresp, err := etcd.Get(ctx, "foo")
+			resp, err := etcd.Get(ctx, "foo")
```
elhimov: In 'cli/cluster/integration_test.go' the similar identifier `presp` changed to `pResp`. Please consider similar approach here.

lib/cluster/etcd.go
```diff
@@ -502,6 +502,7 @@ func (publisher EtcdAllDataPublisher) Publish(revision int64, data []byte) error
 			txn = txn.If(cmps...)
 		}
 
+		// spell-checker:ignore tresp
```
elhimov: In 'cli/cluster/integration_test.go' the similar identifier `presp` changed to `pResp`. Please consider similar approach here.

lib/cluster/etcd.go
```diff
@@ -849,12 +850,12 @@ func СonnectEtcdUriOpts(
 func DoOnStorage(connOpts ConnectOpts, opts libconnect.UriOpts,
 	tarantoolFunc func(tarantool.Connector) error, etcdFunc func(*clientv3.Client) error,
 ) error {
-	etcdcli, errEtcd := СonnectEtcdUriOpts(opts, connOpts)
+	etcdcli, errEtcd := ConnectEtcdUriOpts(opts, connOpts)
```
elhimov, line 852: So it turns out that we had public identifier with cyrillic symbol. Nice catch! )

lib/cluster/etcd.go
```diff
@@ -849,12 +850,12 @@ func СonnectEtcdUriOpts(
...
 	if errEtcd == nil {
 		return etcdFunc(etcdcli)
 	}
 
-	conn, errTarantool := СonnectTarantool(opts, connOpts)
+	conn, errTarantool := ConnectTarantool(opts, connOpts)
```
elhimov, line 857: And it's not alone! )))

test/integration/replicaset/test_replicaset_promote.py
```diff
@@ -155,7 +155,7 @@ def test_promote_cconfig_failovers(
                 "manual-failover-2",
                 "election-failover-1",
                 "election-failover-2",
-                "eleciton-failover-3",
+                "eleciton-failover-3",  # spell-checker:disable-line
```
elhimov, line 158:
```suggestion
                "election-failover-3",
```

test/integration/completion/conftest.py
```diff
@@ -44,7 +46,7 @@ def skip_no_helpers(request: pytest.FixtureRequest, completion: Completion) -> N
 
 
 def pytest_configure(config):
-    config.addinivalue_line(
+    config.addinivalue_line(  # cSpell:words addinivalue
```
elhimov: If it's possible I'd suggest to introduce separate project-words file for integration tests directory. All pytest identifiers that are currently in use could be listed there.

elhimov (changes requested):

.github/actions/static-code-check/action.yml
dmyger: Checking “all” files (including old code) is done with lightened requirements, but even this required minor changes. However, any changes are treated as “new” code. The purpose of this PR is to enable rules, not to fix problems in old code.
The step that is turned off here is just checking for changes to the code. Since any new code is checked under stricter rules.

.cspell_project-words.txt
dmyger: No, it doesn't work that way.
`XXX*` can be joined only with `*YYY` patterns.

.cspell_project-words.txt
dmyger: See, comment about `etcd*`

test/integration/replicaset/test_replicaset_promote.py
dmyger: No, the test will fail in this case.
The purpose of this PR is not to fix problematic code, but to enable spell checking.

test/integration/completion/conftest.py
dmyger: My opinion is that lists of “bad” words are workaround, because they should not exist. The fact that now a list of words is made at the project level is because there are too many such words already found throughout the old code. Ideally, the names should immediately meet the requirements of the checker.

And such marks raise a “red” flag - that something needs to be done here with refactoring. The word list on other hand hides this problem.

lib/cluster/integration_test.go
dmyger: `pResp`

lib/cluster/integration_test.go
dmyger: What will the letter `p` in this variable mean?
I don't see the need for it in this case.

lib/cluster/etcd.go
dmyger: `tResp`

lib/cluster/integration_test.go
elhimov: By the similar approach I mean that if `presp` is changed to `pResp` then `gresp` should be changed to `gResp` (AFAIU pResp stands for PUT response and `gResp` stands for GET response).

test/cartridge_helper.py
```diff
@@ -162,7 +162,7 @@ def __init__(self, workdir, tt_cmd) -> None:
     def truncate(self, bootstrap_vshard=True):
         self.stop()
         shutil.rmtree(os.path.join(self.workdir, cartridge_name, var_path))
-        self.start(boostrap_vshard=bootstrap_vshard)
+        self.start(bootstrap_vshard=bootstrap_vshard)
```
elhimov, line 165: Comment just to be able to refer this change.

test/integration/replicaset/test_replicaset_promote.py
elhimov: There are tons of similar corrections in this PR and even in this commit ([example](https://github.com/tarantool/tt/pull/1184/commits/fcaafe1e199fc103733bf582b3d8252a5f2708eb#r2228264919)) and I don't see any difference against this one. Don't they fail?

.pre-commit-config.yaml
```diff
@@ -123,16 +115,16 @@ repos:
           - github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.1.6
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.12.0
+    rev: v0.12.2
     hooks:
-      - id: ruff-check
-        name: "Py: Run the linter"
-        args: [--fix]
+      - id: ruff-format
```
elhimov, line 120: Is there any reason to swap them (check <-> format)?

test/integration/completion/conftest.py
elhimov: Well... I agree with statement that "mark is better than hide", but what is need to be done with "third-party"  identifiers (pytest identifier in this particular case) considering that we are not able to change them?

.github/actions/static-code-check/action.yml
elhimov: Got it, thanks!

.pre-commit-config.yaml
dmyger: Yes, they both do code formatting. And in this order it's a bit more stable in use.

test/integration/completion/conftest.py
dmyger: What exactly identifier do you mean?

test/integration/replicaset/test_replicaset_promote.py
dmyger: I've spent a few hours trying to fix it, and this way seems like the most sensible way to do it so I don't have to spend extra time on it.

lib/cluster/integration_test.go
dmyger: `gResp`

test/integration/completion/conftest.py
elhimov: addinivalue_line

test/integration/replicaset/test_replicaset_promote.py
elhimov: Well then OK.

test/integration/completion/conftest.py
dmyger: In the current code it used only once at `test/integration/completion/conftest.py:49`? So I don't see a problem add a single comment to ignore it `spell-checker:ignore ...`

elhimov: And according to you it raises a "red" flag and something needs to be done. But since it is third-party identifier we are not able to do "something" with it (except of stop using it, but that's not the option obviously). Considering this project-words file might hold only "bad" words that can't be fixed.

elhimov: Anyway it's just a subject to discuss. I've approved PR already.

PR #1185, opened 2025-07-10
ci: turn on linters

dmyger:
Turn on static code analyzes.

Second part for #1184

Part of #TNTP-3105


elhimov (changes requested):
Sorry, but I don't catch the point of splitting into 2 PRs. The first one (#1184) turns off the job and the second one (this PR) turns it back. I would understand if this job fails in #1184 and then I would expect that it is fixed with this PR and turned back, but I see no fix, it just turned back. Would you please clarify? Why don't we just keep this job intact in #1184 and drop this PR?

PR #1191, opened 2025-08-04
cd: add CentOS 8

Totktonada:
Similarly to commit 84850608a5c2 ("cd: add Ubuntu Noble") it is sometimes convenient to have the `tt` package right in the distro specific repository even if it is available in the general `linux-rpm/static` repository.

Since it costs nothing (the same RPM file is published to different repositories), I propose to publish the `tt` package to the `el/8` repository as well.

See tarantool/installer.sh#42 for the story behind it.


Totktonada:
The Tests/check-commit-msg job is failed:

```
Spell checking commit message............................................Passed
Verify commit message format.............................................Failed
- hook id: gitlint
- exit code: 1

13: B8:reference_task_id Body does not match regex ((?m)^(?:Closes|Fixes|Related to|Part of|Needed for) #(?:[A-Z]+-)?\d+$)
```

The body doesn't actually match the regex. Should I link the installer.sh issue using the 'Related to' or 'Needed for' tag instead of a free form link?

dmyger:
> The body doesn't actually match the regex.

The checker expecting find in commit message a reference for task. It should be a line likes one of following:
```
Closes #123
Fixes #TNTP-123
Related to #123
Part of #TNTP-123
Needed for #123
```

dmyger (approved):

oleg-jukovec (approved):

PR #1197, opened 2025-08-11
switch: support TcS as config storage

### LINKED ISSUE #1193: `tt cluster failover switch` doesn't work with Tarantool Config Storage
_Duplicate of TNTP-3611_

elhimov:

- [ ] Don't forget about TarantoolBot in a commit message (see [example][tarantoolbot-example])
- [ ] Tests (see [documentation][go-testing] for a testing package)
- [ ] Changelog (see [documentation][keepachangelog] for changelog format)

Related issues:

Closes #1193


cli/cluster/cmd/failover.go
```diff
@@ -107,56 +91,44 @@ func Switch(url string, switchCtx SwitchCtx) error {
 
 	uuid := uuid.New().String()
 	key := uriOpts.Prefix + failoverPath + uuid
+	log.Infof("Switch: key=%s", key)
+	log.Infof("Switch: val=%s", string(yamlCmd))
```
dmyger: Is this for debugging - did you forget to delete it, or is it for production?

cli/cluster/cmd/failover.go
```diff
@@ -63,36 +61,22 @@ type SwitchStatusCtx struct {
...
-		CaFile:         uriOpts.CaFile,
-		SkipHostVerify: uriOpts.SkipHostVerify,
-		Timeout:        uriOpts.Timeout,
-	}
-
-	return opts
-}
-
 // Switch master instance.
 func Switch(url string, switchCtx SwitchCtx) error {
```
dmyger, line 65: Maybe we should rename it to `ctx`?

test/conftest.py
```diff
@@ -165,26 +166,41 @@ def cartridge_app(request, cartridge_app_session):
...
+        path_to_cfg_dir=request.path.parent / "test_tcs_app",
+        connection_test=True,
+        connection_test_user="client",
+        connection_test_password="secret",
+        instance_name="instance-001",
+        instance_host="localhost",
+        instance_port="4401",
+    )
+    if tcs_params is not None:
+        params.update(tcs_params)
```
dmyger: Is it correct to "update" parameters if external ones are explicitly specified?
It is probably better to use **default** values only if there is no explicit specification, otherwise the behaviour becomes unclear — you specify one set of values, but get a slightly different set.

lib/cluster/etcd.go
```diff
@@ -882,3 +883,74 @@ func MakeEtcdOptsFromUriOpts(src libconnect.UriOpts) EtcdOpts {
...
+	cli, err := ConnectEtcdUriOpts(uriOpts, connOpts)
+	if err != nil {
+		return nil, err
+	}
+	return &etcdCSConnection{
+		cli: cli,
+	}, nil
+}
+
+func (c *etcdCSConnection) Close() error {
```
dmyger, line 902: Here and below, we need documentation comments for all public methods.

lib/cluster/tarantool.go
```diff
@@ -618,3 +618,109 @@ func ConnectTarantool(uriOpts libconnect.UriOpts,
...
+	conn, err := ConnectTarantool(uriOpts, connOpts)
+	if err != nil {
+		return nil, err
+	}
+	return &tarantoolCSConnection{
+		conn: conn,
+	}, nil
+}
+
+func (c *tarantoolCSConnection) Close() error {
```
dmyger, line 637: Here and below, we need documentation comments for all public methods.

test/conftest.py
```diff
@@ -243,24 +266,120 @@ def tt_instances(tt_path, request):
...
-    mark_app = request.node.get_closest_marker("tt")
+    mark_app = request.node.get_closest_marker("tt_app")
     return mark_app.kwargs.get("post_start")
 
 
 @pytest.fixture(scope="function")
-def tt(tt_cmd, tt_path, tt_instances, tt_running_targets, tt_post_start):
-    tt_ = Tt(tt_cmd, tt_path, tt_instances)
+def tt_app(tt, tt_path, tt_instances, tt_running_targets, tt_post_start):
+    tt_app_ = tt_helper.TtApp(tt, tt_path, tt_instances)
```
dmyger:
```suggestion
    app = tt_helper.TtApp(tt, tt_path, tt_instances)
```
`_` - the variable with this suffix is a bit confusing.

test/conftest.py
```diff
@@ -243,24 +266,120 @@ def tt_instances(tt_path, request):
...
+
+    # Launch failover coordinator.
+    coordinator = subprocess.Popen(["tarantool", "--failover", "--config", cluster.config_path])
+
+    cluster.config_storage = config_storage
+
+    yield cluster
+
+    # Kill failover coordinator.
+    coordinator.kill()
```
dmyger:
```suggestion
    request.addfinalizer(lambda: coordinator.kill())
    return cluster
```
As advice, it is better to use `addfinalizer(...)` — without a finaliser, the teardown cannot be performed if any exceptions or abnormal terminations occur — so this is only a possible place for flaky tests.

lib/cluster/tarantool.go
elhimov: It's one of the private implementations of CSConnection interface. Interface methods are commented at interface declaration. Are you sure that comments for interface implementation are required too? I don't see much sense in cloning interface methods comments.

test/conftest.py
elhimov: yield-approach is [recommended](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#yield-fixtures-recommended) way for fixture finalization.

cli/cluster/cmd/failover.go
elhimov: I think such a kind of "verbose" names is used for all commands contexts to avoid confusing with `ctx context.Context`, which is usually passed as first argument.

test/conftest.py
elhimov: This way you can parametrize only specific parameter(s). The rest ones will be default.

cli/cluster/cmd/failover.go
elhimov: Missed it, thanks.
Removed.

test/conftest.py
elhimov: Fixed.

test/conftest.py
elhimov: Usage examples:

Parametrize `num_replicas` only:
```python
@pytest.mark.parametrize("cluster_params",
    [
        pytest.param(dict(num_replicas=3), id="3"),
        pytest.param(dict(num_replicas=5), id="5"),
        pytest.param(dict(num_replicas=7), id="7"),
    ],
)
def test_aaa(tt, cluster):
    ...
```

Parametrize both `num_replicas` (default is 3) and `num_replicasets` (default is 1):
```python
@pytest.mark.parametrize("cluster_params",
    [
        pytest.param(dict(num_replicas=3), id="1x3"),
        pytest.param(dict(num_replicas=5, num_replicasets=2), id="2x5"),
        pytest.param(dict(num_replicasets=2), id="2x3"),
    ],
)
def test_bbb(tt, cluster):
    ...
```

elhimov: There is another way: instead of single compound fixture `cluster_params` use multiple simple fixtures, like `num_replicas`, `num_replicasets` etc. Each of them will have its default value. This way parametrization will look less verbose
```python
@pytest.mark.parametrize("num_replicasets, num_replicas",
    [
        pytest.param(1, 3, id="1x3"),
        pytest.param(2, 5, id="2x5"),
        pytest.param(2, 3, id="2x3"),
    ],
)
def test_bbb(tt, cluster):
    ...
```

but defaults cluster params will be spreaded out in the code. Except of that I have no pros/cons for any of the approaches (at least at the moment). What would you prefer? Or may be something else?

dmyger: Let's put this initialization in a separate function to make it explicit.
Something like following, just as concept:
```py
def MakeTcsParams(**kwargs):
     params = dict(
        path_to_cfg_dir=request.path.parent / "test_tcs_app",
        connection_test=True,
        connection_test_user="client",
        connection_test_password="secret",
        instance_name="instance-001",
        instance_host="localhost",
        instance_port="4401",
    )
    if kwargs is not None:
        params.update(kwargs)
    return params
```

And usage:
```py
@pytest.mark.parametrize("cluster_params", 
    [
        MakeTcsParams(num_replicas=3),
        MakeTcsParams(num_replicas=5, num_replicasets=2),
        MakeTcsParams(num_replicasets=2),
    ],
)
def test_bbb(tt, cluster):
    ...
```

dmyger: Or maybe so:
```py
def MakeTcsParams(id, **kwargs) -> pytest.param:
  # ...
  return pytest.param(params, id=id)
```

test/conftest.py
dmyger: You just need to be very careful during initialization, because if there is an error, finalization will not occur. [The documentation says this](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#handling-errors-for-yield-fixture), as I said:

> ### Handling errors for yield fixture
> If a yield fixture raises an exception before yielding, pytest won’t try to run the teardown code after that yield fixture’s yield statement.

elhimov: Still don't catch... I understand statement about errors handling, but if exception occurs before yield, `request.addfinalizer` in the suggested code will not be reached as well, so I don't see the advantage of `addfinalizer` approach.

elhimov: I think I got it finally )
Using `request.addfinalizer` is somewhat easier than yield-fixture because one can put `request.addfinalizer` at any place and thus it turns to a quite simple and safe approach -- just put `request.addfinalizer` right after resource acquisition to be sure corresponding cleanup code will be executed. Do you mean that?
I agree in general, but that means there is no place left for yield-fixture.

elhimov: Done.

lib/cluster/tarantool.go
dmyger: How can you tell that this method is an interface method and not just a method for a given structure?
And what kind of interface exactly is completely unclear.
Usually, they always write that this method implements the method of such-and-such an interface.

test/conftest.py
dmyger: > I think I got it finally )

Yes, that's right, that's exactly the problem with the yield approach.

lib/cluster/tarantool.go
elhimov: Reasonable. Added comments.

lib/cluster/etcd.go
elhimov: Done.

test/conftest.py
elhimov: Done.

dmyger (approved):

patapenka-alexey:
Due to the complex setup of the environment, it is difficult to create a reproducer. But as far as I can see, this command works.
[failover_output.json](https://github.com/user-attachments/files/22089668/failover_output.json)

```json
{"timestamp":"2025-09-02T09:44:28.699+0300","component":"failover","replicaset_name":"replicaset-001","message":"Found 1 RW instance.","rw_instance":"replicaset-001-c"}
```

patapenka-alexey (approved):

PR #1203, opened 2025-09-02
cartridge: fix discovery to avoid missing aliases

elhimov:
Implementation of cartridge bootstrap in `tt` needs to know aliases of all instances to work properly. Aliases of instances are managed with `membership` module which spreads out certain instance alias across the rest of instances using `gossip` protocol. At the very start of an instance only that instance "knows" its alias, others get this knowledge with some delay. That means in turn that the topology obtained from the different instances might differ in terms of aliases.

Prior to this patch topology from a single instance was used and if at that moment this instance haven't collected information about all the aliases, bootstrap failed. Now if any alias is missing we are trying to get it from the topologies obtained from the other instances until all aliases are resolved.

The patch fixes flaky test so there is no need to introduce new tests.

I didn't forget about:

Related issues:

_Closes [TNTP-3709](https://jira.vk.team/browse/TNTP-3709)_

### DIFF
--- a/cli/replicaset/cartridge.go
+++ b/cli/replicaset/cartridge.go
@@ -245,6 +245,7 @@ func (c *CartridgeApplication) discovery() (Replicasets, error) {
 			}
 			if topology.IsBootstrapped {
 				if newTopology.IsBootstrapped {
+					updateInstancesAliases(newTopology.Replicasets, topology.Replicasets)
 					topology = newTopology
 				}
 			} else {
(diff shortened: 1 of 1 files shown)


patapenka-alexey (approved):

.github/workflows/full-ci.yml
```diff
@@ -48,12 +48,16 @@ jobs:
 
       # This server starts and listen on 8084 port that is used for tests.
       - name: Stop Mono server
-        run: sudo systemctl kill mono-xsp4 || true
+        run: |
+          sudo netstat -tupln
```
oleg-jukovec:
```suggestion
        run: | ### Remove after https://jira.vk.team/browse/TNTP-3709 .
          sudo netstat -tupln
```

elhimov: Done.

patapenka-alexey (approved):

PR #1216, opened 2025-11-24
pack: corrupting relative symbolic links

oleg-jukovec:
Links were made relative only if they were located in an instance enabled directory inside temporary build path, but we need to do this for the entire temporary build path.

Closes #TNTP-5006

### DIFF
--- a/test/integration/pack/test_pack.py
+++ b/test/integration/pack/test_pack.py
@@ -2277,9 +2277,11 @@ def test_pack_app_local_tarantool(tt_cmd, tmpdir_with_tarantool, tmp_path):
         text=True,
     )
     tt_process.wait()
-    assert tt_process.returncode == 0
-
     build_output = tt_process.stdout.read()
+
+    print(build_output)
+
+    assert tt_process.returncode == 0
     assert "Bundle is packed successfully" in build_output
 
 


test/integration/pack/test_pack.py
```diff
@@ -2240,9 +2277,11 @@ def test_pack_app_local_tarantool(tt_cmd, tmpdir_with_tarantool, tmp_path):
         text=True,
     )
     tt_process.wait()
-    assert tt_process.returncode == 0
-
     build_output = tt_process.stdout.read()
+
+    print(build_output)
```
bigbes, line 2282: debug output?

oleg-jukovec: Yes and no.

With this output, ok tests won't print anything to the console. But if a fail occurs, there will be output that will make debugging easier.

cli/pack/tgz.go
```diff
@@ -44,22 +42,34 @@ func WriteTarArchive(srcDirPath string, compressWriter io.Writer,
...
-			// we should make it relative. Apriori it is known,
-			// that the source path of the link will be located
-			// in a directory higher.
+			// we should make it relative.
 			srcPath, _ := filepath.EvalSymlinks(filePath)
-			resolvedPath := filepath.Join("..", filepath.Base(srcPath))
-			tarHeader, err = tar.FileInfoHeader(fileInfo, resolvedPath)
-			if err != nil {
-				return err
+			if strings.Contains(srcPath, srcDirPath) {
```
elhimov, line 49: What if it is not true? Or it is known that link destination always under `srcDirPath`?

oleg-jukovec: We will fallthrough here:

https://github.com/tarantool/tt/blob/7503d0155967a80c4e5f399ff719217e37367321/cli/pack/tgz.go#L69-L74

PR #1232, opened 2026-01-11
ci: fix update workflow

patapenka-alexey:
This patch restores update workflow functionality.

Closes TNTP-4298

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -16,6 +16,10 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
     - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+        submodules: recursive
+        fetch-tags: true
 
     - name: Setup Go
       uses: actions/setup-go@v4
--- a/.gitlint-precommit.ini
+++ b/.gitlint-precommit.ini
@@ -6,6 +6,10 @@ regex-style-search = true
 regex = ^build\(deps\):.*dependabot\[bot\]$
 ignore = title-max-length,body-min-length
 
+[ignore-by-author-name]
+regex=^TarantoolBot(.*)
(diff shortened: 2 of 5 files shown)


.github/workflows/update.yml
```diff
@@ -48,7 +52,12 @@ jobs:
         author: 'TarantoolBot <TarantoolBot@users.noreply.github.com>'
         body: 'Automated changes by [update](https://github.com/tarantool/tt/blob/master/.github/workflows/update.yml) job.'
         branch: 'github-actions/update'
-        commit-message: 'generate: autoupdate'
+        commit-message: |
+          generate: update auto-generated code
+
+          This patch updates auto-generated code.
+
+          Related to #1231
```
oleg-jukovec: Please, make links to issues unnecessary in commit linter checks. It just complicate things.
```suggestion
```

oleg-jukovec (changes requested):

.github/workflows/update.yml
patapenka-alexey: Updated.

oleg-jukovec (approved):

.gitlint-precommit.ini
```diff
@@ -25,20 +25,3 @@ min-length=10
 
 [body-max-line-length]
 line-length = 72
-
-[body-match-regex:reference_task_id]
```
elhimov, line 29: I'd suggest to use [ignore rules](https://jorisroovers.com/gitlint/dev/ignoring_commits/) instead of eliminating this one.

It might be something like (just as example, not tested)
```
[ignore-by-author-name]
# Match commits by author name (e.g. ignore dependabot commits)
regex=^TarantoolBot(.*)
ignore=body-match-regex
```

PR #1236, opened 2026-01-29
ci: put static code check into a separate job

elhimov:
With this approach static check failures don't block unit/integration tests.

I didn't forget about:

Related issues:

Closes TNTP-6083


.pre-commit-config.yaml
```diff
@@ -47,7 +47,11 @@ repos:
         exclude: |
           (?x)^(
                 .*\.pb\.go|
-                .*/third_party/.*
+                .*/third_party/.*|
+                test/.*|
+                .*_test\.go|
```
patapenka-alexey: I think we should try to solve the problems in these files.

elhimov: You are right, it's better to fix them.

elhimov: Done.

cli/replicaset/cmd/demote.go
```diff
@@ -24,7 +24,7 @@ type DemoteCtx struct {
 	Conn connector.Connector
 	// RunningCtx is an application running context.
 	RunningCtx running.RunningCtx
-	// Force true if unfound instances can be skipped.
+	// Force true if unavailable instances can be skipped.
```
bigbes, line 27: Are you sure that "unfound" here means the same thing as "unavailable"? I haven't read this functionality in-depth, so feel free to resolve this comment if it's not applicable.

bigbes: I read patch and I think its fine :)

PR #1250, opened 2026-03-04
start: insufficient permissions to directories

### LINKED ISSUE #1238: tt does not show error message if tarantool fails on very start
`tt start` loses tarantool logs if tarantool has failed to start. For example, if
tarantool fails to start due permission issue.

Reproducer:

```
$ tarantool --version
Tarantool Enterprise 3.6.0-0-gb80f9ff9a
Target: Linux-x86_64-RelWithDebInfo
Build options: cmake . -DCMAKE_INSTALL_PREFIX=/builds/tarantool/delivery/sdk/build.sdk/tarantool-ee-3.6/static-build/tarantool-prefix -DENABLE_BACKTRACE=TRUE
Compiler: GNU-9.3.1
C_FLAGS: -fexceptions -funwind-tables -fasynchronous-unwind-tables -static-libstdc++ -fno-common -msse2 -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIC -fmacro-prefix-map=/builds/tarantool/delivery/sdk/tarantool-ee-3.6=. -std=c11 -Wall -Wextra -Wno-gnu-alignof-expression -fno-gnu89-inline -Wno-cast-function-type -O2 -g -DNDEBUG -ggdb -O2 -flto -fno-fat-lto-objects
CXX_FLAGS: -fexceptions -funwind-tables -fasynchronous-unwind-tables -static-libstdc++ -fno-common -msse2 -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIC -fmacro-prefix-map=/builds/tarantool/delivery/sdk/tarantool-ee-3.6=. -std=c++17 -Wall -Wextra -Wno-invalid-offsetof -Wno-gnu-alignof-expression -Wno-cast-function-type -O2 -g -DNDEBUG -ggdb -O2 -flto -fno-fat-lto-objects
$ tt version
Tarantool CLI EE 2.11.0, linux/amd64. commit: 0bacf70
$ tt create single_instance --name repro
   • Creating application in "/home/i-palagecha/code/palage4a/hub-playground/tarantool-playground/repro"
   • Using built-in 'single_instance' template.
   • Application 'repro' created successfully
$ ll repro
total 12K
-rw-r--r-- 1 i-palagecha i-palagecha 205 Feb  5 15:41 config.yml
-rw-r--r-- 1 i-palagecha i-palagecha  86 Feb  5 15:41 init.lua
-rw-r--r-- 1 i-palagecha i-palagecha  13 Feb  5 15:41 instances.yml
$ mkdir repro/var
$ id
uid=1000(i-palagecha) gid=1000(i-palagecha) groups=1000(i-palagecha),4(adm),20(dialout),24(cdrom),27(sudo),30(dip),46(plugdev),100(users),105(tss),114(lpadmin),984(docker)
$ sudo chown -R 1001:1001 repro/var
$ ll repro/var
total 0
$ cat tt.yaml
modules:
  # Directory where the external modules are stored.
  directory: modules

env:
  # Restart instance on failure.
  restart_on_failure: false

  # Directory that stores binary files.
  bin_dir: bin

  # Directory that stores Tarantool header files.
  inc_dir: include

  # Path to directory that stores all applications.
  # The directory can also contain symbolic links to applications.
  instances_enabled: instances.enabled

  # Tarantoolctl artifacts layout compatibility: if set to true tt will not create application
  # sub-directories for control socket, pid files, log files, etc.. Data files (wal, vinyl,
  # snap) and multi-instance applications are not affected by this option.
  tarantoolctl_layout: false

app:
  # Directory that stores various instance runtime
  # artifacts like console socket, PID file, etc.
  run_dir: var/run

  # Directory that stores log files.
  log_dir: var/log

  # Directory where write-ahead log (.xlog) files are stored.
  wal_dir: var/lib

  # Directory where memtx stores snapshot (.snap) files.
  memtx_dir: var/lib

  # Directory where vinyl files or subdirectories will be stored.
  vinyl_dir: var/lib

# Path to file with credentials for downloading Tarantool Enterprise Edition.
# credential_path: /path/to/file
ee:
  credential_path:

templates:
  # The path to templates search directory.
  - path: templates

repo:
  # Directory where local rocks files could be found.
  rocks:
  # Directory that stores installation files.
  distfiles: distfiles
$ tt start repro
   • Starting an instance [repro:instance001]...
$ tt status repro
 INSTANCE           STATUS       PID  MODE  CONFIG  BOX  UPSTREAM
 repro:instance001  NOT RUNNING
$ tt log repro
$ tt start -i repro
$ tt log repro
$ cd repro
$ tarantool --name instance001 --config config.yml init.lua
builtin/config/applier/mkdir.lua:26: mkdir.apply[wal.dir]: failed to create directory var/lib/instance001: Error creating directory /home/i-palagecha/code/palage4a/hub-playground/tarantool-playground/repro/var/lib: fio: Permission denied
```

I am expecting that `tt log repro` and `tt start -i repro` will have a log message about error.

Mockird31:
Fixed a bug where an error did not appear when access rights to the var directive were insufficient.

Closes #1238

### DIFF
--- a/cli/running/running.go
+++ b/cli/running/running.go
@@ -26,6 +26,7 @@ import (
 	"github.com/tarantool/tt/cli/util/regexputil"
 	libcluster "github.com/tarantool/tt/lib/cluster"
 	"github.com/tarantool/tt/lib/integrity"
+	"golang.org/x/sys/unix"
 )
 
 const defaultDirPerms = 0o770
@@ -789,14 +790,18 @@ func RunInstance(ctx context.Context, cmdCtx *cmdcontext.CmdCtx, inst InstanceCt
 
 // Start an Instance.
 func Start(cmdCtx *cmdcontext.CmdCtx, inst *InstanceCtx) error {
-	if err := createInstanceDataDirectories(*inst); err != nil {
-		return fmt.Errorf("failed to create a directory: %s", err)
-	}
 	logger, err := createLogger(inst)
 	if err != nil {
 		return fmt.Errorf("cannot create a logger: %s", err)
 	}
 	logger.Println("[INFO] Start") // Create a log file before any other actions.
+
+	if err := createInstanceDataDirectories(*inst); err != nil {
+		logger.Fatalf("[ERROR] Failed to create data directories for instance %q: %s",
+			inst.InstName,
+			err)
+		return fmt.Errorf("failed to create a directory: %s", err)
+	}
 
 	provider := providerImpl{cmdCtx: cmdCtx, instanceCtx: inst}
 	preStartAction := func() error {
@@ -966,6 +971,17 @@ Cluster config path: %q`, tntVersion.Str, inst.ClusterConfigPath)
 func StartWatchdog(cmdCtx *cmdcontext.CmdCtx, ttExecutable string, instance InstanceCtx,
 	args []string,
 ) error {
+	err := util.CreateDirectory(instance.LogDir, defaultDirPerms)
+	if err != nil {
+		return fmt.Errorf("failed to create log directory for instance %q: %s",
+			GetAppInstanceName(instance), err)
+	}
+
+	if err := syscall.Access(instance.LogDir, unix.W_OK); err != nil {
+		return fmt.Errorf("instance %q has non-writable log directory %q: %s",
+			instance.InstName, instance.RunDir, err)
+	}
+
 	appName := GetAppInstanceName(instance)
 	// If an instance is already running don't try to start it again.
 	// To restart an instance use tt restart command.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- `tt start`: fixed a bug where an error did not appear when access rights
+  to the var directive were insufficient.
+
 ## [2.12.0] - 2026-03-30
 
 This maintenance release marks the end of active development on the v2
--- a/cli/cmd/start.go
+++ b/cli/cmd/start.go
@@ -10,6 +10,7 @@ import (
 	"sync"
 	"syscall"
 
+	"github.com/apex/log"
 	"github.com/spf13/cobra"
 	"github.com/tarantool/tt/cli/cmd/internal"
 	"github.com/tarantool/tt/cli/cmdcontext"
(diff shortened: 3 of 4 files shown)


cli/running/running.go
```diff
@@ -966,6 +966,16 @@ Cluster config path: %q`, tntVersion.Str, inst.ClusterConfigPath)
...
 ) error {
+	for _, dataDir := range [...]string{
+		instance.WalDir, instance.VinylDir,
+		instance.MemtxDir, instance.RunDir,
+	} {
+		if err := util.CreateDirectory(dataDir, defaultDirPerms); err != nil {
+			return fmt.Errorf("failed to run instance %q: %s",
+				GetAppInstanceName(instance)+" ", err)
+		}
+	}
```
oleg-jukovec: Will this check work if the directories are already created, but the user does not have permission to change them, only to create files inside?

What will happen if a user unable to create files inside the directory?

We need tests for the cases.

oleg-jukovec (dismissed):

cli/running/running.go
Mockird31: Done.

cli/running/running.go
```diff
@@ -753,7 +753,12 @@ func RunInstance(ctx context.Context, cmdCtx *cmdcontext.CmdCtx, inst InstanceCt
 ) error {
 	for _, dataDir := range [...]string{inst.WalDir, inst.VinylDir, inst.MemtxDir, inst.RunDir} {
 		if err := util.CreateDirectory(dataDir, defaultDirPerms); err != nil {
-			return err
+			return fmt.Errorf("failed to run instance %q: %s",
+				GetAppInstanceName(inst)+" ", err)
```
bigbes:
```suggestion
			return fmt.Errorf("failed to run instance %q : %w",
				GetAppInstanceName(inst), err)
```

bigbes: Are you sure space is needed here, at all?

Mockird31: Done

bigbes (approved):

bigbes (approved):

cli/util/util.go
```diff
@@ -752,6 +767,33 @@ func CreateDirectory(dirName string, fileMode os.FileMode) error {
 	return nil
 }
 
+// IsWritable checks if all files and subdirectories inside the directory
+// are writable by the current user.
+func IsWritable(dirName string) ([]PermissionError, error) {
```
elhimov: Well strictly speaking we don't need _all_ files to be writable.
It seems the only file we should care about is log file. If we had created it successfully the rest of the problems (if any) will be logged in a regular way.

elhimov (changes requested):

cli/util/util.go
Mockird31: Done

elhimov (approved):

elhimov:
There is still a gap between checking a log directory and creation of a log file, so technically it's possible that creation of the log file is failed even if the log directory checking was succeed. It seems the only way to make sure log file is created is... create log file. But it is created in a child process (which becomes watchdog of the corresponding `tarantool` instance) rather than initial `tt` process and it doesn't look trivial to catch it over the processes boundary, so for now let's keep it as suggested in this patch.

PR #1252, opened 2026-03-08
running: fix watchdog tests with race detector

sssciel:
Watchdog tests completed with datarace warning with go test race detector enabled. This was because there was unsafe access to the proccess_controller.

Closes TNTP-6356

### DIFF
--- a/cli/running/script_instance.go
+++ b/cli/running/script_instance.go
@@ -21,7 +21,7 @@ const (
 
 // scriptInstance represents a tarantool invoked with an instance script provided.
 type scriptInstance struct {
-	baseInstance
+	*baseInstance
 }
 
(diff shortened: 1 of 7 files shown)


cli/running/script_instance.go
```diff
@@ -170,9 +170,12 @@ func (inst *scriptInstance) Start(ctx context.Context) error {
 	inst.setTarantoolLog(cmd)
 
 	// Start an Instance.
+	inst.mu.Lock()
```
psergee, line 173: The current approach raises questions. We are protecting only a single member of `inst`, even though the cluster instance employs the exact same process controller initialization code without any locking. Is this a test-specific fix for race conditions, rather than a comprehensive thread‑safety mechanism for the instances API?

PR #1271, opened 2026-04-01
cartridge: remove from build, pack

patapenka-alexey:
This patch removes cartridge support from `build` and `pack` commands.

Closes #TNTP-6976

### DIFF
--- a/cli/pack/common.go
+++ b/cli/pack/common.go
@@ -199,9 +199,9 @@ func getAppNamesToPack(packCtx *PackCtx, cliOpts *config.CliOpts) []string {
 
 // updateEnvPath sets base path for the tt environment in temporary package directory.
 // By default it is a base directory passed as an argument. Or an application name sub-dir
-// in case of cartridge compat or single application environment.
+// in case of single application environment.
 func updateEnvPath(basePath string, packCtx *PackCtx, cliOpts *config.CliOpts) (string, error) {
-	if cliOpts.Env.InstancesEnabled == "." || packCtx.CartridgeCompat {
+	if cliOpts.Env.InstancesEnabled == "." {
 		basePath = util.JoinPaths(basePath, packCtx.Name)
 		if err := os.MkdirAll(basePath, dirPermissions); err != nil {
 			return basePath, fmt.Errorf("cannot create bundle directory %q: %s", basePath, err)
@@ -212,7 +212,7 @@ func updateEnvPath(basePath string, packCtx *PackCtx, cliOpts *config.CliOpts) (
 
 // copyEnvModules copies tt modules.
 func copyEnvModules(bundleEnvPath string, packCtx *PackCtx, cliOpts, newOpts *config.CliOpts) {
-	if packCtx.WithoutModules || packCtx.CartridgeCompat || cliOpts.Modules == nil ||
+	if packCtx.WithoutModules || cliOpts.Modules == nil ||
 		len(cliOpts.Modules.Directories) == 0 {
 
 		return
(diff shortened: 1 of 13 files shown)


bigbes (approved):

cli/pack/common.go
```diff
@@ -254,10 +254,9 @@ func copyBinaries(bundleEnvPath string, packCtx *PackCtx, cmdCtx *cmdcontext.Cmd
 	}
 
 	pkgBin := bundleEnvPath
-	if !packCtx.CartridgeCompat {
-		// In cartridge compat mode copy binaries directly to the env dir.
-		pkgBin = util.JoinPaths(bundleEnvPath, newOpts.Env.BinDir)
-	}
+
+	// Copy binaries directly to the env dir.
+	pkgBin = util.JoinPaths(bundleEnvPath, newOpts.Env.BinDir)
```
elhimov:
```suggestion
	pkgBin := util.JoinPaths(bundleEnvPath, newOpts.Env.BinDir)
```

Note that comment should be dropped. The original comment acted a bit unusual as it described the behavior "outside" of "if".

cli/pack/common.go
```diff
@@ -418,18 +417,6 @@ func prepareBundle(cmdCtx *cmdcontext.CmdCtx, packCtx *PackCtx,
...
-		if err := generateVersionFile(bundleEnvPath, cmdCtx, packCtx); err != nil {
-			log.Warnf("Failed to generate VERSION file: %s", err)
-		}
-
-		// Generate VERSION.lua file.
-		if err := generateVersionLuaFile(bundleEnvPath, packCtx); err != nil {
-			log.Warnf("Failed to generate VERSION.lua file: %s", err)
-		}
-	}
-
```
elhimov, line 432: I doubt that we need to delete this code completely. May be we should generate them always?
@oleg-jukovec

cli/pack/common.go
```diff
@@ -670,26 +652,17 @@ func buildAppRocks(cmdCtx *cmdcontext.CmdCtx, packCtx *PackCtx,
...
 	packageVersion := defaultVersion
 	if packCtx.Version == "" {
-		// Get version from git only if packing an application from the current directory,
-		// or packing with cartridge-compat enabled.
+		// Get version from git only if packing an application from the current directory.
 		appPath := opts.Env.InstancesEnabled
-		if opts.Env.InstancesEnabled != "." && packCtx.CartridgeCompat {
+		if opts.Env.InstancesEnabled != "." {
 			appPath = filepath.Join(appPath, packCtx.Name)
 		}
```
elhimov:
```suggestion
```

this section doesn't make sense anymore

cli/pack/common.go
```diff
@@ -670,26 +652,17 @@ func buildAppRocks(cmdCtx *cmdcontext.CmdCtx, packCtx *PackCtx,
 func getVersion(packCtx *PackCtx, opts *config.CliOpts, defaultVersion string) string {
 	packageVersion := defaultVersion
```
elhimov: To improve readability I'd suggest to simplify implementation of `getVersion` as follows:
```go
func getVersion(packCtx *PackCtx, opts *config.CliOpts, defaultVersion string) string {
	if packCtx.Version != "" {
        return packCtx.Version
    }

	// Try to get version from git only if packing an application from the current directory.
	if opts.Env.InstancesEnabled == "." {
		version, err := util.CheckVersionFromGit(opts.Env.InstancesEnabled)
		if err == nil {
			return version
		}
	}

	return defaultVersion
}
```

elhimov (changes requested):

cli/pack/common.go
bigbes: This part will be remade later when new `tt pack` flow will be implemented, but I get your idea.

elhimov: Want to clarify just in case. Got it, thank you!

elhimov (approved):

PR #1273, opened 2026-04-03
cartridge: remove vendored cartridge-cli and all references

bigbes:
Remove cli/cartridge/ directory (vendored code + 9 patches), cartridge-cli from go.mod, cartridge backward-compat code from running/, dead test fixtures, and documentation sections.

Closes TNTP-6979


patapenka-alexey (approved):

cli/pack/archive.go
```diff
@@ -62,7 +62,7 @@ func (packer *archivePacker) Run(cmdCtx *cmdcontext.CmdCtx, packCtx *PackCtx,
 	return nil
 }
 
-// generateVersionLuaFile generates VERSION.lua file (for cartridge-cli compatibility).
+// generateVersionLuaFile generates VERSION.lua file.
```
elhimov: If I understand correctly this function (and the below one) is not used anymore. Is it kept for future use or just in case?

elhimov (approved):

cli/pack/archive.go
bigbes: fixed, ty!

PR #1282, opened 2026-04-29
lib/cluster: remove file/env collectors superseded by go-config

bigbes:
go-config ships a Tarantool builder with schema-aware env var resolution and an embedded JSON Schema, so the hand-maintained file/env collectors in lib/cluster are no longer needed by cli/cluster.

Closes TNTP-7389


sssciel (approved):

sssciel (changes requested):
CI is red

sssciel (approved):

patapenka-alexey (approved):

PR #1305, opened 2026-06-18
cli/manifest: add manifest and lock data model

bigbes:
Introduce cli/manifest, the data layer for tt packages: Go types for the manifest (app.manifest.toml) and lock (app.manifest.lock), TOML read/write, structural validation, format-version handling and manifest_hash. No git, no dependency resolution, no .rocks/ - only the standard library and the TOML library.

* Manifest/Package/Platform with the platform constraint parsed into semver part and [ce]/[ee] flavor (Constraint via TextMarshaler).
* Dependency decodes both the short (bare constraint string) and long (table) forms from one map via go-toml's unstable.Unmarshaler.
* Lock serializes to the [lock.products.<name>] table tree through an internal wire struct (go-toml does not split dotted struct tags).
* Validate() does structural checks; format-version rules split unknown fields (warn on newer minor, fail otherwise, refuse newer major) from unknown enum values (always fail). manifest_hash is SHA-256 of the raw file bytes.

Closes TNTP-8220

### DIFF
--- a/cli/manifest/constraint.go
+++ b/cli/manifest/constraint.go
@@ -1,0 +1,98 @@
+package manifest
+
+import "strings"
+
+// Constraint is a platform version requirement parsed into its semver part and
+// an optional flavor. It is stored as "<semver-constraint>[<flavor>]" in TOML,
+// where <flavor> is [ce] or [ee].
+//
+// The semver constraint itself is kept verbatim - validating the range is the
+// resolver's job, not this layer's. Only the flavor suffix is parsed here.
+//
+//nolint:recvcheck // TextMarshaler needs a value receiver, UnmarshalText a pointer.
+type Constraint struct {
+	// Version is the semver constraint part, e.g. ">=3.0.0,<4.0.0". Empty for
+	// an unset (omitted) constraint.
+	Version string
+	// Flavor is "ce", "ee" or "" (unspecified). For tarantool/tt an empty
+	// flavor means [ce]; see EffectiveFlavor.
+	Flavor string
+}
+
+// IsZero reports whether the constraint is unset.
+func (c Constraint) IsZero() bool {
+	return c.Version == "" && c.Flavor == ""
+}
+
+// EffectiveFlavor returns the flavor with the [ce] default applied to an
+// unspecified flavor.
+func (c Constraint) EffectiveFlavor() string {
+	if c.Flavor == "" {
+		return "ce"
+	}
+
+	return c.Flavor
+}
+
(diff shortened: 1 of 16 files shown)


cli/manifest/constraint.go
```diff
@@ -0,0 +1,80 @@
...
+	case "ce", "ee":
+	default:
+		return fmt.Errorf("invalid version constraint %q: unknown flavor %q (want [ce] or [ee])", s, flavor)
+	}
+	if version == "" {
+		return fmt.Errorf("invalid version constraint %q: missing version before flavor", s)
+	}
+	c.Version = version
+	c.Flavor = flavor
+	return nil
```
patapenka-alexey: Please, add a newlines for better readability.

cli/manifest/dependency.go
```diff
@@ -0,0 +1,71 @@
...
+	Version  string `toml:"version,omitempty"`  // constraint; required for registry
+	Path     string `toml:"path,omitempty"`     // required for path
+	Registry string `toml:"registry,omitempty"` // overrides the server
+	Kind     string `toml:"kind,omitempty"`     // in v0 only "library"
+}
+
+// UnmarshalTOML implements the go-toml unstable.Unmarshaler interface so a
+// dependency can decode from either a bare string or a table. The decoder must
+// have EnableUnmarshalerInterface set (ParseManifest/ParseLock do).
+func (d *Dependency) UnmarshalTOML(node *unstable.Node) error {
```
patapenka-alexey, line 29: And here.

cli/manifest/manifest.go
```diff
@@ -0,0 +1,173 @@
...
+	raw []byte // raw file bytes, source for manifest_hash
+}
+
+// Package holds the package identity and metadata ([package]).
+type Package struct {
+	Name               string   `toml:"name"` // regex [a-z][a-z0-9-]*; not "bin", not "manifests"
+	Description        string   `toml:"description,omitempty"`
+	License            string   `toml:"license,omitempty"`
+	LicenseFiles       []string `toml:"license_files,omitempty"`
+	Include            []string `toml:"include,omitempty"`
```
sssciel, line 49: For what is this field?

sssciel (approved):

cli/manifest/manifest.go
bigbes: It's for an additional files to be passed with result artefact, like README or whatsnot expected by the user.

patapenka-alexey (approved):

PR #1308, opened 2026-06-23
backup: add storage interface and fs/s3 impl

sssciel:
Implement local filesystem and S3-compatible backends.

Closes TNTP-8210


patapenka-alexey:
Please, fix static-code-check.

patapenka-alexey (approved):

lib/backup/storage/s3/s3.go
```diff
@@ -0,0 +1,206 @@
...
+	objectName := s.objectName(key)
+
+	err = s.client.RemoveObject(ctx, s.bucket, objectName, minio.RemoveObjectOptions{})
+	if err != nil {
+		if isNotFound(err) {
+			return nil
+		}
+
+		return fmt.Errorf("failed to delete s3 object %q: %w", key, err)
+	}
```
bigbes, line 200: NoSuchBucket is 'isNotFound' error, but it shouldn't be. Absence of bucket is an invalid state of our client. Pls rethink this part of PR.

lib/backup/storage/s3/s3.go
```diff
@@ -0,0 +1,206 @@
...
+		return errAccessKeyIDRequired
+	case strings.TrimSpace(cfg.SecretAccessKey) == "":
+		return errSecretAccessKeyRequired
+	default:
+		return nil
+	}
+}
+
+// isNotFound reports whether err is an S3 missing-key or missing-bucket error.
+func isNotFound(err error) bool {
```
bigbes: If client returns 404 - this work won't do a thing, I think. But it's possible that server returns 404 instead of standard S3 error.

lib/backup/storage/fs/fs.go
```diff
@@ -0,0 +1,284 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve storage root %q: %w", cfg.Path, err)
+	}
+
+	return &Storage{root: root}, nil
+}
+
+// List returns objects whose keys start with the given prefix, sorted by key.
+func (s *Storage) List(ctx context.Context, prefix string) ([]storage.ObjectInfo, error) {
+	prefix, err := storage.CleanPrefix(prefix)
```
bigbes: You're rewriting prefix here, so when you print it later in case of an error - it'll be empty.

lib/backup/storage/fs/fs.go
```diff
@@ -0,0 +1,284 @@
...
+	return n, nil
+}
+
+// isTempFile reports whether name is a temporary file created during Put.
+func isTempFile(name string) bool {
+	return strings.HasPrefix(name, tempFilePrefix)
+}
+
+// scanDir returns the deepest directory that can be walked for the given prefix.
+func scanDir(prefix string) string {
```
bigbes: I don't understand this function. Why not simply 'https://pkg.go.dev/path#Dir'?

lib/backup/storage/fs/fs.go
```diff
@@ -0,0 +1,284 @@
...
+
+// Put stores an object to the filesystem. The size parameter is unused;
+// it exists to satisfy the storage.Storage interface.
+func (s *Storage) Put(ctx context.Context, key string, r io.Reader, _ int64) error {
+	path, err := s.objectPath(key)
+	if err != nil {
+		return fmt.Errorf("failed to resolve object path %q: %w", key, err)
+	}
+
+	if err := ctx.Err(); err != nil {
```
bigbes: Are you sure this is needed? If it's needed - add it to other places too.

This check should be the first check in the function, btw.

bigbes (changes requested):

lib/backup/storage/s3/s3.go
sssciel: Yes, thanks, it was changed.

lib/backup/storage/s3/s3.go
sssciel: Done

lib/backup/storage/fs/fs.go
sssciel: Yes, validation was added after, so i missed this, thanks. Done.

lib/backup/storage/fs/fs.go
sssciel: I didn't know about this func. Thanks, changed.

lib/backup/storage/fs/fs.go
sssciel: Thanks. Done.

bigbes (approved):

PR #1310, opened 2026-06-29
cli/manifest: derive package version and generate version.lua

bigbes:
Add cli/manifest/version, the source of truth for a package's version so it is never written by hand. Derive resolves, in order, a VERSION file next to the manifest, the most recent v* tag (git describe), then a tagless 0.0.0-dev.<N>+g<sha> fallback. It strips a leading v, keeps only SemVer 2.0.0 tags, passes pre-releases through, opens a -dev.N cycle after a release tag (bumping the patch) or a .dev.N cycle after a pre-release tag (patch frozen), and appends .dirty build metadata for uncommitted changes - excluding the manifest lock so a regenerated lock alone does not stale a clean release.

GenerateVersionLua renders the version.lua descriptor (version, commit, dirty, flavor, built_at); the build decides whether to emit it via package.generate_version_lua and where to drop it. Versions are parsed and compared with go-version so pre-release ordering matches the resolver.

The new code sits under the strict cli/manifest golangci ruleset; go-version is added to its depguard allow-list as the package's one dependency.

Closes TNTP-8229

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -90,11 +90,15 @@ linters:
           allow:
             - $gostd
             - "github.com/pelletier/go-toml/v2"
+            # cli/manifest/version derives SemVer with go-version so its
+            # pre-release ordering matches the dependency resolver.
+            - "github.com/hashicorp/go-version"
         test:
           files:
             - "$test"
           allow:
             - $gostd
             - "github.com/pelletier/go-toml/v2"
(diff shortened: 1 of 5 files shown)


bigbes:
Superseded: re-opening on bigbes/tntp-8229-package-version with the TNTP-8229 ticket reference.

PR #1321, opened 2026-07-15
backup: add manifest chain impl

sssciel:
Build cluster recovery points from manifest segments and resolve a
target time to the nearest point not later than it. Report topology
boundaries and chain breaks when the time falls into a gap, and mark
orphan, fork, and vclock-mismatch problems per shard.

Closes TNTP-8391

### DIFF
--- a/lib/backup/chain/build.go
+++ b/lib/backup/chain/build.go
@@ -1,0 +1,383 @@
+package chain
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"maps"
+	"slices"
+	"strings"
+
+	"github.com/tarantool/tt/lib/backup"
+	"github.com/tarantool/tt/lib/backup/storage"
+)
+
+// Load reads every stored manifest and builds a fully marked chain.
+func Load(ctx context.Context, store storage.Storage) (*Chain, error) {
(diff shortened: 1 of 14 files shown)


lib/backup/chain/build.go
```diff
@@ -0,0 +1,346 @@
...
+}
+
+// propagateProblem attaches a problem to its source and creates an inherited
+// copy for each descendant. All writes affect the Entries stored in groups.
+func propagateProblem(
+	source *Entry,
+	problem *Problem,
+	children map[backup.BackupID][]*Entry,
+) {
+	source.Problems = append(source.Problems, problem)
```
bigbes, line 205: why is it done here?

lib/backup/chain/points.go
patapenka-alexey, line 1: The coding style we have adopted involves separating logical blocks of code with empty lines, without which it is very difficult to read.

sssciel: Thanks. Code was separated.

lib/backup/chain/build.go
sssciel: Because of duplication.

bigbes (approved):

patapenka-alexey (approved):

patapenka-alexey (approved):

PR #1331, opened 2026-07-22
cluster: add cluster topology command

sssciel:
The tt cluster topology command has been added. It shows the current cluster topology based on the cluster manifest (Tarantool 3.0) with a bypass of all instances for the current status. The credits are also taken from the cluster manifest.

Part of TNTP-8609

### DIFF
--- a/cli/cmd/topology.go
+++ b/cli/cmd/topology.go
@@ -1,0 +1,551 @@
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+	"path/filepath"
+	"sort"
+	"strings"
+	"sync"
+
+	"github.com/apex/log"
+
+	clustercli "github.com/tarantool/tt/cli/cluster"
+	"github.com/tarantool/tt/cli/cmdcontext"
+	"github.com/tarantool/tt/cli/connect"
+	"github.com/tarantool/tt/cli/connector"
+	"github.com/tarantool/tt/cli/replicaset"
+	replicasetcmd "github.com/tarantool/tt/cli/replicaset/cmd"
+	libcluster "github.com/tarantool/tt/lib/cluster"
+	libconnect "github.com/tarantool/tt/lib/connect"
+)
+
+var (
+	topologyFormat           string
+	topologyConfigPath       string
+	topologyUnixConnectMutex sync.Mutex
+)
+
+const (
+	topologyStatusOK           = "OK"
+	topologyStatusNotReachable = "not reachable"
+	topologyFormatJSON         = "json"
+	topologyFormatTable        = "table"
+)
+
+type topologyOutput struct {
+	Replicasets map[string][]topologyInstanceOutput `json:"replicasets"`
+}
+
+type topologyInstanceOutput struct {
+	InstanceUUID string `json:"instance_uuid"`
+	InstanceName string `json:"instance_name"`
+	Hostname     string `json:"hostname"`
+	Mode         string `json:"mode"`
+	Status       string `json:"status"`
+}
+
+// hostnameExpr fetches the instance UUID and the hostname of the node.
+const hostnameExpr = `return box.info.uuid, box.info.hostname`
+
+type topologyDiscoveryResult struct {
+	topology     *replicaset.Replicasets
+	instanceUUID string
+	hostname     string
+	connected    bool
+}
+
+func discoverInstanceTopology(
+	clusterConfig libcluster.ClusterConfig,
+	instName string,
+	configDir string,
+	connectCtx connect.ConnectCtx,
+) topologyDiscoveryResult {
+	instConfig := libcluster.Instantiate(clusterConfig, instName)
+	advertiseData, _ := instConfig.Get([]string{"iproto", "advertise", "client"})
+	uri, _ := advertiseData.(string)
+	if uri == "" {
+		listenData, _ := instConfig.Get([]string{"iproto", "listen"})
+		if listen, ok := listenData.([]any); ok && len(listen) > 0 {
+			endpoint, _ := listen[0].(map[any]any)
+			uri, _ = endpoint["uri"].(string)
+		}
+	}
+	if uri == "" {
+		log.Warnf("instance %q: no client URI found, skipping", instName)
+		return topologyDiscoveryResult{}
+	}
+
+	groupName, rsName, _ := libcluster.FindInstance(clusterConfig, instName)
+	uri = renderConfigTemplate(uri, instName, rsName, groupName)
+
+	network, address := parseListenURI(uri, configDir)
+	connOpts := makeConnOpts(network, address, connectCtx)
+	conn, err := connectTopologyInstance(connOpts)
+	if err != nil {
+		log.Warnf("instance %q: failed to connect: %s", instName, err)
+		return topologyDiscoveryResult{}
+	}
+	defer conn.Close()
+
+	result := topologyDiscoveryResult{connected: true}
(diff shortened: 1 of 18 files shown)


lib/connect/uri_test.go
```diff
@@ -219,6 +219,9 @@ func TestParseBaseURI(t *testing.T) {
 		{"/path/to/socket", connect.UnixNetwork, "/path/to/socket"},
 		{"unix:///path/to/socket", connect.UnixNetwork, "/path/to/socket"},
 		{"unix://..//path/to/socket", connect.UnixNetwork, "..//path/to/socket"},
+		{"unix/:./instance-001.iproto", connect.UnixNetwork, "./instance-001.iproto"},
+		{"unix/:/var/run/tt.sock", connect.UnixNetwork, "/var/run/tt.sock"},
+		{"unix/:instance.sock", connect.UnixNetwork, "instance.sock"},
 		{"..//path", connect.UnixNetwork, "..//path"},
 		{"some_uri", connect.TCPNetwork, "some_uri"}, // Keeps unchanged
```
patapenka-alexey:
```suggestion
		{"some_uri", connect.TCPNetwork, "some_uri"}, // Keeps unchanged.
```

patapenka-alexey:
```sh
   • Active cluster topology
 
   • router-001 (5c23172c-e6e4-42d3-8adb-3decd0ee01b4)
   •     M router-001-a f71c4a57-a06d-4aea-b3b0-dfdd916638e2  ap  rw
   • storage-001 (e79037e9-904b-4084-9765-eff0139c2da6)
   •     M storage-001-a c848a298-c3ef-48e7-a1ad-3349f866f4ef  ap  rw
   •       storage-001-b 230e497f-f2de-47f6-9d84-9cd2b2a9375c  ap  read
```
I propose to change "read" -> "ro".

patapenka-alexey (changes requested):

lib/connect/uri_test.go
sssciel: Done

patapenka-alexey (approved):

cli/cmd/topology.go
```diff
@@ -0,0 +1,472 @@
...
+		}
+
+		password, _ := userMap["password"].(string)
+		if password == "" {
+			continue
+		}
+
+		nameStr, _ := name.(string)
+
+		// Prefer a user with the "super" role.
```
bigbes: but why we should prefer 'super' role instead of given credentials?

cli/cmd/topology.go
```diff
@@ -0,0 +1,472 @@
...
+
+// internalClusterTopologyModule is an entrypoint for `cluster topology` command.
+func internalClusterTopologyModule(cmdCtx *cmdcontext.CmdCtx, args []string) error {
+	switch topologyFormat {
+	case "json", "table", "":
+	default:
+		return fmt.Errorf("unsupported format: %s (use table or json)", topologyFormat)
+	}
+
+	// Read and parse the cluster config.
```
bigbes: cluster config may be in etcd/tcs too.

cli/cmd/topology.go
```diff
@@ -0,0 +1,472 @@
...
+}
+
+func discoverInstanceTopology(
+	clusterConfig libcluster.ClusterConfig,
+	instName string,
+	configDir string,
+	hostnames map[string]string,
+) (*replicaset.Replicasets, bool) {
+	instConfig := libcluster.Instantiate(clusterConfig, instName)
+	listenData, err := instConfig.Get([]string{"iproto", "listen"})
```
bigbes: we should use `{"iproto", "advertise", "client"}` here

cli/cmd/topology.go
```diff
@@ -0,0 +1,472 @@
...
+			name := inst.Alias
+			if name == "" {
+				name = inst.UUID
+			}
+			hostname := lookupHostname(inst.UUID, hostnames)
+			mode := inst.Mode.String()
+			if mode == "read" {
+				mode = "ro"
+			}
+			log.Infof("    %s %s  %s  %s",
```
bigbes: Im not completely sure we shoudl use log for such operations, but since it's human output - maybe it's fine.

cli/cmd/topology.go
```diff
@@ -0,0 +1,472 @@
...
+		return "", ""
+	}
+
+	users, ok := usersData.(map[any]any)
+	if !ok {
+		return "", ""
+	}
+
+	var firstName, firstPassword string
+	for name, userData := range users {
```
bigbes: Users is a map - so this part of code is non-determenistic. Needs some rethough.

cli/cmd/topology.go
```diff
@@ -0,0 +1,472 @@
...
+			if !found {
+				merged.Replicasets = append(merged.Replicasets, r)
+			}
+		}
+	}
+
+	return merged
+}
+
+// filterUnreachable removes instances that could not be connected to
```
bigbes: We should not delete it - this part is a part of cluster info. Even if replica isn't reachable - we should honour topology afaiu.

cli/cmd/topology.go
```diff
@@ -0,0 +1,472 @@
...
+		SslKeyFile:  replicasetSslKeyFile,
+		SslCertFile: replicasetSslCertFile,
+		SslCaFile:   replicasetSslCaFile,
+		SslCiphers:  replicasetSslCiphers,
+	}
+
+	// Build connection options and connect.
+	network, address := parseListenURI(uri, configDir)
+	connOpts := makeConnOpts(network, address, connectCtx)
+	conn, err := connector.Connect(connOpts)
```
bigbes: Will it block?

bigbes: If it will - we should, at least, make connection to be parallel.

cli/cmd/topology.go
```diff
@@ -0,0 +1,472 @@
...
+
+var (
+	topologyFormat     string
+	topologyConfigPath string
+)
+
+// hostnameExpr fetches the instance UUID and the hostname of the node.
+const hostnameExpr = `return box.info.uuid, box.info.hostname`
+
+func newDiscoverer(
```
bigbes: looks like duplicate of `replicasetcmd.makeInstanceOrchestrator`

bigbes:
pls add changelog record

cli/cmd/topology.go
sssciel: Done, now we use -p -u flags for connection.

cli/cmd/topology.go
sssciel: CS support added via URI and envs like in other cluster commands.

cli/cmd/topology.go
sssciel: Fixed.

cli/cmd/topology.go
sssciel: We print in logs only table output. JSON prints in stdout via fmt.

cli/cmd/topology.go
sssciel: Fixed in previous comment.

cli/cmd/topology.go
sssciel: Now disabled replicas have status "not reachable". Others - "OK".

cli/cmd/topology.go
sssciel: Now topology has parallel connections.

cli/cmd/topology.go
sssciel: makeInstanceOrchestrator was renamed to MakeInstanceOrchestrator,

bigbes (approved):

PR #1335, opened 2026-07-29
tests: fix vclock backup tests

sssciel:
The tests in cli/backup were based on the RFC, which differs from the final implementation in tarantool. Full backup does not return vclock_begin, so all golden files have been overwritten. In addition, the tt backup start integration tests have been changed. Now they are compared with golden files, which makes the comparison deterministic.

Part of TNTP-8608


bigbes (approved):

cli/cmd/cluster.go
patapenka-alexey: Please make the lines shorter.

sssciel: Done

patapenka-alexey (approved):

PR #1340, opened 2026-07-30
tail: preserve pending lines during log rotation

bigbes:
- switch `cli/tail` to `tarantool/go-tail` pinned to commit `0e16f7e`
- make rotation tests require pending records from the old file before reopening the replacement
- cancel followers when watched log roots disappear while keeping unaffected application roots active
- bound tail shutdown so `tt log -f` can exit when an upstream watcher does not stop

Closes TNTP-3131

### DIFF
--- a/cli/tail/follow_test.go
+++ b/cli/tail/follow_test.go
@@ -103,10 +103,6 @@ func TestFollow2_ReadExistingContent(t *testing.T) {
 }
 
 func TestFollow2_FollowNewContent(t *testing.T) {
-	if os.Getenv("CI") != "" {
-		t.Skip("Skipping flaky test on CI until issue #TNTP-3131 is fixed")
-	}
-
 	lf := createTmpLogFile(t, linesPerStep, logLineFormat)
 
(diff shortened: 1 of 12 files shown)


cli/tail/follow_test.go
```diff
@@ -195,89 +218,82 @@ func TestFollow2_NonExistentFile(t *testing.T) {
...
 	if err != nil {
-		t.Skipf("Failed to check initial lines in file: %v", err)
-		return
+		t.Fatalf("Failed to check initial lines in file: %v", err)
 	}
 
-	err = os.Rename(lf, lf+".bak")
+	// TailFile starts its filesystem watcher asynchronously. Keep that startup
+	// outside the write-plus-rename scenario this test exercises.
+	time.Sleep(watcherStartDelay)
```
sssciel, line 201: Is 100ms enough for startup? Can this lead to flacks?

Mockird31 (approved):

sssciel (approved):

PR #1352, opened 2026-08-18
status: add parallel status execution

sssciel:
Previously, the status call was called linearly on all instances,
which could lead to long delays. Now the calls are parallel.

On 200 instances, the total time gain is 12 times on
the test machine, from 5 seconds to 0.3 seconds. 

Part of TNTP-6354

### DIFF
--- a/cli/connector/connector.go
+++ b/cli/connector/connector.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"sync"
 	"time"
 
 	"github.com/tarantool/go-tarantool"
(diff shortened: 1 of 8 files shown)


cli/connector/connector.go
```diff
@@ -18,6 +19,59 @@ const (
...
+	}
+
+	shortAddress := "./" + filepath.Base(address)
+	if pathNeedsShortening && len(shortAddress)+1 > maxSocketPath {
+		return "", nil, fmt.Errorf("socket name is longer than %d symbols: %s",
+			maxSocketPath-3, filepath.Base(address))
+	}
+
+	// Relative paths also depend on the process-wide working directory.
+	connectMutex.Lock() // unlock in cleanup.
```
patapenka-alexey:
```
connectMutex.Lock() // **U**nlock in cleanup.
```

cli/status/status.go
```diff
@@ -195,34 +209,88 @@ func collectInstanceState(run running.InstanceCtx, fullInstanceName string,
...
+) <-chan statusResult {
+	statuses := make(chan statusResult, len(instances))
+	sem := semaphore.NewWeighted(maxParallelStatusRequests)
+	ctx := context.Background()
+
+	var wg sync.WaitGroup
+	for _, instance := range instances {
+		if err := sem.Acquire(ctx, 1); err != nil {
+			break
 		}
```
patapenka-alexey, line 232: What if we move Acquire() inside the goroutine (234:237)?

patapenka-alexey (commented):
Comments above.

Mockird31 (approved):

cli/connector/connector.go
sssciel: Fixed.

cli/status/status.go
sssciel: Discussed.

patapenka-alexey: Yes, let's leave everything as it is. Limiting the creation of goroutines to 128 pieces is what we wanted to achieve.

patapenka-alexey (approved):

Mockird31 (approved):

PR #1359, opened 2026-08-31
cli/cmd: tt package add, remove, update

bigbes:
`Engine.Resolve` always took the newest version a constraint allowed, so any command that rewrites the manifest — `add`, `remove` — would drag every unrelated dependency forward on the re-resolve the changed `manifest_hash` forces. That contradicts the rule `IsStale` states: only an explicit `tt package update` pulls fresh registry versions.

- cli/manifest/resolve: prefer pinned versions — `ResolvePinned` prefers a caller-supplied version per rock and `PinsFromLock` builds that set from an existing lock minus the names the caller wants freed. `Resolve` is now `ResolvePinned` with an empty set, so its callers are unaffected. A pin is a preference, not a constraint: when it no longer fits it is dropped, the rock resolves normally and a warning says so.
- cli/manifest: edit dependency tables in place — an `Editor` rewrites `[dependencies]`/`[dev_dependencies]` in the manifest's own TOML text instead of re-marshaling the model, so comments, key order and the formatting of every untouched line survive an edit. Short, inline-table, sub-table and root dotted-key declarations are all editable; anything else is refused by name rather than guessed at.
- cli/manifest/deps: edit and re-resolve deps — edit the declaration with the position-preserving TOML editor, write the manifest, re-parse it so `manifest_hash` matches what is on disk, re-resolve under a per-command pin set and rewrite the lock. The pin set is what separates the three commands: `add` and `remove` hold every version the lock already chose, a bare `update` pins nothing, `update NAME` frees that one rock and holds the rest.
- cli/cmd: tt package add, remove, update — `add` takes an optional constraint and a `--dev` flag choosing the table; `remove` and `update` take a name, and `update` takes none for a whole-closure refresh. Each prints what it changed and then the closure diff, one rock per line: a rock that arrived shows only its new version, one that left is marked dropped, one that moved shows both ends.
- lint/ci: enforce the strict ruleset on every push — the strict `default: all` ruleset had a job in CI and never ran, because that job is gated on a `full-commit-check` PR label, and two packages joined its declared scope in the meantime and accumulated 254 findings between them. The scope now names the packages that are actually clean under it, the ruleset gets a job with no label condition, and both lint targets state and verify the linter major they need instead of shelling out to whatever `golangci-lint` is on `PATH`.

Closes TNTP-9958

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -38,6 +38,46 @@ jobs:
       - name: Static code check
         uses: ./.github/actions/static-code-check
 
+  # The strict `default: all` ruleset (.golangci.yml), over the packages its
+  # path-except rule names as adopted. Unlike static-code-check this job carries
+  # no label condition: it runs on every push and pull request, which is the
+  # point. The same ruleset was already wired as a pre-commit hook, but only
+  # inside the label-gated job above, so in practice it never ran - `pack` and
+  # `inventory` were added to its scope and accumulated 254 findings between
+  # them without anything going red.
+  strict-lint:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          # golangci-lint type-checks the whole module, and cli/aeon/protos is a
+          # submodule.
+          submodules: recursive
+
+      - name: Setup Go
+        uses: actions/setup-go@v7
+        with:
+          go-version: '${{ env.GO_VERSION }}'
+
+      - name: Setup Mage
+        run: |
+          git clone https://github.com/magefile/mage
+          cd mage
+          go run bootstrap.go
+        shell: bash
+
+      # Pinned, and pinned to the same version the golangci-lint-strict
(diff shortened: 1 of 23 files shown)


.github/workflows/tests.yml
```diff
@@ -38,6 +38,46 @@ jobs:
...
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@master
+        with:
+          # golangci-lint type-checks the whole module, and cli/aeon/protos is a
+          # submodule.
+          submodules: recursive
+
+      - name: Setup Go
+        uses: actions/setup-go@v4
```
patapenka-alexey:
```suggestion
        uses: actions/setup-go@v7
```
Up to you.

patapenka-alexey (approved):

sssciel (approved):

.github/workflows/tests.yml
bigbes: fixed, thanks!

PR #1378, opened 2026-09-10
cli/cmd: manifest-aware tt run and tt test

bigbes:
`tt run` and `tt test` are the two workflow shortcuts over a manifest package.

## `tt run [ARGS...]`

Replaces the environment-driven command. The working directory must hold `app.manifest.toml` and nothing else is required — no `tt.yaml`, no tt environment, and none is looked for. The interpreter is the one the package bundles under `_runtime/` when it has one and the host's `tarantool` otherwise; `TT_USE_SYSTEM_TARANTOOL=1` forces the host's either way. Arguments reach Tarantool untouched, `--` included, and tt replaces itself with the interpreter, so signals and the exit code are Tarantool's own.

The project root is the working directory and is never searched for upwards, so which package runs is exactly the directory the shell is in.

## `tt test [SUB-PATH] [-- LUATEST-ARGS...]`

Builds the package, makes sure a runner is in `.rocks/`, then execs luatest under the same interpreter `tt run` would choose. Tests come from `test/`, or `tests/` when there is no `test/`; a sub-path narrows the run to one directory or one file and must exist, so a typo is an error rather than a run of nothing the runner would report as a pass. Everything after `--` is luatest's, and so is the exit code.

luatest need not be declared. When `[dev_dependencies]` does not name it and the tree does not already hold it, `build.EnsureDevRocks` resolves it as an implicit `*` requirement on top of the declared dev set, pinned to the lock's own picks, and installs what is missing. `app.manifest.toml` and `app.manifest.lock` are left byte-identical: writing either would change the manifest hash and leave the next `--locked` build refusing a lock the test run made stale.

Part of TNTP-9957


bigbes:
Diffs per group of changes, rendered on diffshub:

- **`tt run` over the manifest package** — https://diffshub.com/tarantool/tt/compare/fa2f0146e1da57065fb6558dfcac4d8ccda942b4...3ac9dfb3e6f6b4b86ecd8086b47e745f1be9fa21
  Interpreter selection (`_runtime/` first, then the tt environment's tarantool, then PATH) and the command that execs it with the arguments untouched; the launch tests follow the command to the manifest.
- **`tt test` with an implicit luatest** — https://diffshub.com/tarantool/tt/compare/3ac9dfb3e6f6b4b86ecd8086b47e745f1be9fa21...9fcb0df6400d8214d30695c53b9beb5291c3544d
  Extra dev requirements in the resolver, `build.EnsureDevRocks` that installs them without touching the manifest or the lock, and the command itself.
- **Tests, cleanup and CHANGELOG** — https://diffshub.com/tarantool/tt/compare/9fcb0df6400d8214d30695c53b9beb5291c3544d...364c311f7da14227791c6339c24a183eb85b6587
  Integration suites for both commands, the dead interpreter exec path in `cli/running` removed, `tt run` dropped from the config-required cases.

Whole PR: https://diffshub.com/tarantool/tt/pull/1378

sssciel (approved):

patapenka-alexey (approved):

PR #1384, opened 2026-09-17
cli/manifest: fix gaps in the package commands

bigbes:
Six gaps found in the manifest package commands. Each carries its own tests, and every commit builds and passes the manifest and command tests on its own.

- **cli/manifest: default source in sub-table deps**: a dependency written as a standard sub-table (`[dependencies.x]` with `version` beneath it) was refused with `source "" is not supported`, while the same declaration written inline was accepted. go-toml hands a value to an `Unmarshaler` only when it is a node of its own, so the `registry` default in `UnmarshalTOML` never ran for that form. The default is now applied after the decode, in every table a dependency can be declared in.
- **cli/cmd: work on a project elsewhere with -C**: declared once on the `tt package` group and on `tt new` and `tt test`, and read in `absoluteWorkingDir`, the one helper every project-scope command already calls. `tt run` does not take it: it parses no flags of its own, because every argument it gets belongs to Tarantool.
- **test/integration: cover build, pack and install**: the three commands at the centre of the pipeline had Go tests and no integration suite. The new suite drives the real binary through build, pack and install in one chain, offline, against the fixture rock repository as a local directory registry. `zstandard` joins the test requirements because a `.tt` is `tar.zst`, which `tarfile` reads only from Python 3.14 on.
- **cli/manifest: imply a product when none declared**: a manifest without `[products]`, the shape `tt new` writes, was read three ways: the resolver locked its dependencies into no product and reported the empty lock as current, `tt package build` refused it, and `tt package deps` showed an unnamed group. It now has one implicit product, `default`, built from every component, and a lock already written empty is reported stale.
- **cli/manifest: exit 2 when the system fails**: a rock server that refused the connection, a host name that did not resolve and an unreadable manifest all exited 1. They now exit 2, classified once in `ExitCode` by error type. A malformed registry URL and a missing file stay 1, and an explicit code such as 3 for a partly successful install still stands.
- **help: describe exit codes in tt help errors**: a help topic explaining codes 0 to 3, listed in the root help under a new HELP TOPICS section.

Related to TNTP-8220
Related to TNTP-8611
Related to TNTP-8612
Related to TNTP-8613
Related to TNTP-9956
Related to TNTP-9959


bigbes:
Six independent changes, one commit each:

- [Sub-table dependencies take the registry default](https://diffshub.com/tarantool/tt/commit/f10008c4f96ed59f113d75b226e5cf906e276e3d): `[dependencies.x]` with only a `version` was refused while the inline form passed.
- [-C works on a project that is not the cwd](https://diffshub.com/tarantool/tt/commit/1e2cfba2bd47ac46c603cb5b1e1184e05217ac31): declared once on `tt package`, `tt new` and `tt test`; `tt run` keeps taking no flags of its own.
- [Integration coverage for build, pack and install](https://diffshub.com/tarantool/tt/commit/416b91790232e36ea0ccb76faab0a8c3c579a27a): the real binary driven through build, pack and install, offline.
- [A manifest without products gets an implicit one](https://diffshub.com/tarantool/tt/commit/8963ed0692bf244fa63704781d246aaa6e2e5a9c): `tt new`, `tt package add`, `tt package build` now installs the dependency instead of locking nothing.
- [System failures exit 2](https://diffshub.com/tarantool/tt/commit/1d1bf74a34087ffe6a3ad355d615a3dcc6c0e99f): unreachable registry, DNS, permission and disk errors, classified once in `ExitCode`.
- [tt help errors](https://diffshub.com/tarantool/tt/commit/0561f197df04919cd292b91560630d7010ff7ad8): the exit codes described, and a HELP TOPICS section in the root help.

Whole PR: https://diffshub.com/tarantool/tt/pull/1384

sssciel (approved):

Mockird31 (approved):

PR #1385, opened 2026-09-21
backup: add tt backup-id command

bigbes:
Generate the backup identifier in tt instead of the orchestrator.

- `tt backup-id` prints the current UTC second as `20060102T150405Z` and a newline; the id passes as is to `--backup-id` of `tt backup start`, `upload` and `finalize`.
- The command exits only once the printed second is over, so two consecutive calls on one host never print the same id, and ids keep sorting in the order they are taken, as upload requires.

Closes TNTP-10353


bigbes:
- [`tt backup-id` command](https://diffshub.com/tarantool/tt/commit/789d568e0636d0ffed64779384ec1649a8594a74): generates the backup id (UTC second, waits the second out) and fails when it cannot be printed.

Whole PR: https://diffshub.com/tarantool/tt/pull/1385

Mockird31 (approved):

sssciel (approved):

PR #1386, opened 2026-09-21
restore: add --replicasets to restore plan

bigbes:
A selection is a set of instance names naming the replicasets a recovery is meant to cover. The chain sees no replicaset names - manifests carry UUIDs and instance names - so a replicaset counts as selected when at least one of its topology instances is named, the same intersection rule the restore topology check uses. `Build`, `Load` and `LoadPartial` take functional options, and `WithSelection` carries the set down to point stitching: a label becomes a cluster point once every selected replicaset holds it, whatever the others carry, and the point's `Shards` hold the selected UUIDs alone. A stateless vshard router, which keeps no recovery points, no longer keeps the storages from stitching one. `ClusterPoint.Topology` stays the full segment topology: it is what the consumer compares against the live cluster, and it has to show the replicasets left out so they can be reported rather than silently dropped.

- chain: select replicasets for cluster points
- restore: add --replicasets to restore plan
  - `--replicasets` names the replicasets to restore, by the names the cluster config gives them, and leaves the rest to be bootstrapped fresh. The flag requires `-c`; an unknown or empty name, or a flag that parses into no name at all, is refused before the storage is read. The topology check narrows both sides: a configured replicaset left out is a warning, a selected one the backup has no counterpart for is `topology_mismatch`. Without the flag the plan is what it was.
- restore: cover --replicasets in integration tests
- backup: add vshard restore lifecycle test
  - Backs up all three replicasets of a live vshard cluster, takes a cluster point through the router, plans the restore over the storages, applies on each of them from the plan document alone, and reads the rows back through the router. The control run without the flag resolves to no cluster point at all.
- restore: refuse a selection the backup holds nowhere
  - A selected replicaset no manifest carries is now `topology_mismatch` with the name in `extra_replicasets`, whether named alone or beside a backed-up one, instead of `no_recovery_point` with empty warnings.

Closes TNTP-10389


bigbes:
Diffs by group:

- https://diffshub.com/tarantool/tt/commit/c573a718e3dfce32a7edc433f8dd0cb8e5f4ccc5 — `chain`: a `Selection` of instance names threaded through `WithSelection` into point stitching; a cluster point needs its label on the selected replicasets only.
- https://diffshub.com/tarantool/tt/compare/c573a718e3dfce32a7edc433f8dd0cb8e5f4ccc5...f26503b5647e4d931b773f5963a50b52d00e96fd — `tt restore plan --replicasets`: the flag, the narrowed topology check, help, CHANGELOG, unit tests for multi-instance replicasets and the nearest-safe hint under a selection, and the pytest cases on a synthetic storage with a stateless router.
- https://diffshub.com/tarantool/tt/commit/c01526886629b739c760164cff2228ac1832df18 — end-to-end vshard lifecycle test: backup from all replicasets, plan over the storages, `restore apply` from the plan document, data read back through the router once it reaches every bucket with no alerts.
- https://diffshub.com/tarantool/tt/commit/0bf845b0659485e7d930a5a96a07d329467241c2 — review fix: a selection the backup holds nowhere is `topology_mismatch`, and the preflight pytest cases pin the flag-before-storage order.

Whole PR: https://diffshub.com/tarantool/tt/pull/1386

sssciel (approved):

Mockird31 (approved):

